PR #25, opened 2018-02-14
Conditionals tests && Proposal new switch translation

lolleko:


test/integration/lua/conditionals.spec.ts
```diff
@@ -0,0 +1,213 @@
...
+            return result;`
+            , util.dummyTypes.Number
+        );
+
+        // Execute
+        let result = util.executeLua(lua);
+
+        // Assert
+        Expect(result).toBe(expected);
+    }
```
Perryvw: Could you add a testcase where fallthrough has a visible effect? Like replacing the = with += here? Would like to test the return for inp 2 in this case. (should be 5)

src/Transpiler.ts
```diff
@@ -347,37 +347,59 @@ export class LuaTranspiler {
 
         let result = this.indent + "-------Switch statement start-------\n";
 
+        let jumpTableName = "____switch" + ++this.genVarCounter
```
Perryvw: Don't inline assignments (incrementor in this case).

src/Transpiler.ts
```diff
@@ -347,37 +347,59 @@ export class LuaTranspiler {
...
-            if (index < clauses.length - 1) {
-                result += this.indent + `goto switchCase${this.genVarCounter + index + 1}\n`;
+            let i = index + 1;
+            if (i < clauses.length && !tsEx.containsStatement(clause.statements, ts.SyntaxKind.BreakStatement)) {
+                let nextClause = clauses[i];
+                while(i < clauses.length
+                    && ts.isCaseClause(nextClause)
+                    && nextClause.statements.length === 0
+                ) {
+                    nextClause = clauses[++i];
```
Perryvw: Don't inline assignments (incrementor).

test/integration/lua/conditionals.spec.ts
lolleko: Added in b5c1582

PR #41, opened 2018-02-21
Fixed concat if seperator is omitted

lolleko:


src/Transpiler.ts
```diff
@@ -741,7 +741,12 @@ export class LuaTranspiler {
             case "splice":
                 return `TS_splice(${caller}, ${params})`;
             case "join":
-                return `table.concat(${caller}, ${params})`;
+                if (!params) {
```
Perryvw: params == ""

PR #54, opened 2018-02-25
Tests and Fixes

lolleko:
Better binary expression translation
Test for forbidden binary expressions

Test for continue
Test for break
Test for array.forEach
Test for lualib require
Test for nested namespaces

Fixed minor issue with nested namespaces
Removed dead code

### DIFF
--- a/src/Transpiler.ts
+++ b/src/Transpiler.ts
@@ -465,7 +465,7 @@ export class LuaTranspiler {
     }
 
     transpileThrow(node: ts.ThrowStatement): string {
-        if (ts.isStringLiteral(node.expression))  {
+        if (ts.isStringLiteral(node.expression)) {
             return `error("${node.expression.text}")`;
         } else {
             throw new TranspileError("Unsupported throw expression, only string literals are supported", node.expression)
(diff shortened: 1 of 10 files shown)


test/integration/lua/string.spec.ts
```diff
@@ -49,9 +49,22 @@ export class StringTests {
     @TestCase([42, "hello"], "42hello")
     @Test("string.concat[+]")
     public concat(inp: any[], expected: string) {
+        let concatStr = "";
```
Perryvw: ```let concatStr = inp.map(elem => elem === "string" ? `"${elem}"` : `${elem}`).join(" + ");```

src/Transpiler.ts
```diff
@@ -465,7 +465,7 @@ export class LuaTranspiler {
     }
 
     transpileThrow(node: ts.ThrowStatement): string {
-        if (ts.isStringLiteral(node.expression)) {
+        if (ts.isStringLiteral(node.expression))  {
```
Perryvw: extra space

PR #60, opened 2018-03-02
 Add luaLibDir compiler option and in operator, fix template strings

zapp-brannigan-dota:
`luaLibDir` is a new compiler option that specifies the save location for `typescript_lualib.lua` relative to `outDir`.

Typescript template strings now cast all expressions to string.


src/Compiler.ts
```diff
@@ -113,32 +132,41 @@ function removeInvalidOptionsFromArgv(commandLine: ReadonlyArray<string>, invali
...
     }
 }
 
 function executeCommandLine(args: ReadonlyArray<string>) {
     // Right now luaTarget, version and help are the only cli options, if more are added we should
     // add a more advanced custom parser
-    var argv = minimist(args, {
+    const argv = minimist(args, {
         string: ['l'],
         alias: { h: 'help', v: 'version', l: 'luaTarget' },
```
lolleko: Add CLI options for addHeader and lualiboutdir. Don't forget to update help text and readme.

src/Compiler.ts
```diff
@@ -42,9 +47,23 @@ function compile(fileNames: string[], options: ts.CompilerOptions): void {
...
+    }
+
+    if (!options.luaLibDir) {
+        options.luaLibDir = options.outDir;
+    }
+    else {
+        options.luaLibDir = path.join(options.outDir, options.luaLibDir);
+    }
+    if (!fs.existsSync(options.luaLibDir)){
+        fs.mkdirSync(options.luaLibDir);
```
lolleko: Pretty sure mkdir sync is not recursive. In this case this will error on a nested path.

src/Compiler.ts
```diff
@@ -42,9 +47,23 @@ function compile(fileNames: string[], options: ts.CompilerOptions): void {
...
 
+    if (!('addHeader' in options)) {
+        options.addHeader = true;
+    }
+
+    if (!options.luaLibDir) {
+        options.luaLibDir = options.outDir;
+    }
+    else {
+        options.luaLibDir = path.join(options.outDir, options.luaLibDir);
```
lolleko: should be path.resolve, just to be sure.

src/Compiler.ts
```diff
@@ -11,7 +11,12 @@ import dedent = require("dedent")
 import { LuaTranspiler, TranspileError } from "./Transpiler";
 import { TSHelper as tsEx } from "./TSHelper";
 
-function compile(fileNames: string[], options: ts.CompilerOptions): void {
+interface CompilerOptions extends ts.CompilerOptions {
+    addHeader?: boolean;
+    luaLibDir?: string;
+}
```
lolleko: Add `luaTarget: string` aswell

src/Compiler.ts
```diff
@@ -42,9 +47,23 @@ function compile(fileNames: string[], options: ts.CompilerOptions): void {
...
+        options.luaLibDir = path.join(options.outDir, options.luaLibDir);
+    }
+    if (!fs.existsSync(options.luaLibDir)){
+        fs.mkdirSync(options.luaLibDir);
+    }
+
     // Copy lualib to target dir
     // This isnt run in sync because copyFileSync wont report errors.
-    fs.copyFile(path.resolve(__dirname, "../dist/lualib/typescript.lua"), path.join(options.outDir, "typescript_lualib.lua"), (err: NodeJS.ErrnoException) => {
+    fs.copyFile(path.resolve(__dirname, "../dist/lualib/typescript.lua"), path.join(options.luaLibDir, "typescript_lualib.lua"), (err: NodeJS.ErrnoException) => {
```
lolleko: When you change options.luaLibDir the require call should be resolved, in order to point to that new path, https://github.com/Perryvw/TypescriptToLua/blob/master/src/Transpiler.ts#L36

src/Compiler.ts
lolleko: Typescript has a function to recursively create files / directories. not sure If that function is exposed tho.

src/Compiler.ts
```diff
@@ -113,32 +132,41 @@ function removeInvalidOptionsFromArgv(commandLine: ReadonlyArray<string>, invali
...
+                invalidErrorCount += 1;
+            }
+            else {
+                console.log(err.messageText);
+            }
         });
-        process.exit(1);
+        if (invalidErrorCount != commandLine.errors.length) {
+            process.exit(1);
+        }
```
Perryvw: all of `logError` looks pretty dodgy. Improve or add some comments that explain what/why this function does.

Perryvw:
Not sure if we want to support luaLibDir in the first place, sounds like a lot of hacky stuff for little gain.

Perryvw:
After some discussion we decided not to want the lualibdir stuff in the main branch. Like I said in the previous comment it seems like asking for trouble with limited pay-off. The rest of the pull request looks good and I'd be happy to merge if you remove the lualibdir part.

zapp-brannigan-dota:
luaLibDir solves the problem of a directory structure like this:
```
src/
- game/
  - tsconfig.json
  - scripts/
    - vscripts/
      - ...
```
There is no additional resolving needed. It can only be used to specify the correct subfolder, nothing else.

src/Compiler.ts
zapp-brannigan-dota: See comment below.

lolleko:
Did you try setting rootDir to `./scripts/vscripts/`

zapp-brannigan-dota:
rootDir solves the problem, so I removed luaLibDir. It was broken on Windows but is fixed now. I can't tell if it works on other platforms tho.

zapp-brannigan-dota:
There should tests for compiling different types of skeleton projects. I might make them at some point but I've never done that so some examples would be very helpful.

PR #62, opened 2018-03-10
Add multiple argument support to array.push()

zapp-brannigan-dota:
If this is accepted the wiki can be updated too.

### DIFF
--- a/dist/lualib/typescript.lua
+++ b/dist/lualib/typescript.lua
@@ -115,6 +115,12 @@ function TS_split(str, separator)
     return out
 end
 
+function TS_push(list, ...)
+    for _, v in pairs({...}) do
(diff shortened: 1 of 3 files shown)


lolleko:
Might be better to do that in lualib instead of hacking it into the translation.

zapp-brannigan-dota:
Yeah that's much cleaner.

lolleko:
Might also be better to use `tbl[#tbl + 1] = "newValue"` instead of table.insert since its faster. Opinions @Perryvw?

Perryvw:
If `tbl[#tbl + 1]` is faster I don't see why not.

lolleko:
I think JIT probably already optimises that, but its defiantly faster on other Lua versions.

PR #63, opened 2018-03-10
Trying to fix polymorphism calls

Perryvw:
Messing around with instance calls again, now using `:` for interface and class calls.

Please try this branch and let me know if/what breaks.


lolleko:
Switch statements are broken because they use `self` too (my bad lol), should be an easy fix if you change it from self to something else in this line https://github.com/Perryvw/TypescriptToLua/blob/master/src/Transpiler.ts#L394 and adjust the jump calls accrodingly.

PR #67, opened 2018-03-13
Add compiler tests

zapp-brannigan-dota:
Please also enable [AppVeyor](https://ci.appveyor.com) for Windows testing.

### DIFF
--- a/test/integration/compiler.spec.ts
+++ b/test/integration/compiler.spec.ts
@@ -1,4 +1,4 @@
-import { Expect, Test, TestCase } from "alsatian";
+import { Expect, Teardown, Test, TestCase } from "alsatian";
 import { execSync } from "child_process"
 import { existsSync, removeSync, unlink } from "fs-extra"
 import { resolve } from "path"
@@ -28,7 +28,10 @@ export class CompilerTests {
                 throw error;
             });
         });
+    }
 
+    @Teardown
+    public teardown() {
         // Delete outDir folder
         removeSync('test/integration/project/outDir');
     }
--- a/test/integration/compiler.spec.ts
+++ b/test/integration/compiler.spec.ts
@@ -1,0 +1,36 @@
+import { Expect, Test, TestCase } from "alsatian";
+import { execSync } from "child_process"
+import { existsSync, removeSync, unlink } from "fs-extra"
+import { resolve } from "path"
+
+export class CompilerTests {
+
+    @TestCase('tsconfig.default.json', ['typescript_lualib.lua', 'nested/folder/file.lua'])
+    @TestCase('tsconfig.outDir.json', ['outDir/typescript_lualib.lua', 'outDir/nested/folder/file.lua'])
+    @TestCase('tsconfig.rootDir.json', ['nested/typescript_lualib.lua', 'nested/folder/file.lua'])
+    @TestCase('tsconfig.bothDirOptions.json', ['outDir/typescript_lualib.lua', 'outDir/folder/file.lua'])
+    @Test("Compile project")
+    public compileProject(tsconfig: string, expectedFiles: string[]) {
+        const compilerPath = resolve('dist/Compiler.js');
+        const tsconfigPath = resolve('test/integration/project', tsconfig);
+
+        // Compile project
+        execSync(`node ${compilerPath} -p ${tsconfigPath}`);
+
+        expectedFiles.forEach(relativePath => {
+            const absolutePath = resolve('test/integration/project', relativePath);
+
+            // Assert
+            Expect(existsSync(absolutePath)).toBe(true);
+
+            // Delete file
+            unlink(absolutePath, (error) => {
+                throw error;
+            });
+        });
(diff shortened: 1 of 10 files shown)


test/integration/project/tsconfig.base.json
```diff
@@ -0,0 +1,9 @@
+{
+    "compilerOptions": {
+        "noImplicitAny" : true,
+        "noImplicitThis" : true,
+        "alwaysStrict" : true,
+        "strictNullChecks": true,
```
lolleko: Strict flags (and others) are tsc flags and should not be tested by our compiler.

test/integration/project/tsconfig.base.json
```diff
@@ -0,0 +1,9 @@
...
+        "luaTarget": "JIT"
```
lolleko: luatarget ist not a compilerOption and needs to be declared in the root of the config.json.

test/integration/compiler.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+            // Delete file
+            unlink(absolutePath, (error) => {
+                throw error;
+            });
+        });
+
+        // Delete outDir folder
+        removeSync('test/integration/project/outDir');
+    }
+
```
lolleko: I am not sure this is a good way to test the compiler.

test/integration/compiler.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+            Expect(existsSync(absolutePath)).toBe(true);
+
+            // Delete file
+            unlink(absolutePath, (error) => {
+                throw error;
+            });
+        });
+
+        // Delete outDir folder
+        removeSync('test/integration/project/outDir');
```
lolleko: please use a function with `@teardown` for cleanup https://github.com/alsatian-test/alsatian/wiki/pre-and-post-test.

test/integration/compiler.spec.ts
```diff
@@ -0,0 +1,36 @@
+import { Expect, Test, TestCase } from "alsatian";
+import { execSync } from "child_process"
+import { existsSync, removeSync, unlink } from "fs-extra"
```
lolleko: Why do you need `fs-extra` all those functions exist in default `fs`.

test/integration/compiler.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+    @TestCase('tsconfig.outDir.json', ['outDir/typescript_lualib.lua', 'outDir/nested/folder/file.lua'])
+    @TestCase('tsconfig.rootDir.json', ['nested/typescript_lualib.lua', 'nested/folder/file.lua'])
+    @TestCase('tsconfig.bothDirOptions.json', ['outDir/typescript_lualib.lua', 'outDir/folder/file.lua'])
+    @Test("Compile project")
+    public compileProject(tsconfig: string, expectedFiles: string[]) {
+        const compilerPath = resolve('dist/Compiler.js');
+        const tsconfigPath = resolve('test/integration/project', tsconfig);
+
+        // Compile project
+        execSync(`node ${compilerPath} -p ${tsconfigPath}`);
```
lolleko: Don't use exec to test. Errors won't be reported correctly also coverage reports probably won't work either.

test/integration/compiler.spec.ts
zapp-brannigan-dota: `fs-extra` can handle non empty directories.

test/integration/compiler.spec.ts
zapp-brannigan-dota: What should I use instead?

test/integration/compiler.spec.ts
zapp-brannigan-dota: These tests are about `outDir` and `rootDir` and I think those are appropriately tested with that. But I'd be happy to implement something else if you tell me what you think it should be instead.

test/integration/compiler.spec.ts
lolleko: function executeCommandLine(args: ReadonlyArray<string>)

test/integration/compiler.spec.ts
lolleko: Right makes sense. Might want to rename the test.
I was thinking about checking if paths are resolved correctly directly, but this is hard to test since all the resolve stuff is packed into parseCommandLine() and compile() which makes it hard to test directly. So checking if a file exists is probably the best way right now. I might revisit this when I rewrite the CLI interface.

test/integration/compiler.spec.ts
lolleko: RIP thought default node could do that by now.

test/integration/compiler.spec.ts
zapp-brannigan-dota: I can't figure this out, I tried passing a callback but it never gets called.

lolleko: to be fixed later #68

PR #70, opened 2018-03-14
CLI Rework

### LINKED ISSUE #68: Improve CLI and add more tests
Specifically make executeCommendLine more test friendly.

### LINKED ISSUE #71: compilerOption: dontRequireLualib
First of all, this is an incredibly useful tool! Thank you so much for creating it.

A bit of background/thoughts before I start asking for things:
>I've been using lua for a bit over 6 years now (mainly as the scripting language for Company of Heroes >1 and 2). Lack of static types has been a limiting factor. 
>I'm aware therte are typed variants of Lua but those usually won't solve the other half of the problem: >Editor/IDE support. TypeScript has excellent support on VS Code which makes it the perfect **transpile->to-**<*insert language here*\> language. 

**The issue/request:**
Looks like the option [`dontRequireLualib`](https://github.com/Perryvw/TypescriptToLua/blob/master/src/Transpiler.ts#L34) is at least a planned feature since it works if you add it to the [tstlOptions](https://github.com/Perryvw/TypescriptToLua/blob/master/src/Compiler.ts#L132) configuration:

```typescript
        'dontRequireLualib': {
            alias: 'dontRequireLualib',
            describe: '',
            default: false,
            type: 'boolean'
        }
```
(Otherwise the process [gets terminated without specifying the reason](https://github.com/Perryvw/TypescriptToLua/blob/master/src/Compiler.ts#L126))

Looks like the most [recent pull request is about to touch the compiler options](https://github.com/Perryvw/TypescriptToLua/pull/70), perhaps this feature could be snuck in? I suppose I could submit a pull request for this too.

**Why?**
This is an important functionality in CoH1's/CoH2's enbedded & customized lua enviroment since the developers provided a custom function for importing libraries. The use case is extremely narrow but since it's 99% implemented, I thought I'd ask.

lolleko:
Cleaned up parser

Closes #68, Closes #71


src/CommandLineParser.ts
```diff
@@ -0,0 +1,169 @@
...
+        alias: 'luaTarget',
+        choices: ['JIT', '5.3'],
+        default: 'JIT',
+        describe: 'Specify Lua target version.',
+        type: 'string'
+    },
+    'ah': {
+        alias: 'addHeader',
+        describe: 'Specify if a header will be added to compiled files.',
+        default: false,
```
Perryvw: I think we should add the header by default

src/CommandLineParser.ts
```diff
@@ -0,0 +1,169 @@
...
+    let optNameMap: OptionNameMap = {};
+    for (let key in optionDeclarations) {
+        optNameMap[key] = true
+        let alias = optionDeclarations[key].alias;
+        if (typeof alias === "string") {
+            optNameMap[alias] = true;
+        } else {
+            alias.forEach((val) => optNameMap[val] = true);
+        }
+    }
```
Perryvw: Very unclear what the purpose of this loop is and why it sets values to true, add some explanation.

src/CommandLineParser.ts
```diff
@@ -0,0 +1,169 @@
...
+
+            if (!ignore) {
+                console.log(`error TS${err.code}: ${err.messageText}`);
+                process.exit(1);
+            }
+        });
+    }
+}
+
+/** Find configFile, function form ts api seems to be broken? */
```
Perryvw: from* (typo)

src/CommandLineParser.ts
lolleko: Idk I think it looks cheap. Like those weird video converters youtube downloaders I used as a kid and that left their watermark everywhere.

Or is there an actual benefit in having a header that im not seeing?

Perryvw: No actual benefit, just some credit/advertising for us in case people see the code in other people's repositories. Also makes it clear the code is generated.

lolleko: Yeah that acutally makes sense.
But maybe reduce the it to a one liner?
I will jsut reeanble it and we can change that later if we want to....

Perryvw: I am fine with changing it to be only 1 or 2 lines. I got a nice comment from @zapp-brannigan-dota about maybe adding the version number. I would say the header should at least contain the following things: 'Generated by TypescriptToLua https://github.com/Perryvw/TypescriptToLua', the date, and the version of the transpiler that was used (should be possible right?).

zapp-brannigan-dota: Why is the date important? With it I wouldn't add the header to any files that might end up version controlled.

Perryvw: Figured date would be useful for keeping track of when files changed but I guess it will also be in the file metadata/commits so maybe we can indeed leave it out.

README.md
```diff
@@ -20,30 +20,6 @@ More detailed documentation and info on writing declarations can be found [on th
...
-**Options**
-```
-tstl [options] [files...]
-
-In addition to the options listed below you can also pass options for the
-typescript compiler (For a list of options use tsc -h).
-
-NOTES:
-- The tsc options might have no effect.
-- Options in tsconfig.json are prioritized.
```
zapp-brannigan-dota: I think `Options in tsconfig.json are prioritized.` should be in here or the wiki.

lolleko: Changed it so that CLI is prioritized. This is the same behaviour now as in `tsc` and probably any cli tool out there. So I don't think its **needed** to mention that, but it wouldn't hurt either if this is mentioned somewhere in the docs.

zapp-brannigan-dota: That's much better. It only needed mentioning because it was different from tsc.

PR #100, opened 2018-04-26
Fixed major bug with varargs/spread operator

### LINKED ISSUE #95: No transpilation for spread operator
As titled; example as follows:

image

image

As you can see, `rest => 'a'` which will result in an error.

I have fixed this issue in a subset project of TSTL. However, I cannot perform a pull request as my subset is too far removed from TSTL at this point.

Aeomi:
A bug because usage of the spread operator transpiled without error into Lua, where the varargs array would be transpiled into only the first value of the varags; see [here](https://github.com/Perryvw/TypescriptToLua/issues/95).

Closes #95

### DIFF
--- a/src/Transpiler.ts
+++ b/src/Transpiler.ts
@@ -1151,14 +1151,33 @@ export class LuaTranspiler {
 
         // Build parameter string
         const paramNames: string[] = [];
-        parameters.forEach((param) => {
-            paramNames.push((param.name as ts.Identifier).escapedText as string);
-        });
+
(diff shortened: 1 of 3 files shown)


lolleko:
Looks good if you remove the failing "undefined tests" and make a separate PR/Issue for them, we can merge this.

Aeomi:
I had to grapple with git to undo those changes, I wasn't aware that the new commit would be appended to this pull request. I'm certain it's safe to merge now.

Issue #108, opened 2018-05-30
Distinction between dot and colon syntax for methods


apemanzilla:
Some Lua APIs provide objects with methods that must be invoked using dot syntax (`thing.method(...)`) as opposed to colon syntax (`thing:method(...)`). Having a custom decorator for this would be very useful.

Edit: This behavior already exists for namespaces, it just needs to be expanded for object-oriented use. I propose a new custom decorator, `!DotMethod`, applicable to declarations of classes, interfaces, type literals, and methods of any of these.

Example usages:
```typescript
/** !DotMethod */
declare class SomethingExternal {
	doThing(): SomethingReturned;
}

/** !DotMethod */
interface SomethingReturned {
	doOtherThing(): void;
}

type SomethingElse = {
	/** !DotMethod **/
	dotMethod(): void,
	regularMethod(): void
}
```

lolleko:
You can already use namespaces to accomplish that:

e.g: https://github.com/ark120202/types-lua/blob/master/types/table.d.ts

We should probably mentioned that in the docs somewhere, sorry. I will update the wiki accordingly.

apemanzilla:
Namespaces don't cover everything, though - for example, if you have some function that returns an object with methods that must be called with dot syntax.

Perryvw:
If you declare a method on a class as static it *should* use the dot syntax.

apemanzilla:
That still doesn't help for instance methods that use the dot syntax. Say I have some function that returns an object with methods like so:

```lua
local function getTheObject()
	return { method = function(argument) --[[ do something ]] end }
end

local obj = getTheObject()
obj.method("some argument")
```

There's currently no way to declare this in a way that uses the dot syntax to perform the call.

apemanzilla:
Also, in testing it seems that static methods are still invoked using colon syntax.

Perryvw:
I don't really like the idea of using a decorator for this, feels like there has to be a better way. Will think about it for a bit.

apemanzilla:
For comparison, haxe (4.0+) has an annotation `@:luaDotMethod` which works similarly to the decorator I implemented in #109

apemanzilla:
If you come up with a better alternative, let me know and I'll update my PR.

apemanzilla:
Any changes here? I've been using #109 without issue.

Perryvw:
Contemplating if we could simply use dot methods when the owner of the function is typed as object literal. For example:
```ts
[...]

PR #109, opened 2018-05-31
Improve dot vs colon method invocations

apemanzilla:
This implements #108, as well as a couple other fixes:

- Adds a `!DotMethod` decorator, which can be used on classes, interfaces, and methods to indicate they should be called using dot syntax (e.g. `instance.method()` as opposed to `instance:method()`)
- Fixes static methods, which are now always called using dot syntax
- Adds a test that ensures that the appropriate syntax is used with classes, interfaces, instance methods, static methods, and the decorator.


apemanzilla:
Closed in favor of #111

PR #111, opened 2018-06-03
Dot versus Colon 2

Perryvw:
Counter proposal to #109 and #108

### DIFF
--- a/src/Transpiler.ts
+++ b/src/Transpiler.ts
@@ -917,9 +917,12 @@ export class LuaTranspiler {
                 return this.transpileArrayCallExpression(node);
             }
 
-            if (expType.symbol && ((expType.symbol.flags & ts.SymbolFlags.Namespace)
-                                || (expType.symbol.flags & ts.SymbolFlags.TypeLiteral))) {
+            const expType2 = this.checker.getTypeAtLocation(node.expression);
+            if (expType.symbol &&
                 // Don't replace . with : for namespaces
+                ((expType.symbol.flags & ts.SymbolFlags.Namespace)
+                // If function is defined as property with lambda type use . instead of :
+                || (expType2.symbol.flags & ts.SymbolFlags.TypeLiteral))) {
                 callPath = this.transpileExpression(node.expression);
                 params = this.transpileArguments(node.arguments);
                 return `${callPath}(${params})`;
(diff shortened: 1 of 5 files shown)


Perryvw:
I actually remembered why we use `:` for static methods: when using `this` in a static method it usually refers to the class itself. If we look at what happens in lua this is perfectly preserved by using the `:` syntax and adding a 'self' parameter.

These other ones indeed seem like issues and I will have a look at how to resolve those when I have time.

apemanzilla:
Hmm, that does make sense for `static`, although there are again cases where APIs expect to be called using dot syntax, which complicates things.

Perryvw:
@apemanzilla @lolleko tried to solve by transpiling lambda properties with `.` and methods with `:`. Have a look

src/Transpiler.ts
```diff
@@ -917,8 +917,12 @@ export class LuaTranspiler {
                 return this.transpileArrayCallExpression(node);
             }
 
-            if (expType.symbol && (expType.symbol.flags & ts.SymbolFlags.Namespace)) {
+            const expType2 = this.checker.getTypeAtLocation(node.expression);
```
apemanzilla: It might be worth renaming `expType` and `expType2` to something more descriptive - such as `containerType` and `functionType` respectively.

lolleko: agreed

test/translation/ts/dotColonFunctionCalls.ts
```diff
@@ -0,0 +1,24 @@
...
+
+declare const classInstance: TestClass;
+declare const interfaceInstance: TestInterface;
+
+classInstance.colonMethod();
+classInstance.dotMethod();
+interfaceInstance.colonMethod();
+interfaceInstance.dotMethod();
+TestNameSpace.dotMethod();
+TestNameSpace.dotMethod2();
```
lolleko: Not sure how much sense it makes but maybe add some functional tests too. Probably not that important.

apemanzilla:
Is there anything still holding this back?

apemanzilla:
Found another minor edge case. If you have an instance of an interface extending from two interfaces defining the same method as an arrow function and regular function, it will use the first parent interface to determine the function type. I can't think of any cases where this really matters, though.

```typescript
declare interface A2 {
    method: () => void;
}

declare interface B2 {
    method(): void;
}

declare interface Thing extends B2, A2 {}

declare const thing: Thing;

thing.method();
```

Perryvw:
@apemanzilla I will merge this once I added some functional tests and had a quick look if there is any easy way to deal with union types.

apemanzilla:
Thank you!

PR #115, opened 2018-06-06
Update isArrayType to include tuples

apemanzilla:
Tuples are a type of array and have the same semantics, so this PR changes `isArrayType` to include tuples. This fixes a few issues when using tuples with for-of loops, `Array.prototype` methods, and so on. It also includes a test for these cases.


apemanzilla:
Thanks!

PR #117, opened 2018-06-07
Version split

### LINKED ISSUE #116: Split up transpiler for different Lua Versions

lolleko:
Closes #116


src/Transpiler.ts
```diff
@@ -18,40 +18,15 @@ export class TranspileError extends Error {
...
 export enum LuaTarget {
+    Lua51 = "5.1",
+    Lua52 = "5.2",
     Lua53 = "5.3",
     LuaJIT = "JIT",
 }
 
-export class LuaTranspiler {
+export abstract class LuaTranspiler {
     public static AvailableLuaTargets = [LuaTarget.LuaJIT, LuaTarget.Lua53];
```
apemanzilla: This isn't needed any more

src/targets/Transpiler.52.ts
```diff
@@ -0,0 +1,7 @@
+import { LuaTranspiler, TranspileError } from "../Transpiler";
+import { TSHelper as tsHelper } from "../TSHelper";
+
+import * as ts from "typescript";
+
+export class LuaTranspiler52 extends LuaTranspiler {
```
apemanzilla: Lua 5.2 added the `bit32` library, this could be used to implement bit operations

lolleko: Sure, but we should probably do that in a different PR.

src/Transpiler.ts
lolleko: It prevents the class from being instantiated.

src/targets/Transpiler.52.ts
lolleko: Moved to #119

PR #124, opened 2018-06-25
Disallow usage of goto and label in Lua 5.1

zengjie:
Lua 5.1 does not support `goto` and label. Using `goto` and label to simulate `continue` is not possible for Lua 5.1 target.


Perryvw:
Will merge if you fix the tests.

Perryvw:
Actually, as @lolleko pointed out to me, this would mess up our inheritance tree. You should move the current goto/continue implementation to the 5.2 implementation and then put this 'default' version in the general transpiler.ts.  We will add some documentation to the wiki about this.

zengjie:
OK. I will fix the tests. But I don't understand why this would mess up the inheritance tree. The current goto/continue implementation is valid for Lua 5.2/5.3/JIT and should be the 'default' version. Lua 5.1 is the special case for this problem.

Perryvw:
@zengjie We aim to implement versions as described here: https://github.com/Perryvw/TypescriptToLua/wiki/Different-Lua-Versions

The different versions of transpilers are currently chained so we have as little code duplication as possible. If the current transpilers version is implemented in 5.2 it will also be used for later versions.

zengjie:
@Perryvw I have finished the adjustment.

src/targets/Transpiler.51.ts
```diff
@@ -4,4 +4,5 @@ import { TSHelper as tsHelper } from "../TSHelper";
 import * as ts from "typescript";
 
 export class LuaTranspiler51 extends LuaTranspiler {
+
```
Perryvw: unnecessary

yarn.lock
```diff
@@ -0,0 +1,2070 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
Perryvw: don't commit this please (I would be okay with a commit adding this to `.gitignore`)

Janne252:
Is this available in the `0.5.0` release? I'm still getting generated `::__continue0::` labels with `"luaTarget": "5.1"`.

Perryvw:
It was included in the 0.6.0 release 11 days ago.

Janne252:
It appears that the `npm` [version is still `0.5.0`](https://www.npmjs.com/package/typescript-to-lua?activeTab=versions). I suppose I'll have to work with a local clone of the repository until the npm gets updated. Thanks!

PR #126, opened 2018-06-26
Fix unmatched ts/lua function signatures for Map and Set

zengjie:
In TypeScript Map and Set use different function signatures against current implementation:

- `delete()` instead of `remove()`
- `set()` instead of `put()`
- `size` instead of `count()`

### DIFF
--- a/dist/lualib/typescript.lua
+++ b/dist/lualib/typescript.lua
@@ -144,7 +144,7 @@ function Set.constructor(self, other)
 end
 function Set.add(self, item) self._items[item] = true end
 function Set.contains(self, item) return self._items[item] ~= nil end
-function Set.remove(self, item)
+function Set.delete(self, item)
     local contains = Set.contains(self, item)
     self._items[item] = nil
     return contains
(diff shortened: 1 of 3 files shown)


Perryvw:
I didn't realize these classes were so far off. This pull request is a nice start, but I think we should actually take this a step further and exactly implement the interfaces as defined in ES6: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set and https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map

To do this the following has to happen (and maybe I missed some other stuff):
[ ] Implement `clear()`
[ ] Rename `items()` to `entries()`
[ ] Rename `containsKey(key)` to `has(key)`
[ ] Remove `count()` in favor of size.
[ ] Add functional (executing lua) tests.

I can also do part of this, let me know what is needed.

zengjie:
OK. I will take a further step and add tests.

Perryvw:
We actually might already have a couple of tests for map and set in the lualib test file, they should be extended to test all parts of the interface though.

Perryvw:
Closed in favor of #131

Issue #122, opened 2018-06-11
Fix typescript overload


lolleko:
https://www.typescriptlang.org/docs/handbook/functions.html#overloads
TS overloads are basically syntax sugar for `function(x : TypeOne | TypeTwo)`. So there should be no problem when allowing them.

For that we need to ignore funciton declarations.

```
function pickCard(x: {suit: string; card: number; }[]): number; // ignore
function pickCard(x: number): {suit: string; card: number; };   // ignore
function pickCard(x): any {                                     // transpire only functions with body
    // Body....
}
```

Perryvw:
Done in #128

PR #128, opened 2018-06-30
Implemented typeof, instanceof and overloads

Perryvw:
Implemented support for `typeof`, `instanceof` and the ability to deal with overloads for #122. Also found a random unsupported string type that was added.


apemanzilla:
This `typeof` implementation could cause some issues with type narrowing: IIRC, TS expects `typeof` to return `"string" | "number" | "array" | "function" | "boolean" | "object"`, while lua's `type` will return `"table"` instead of `"array"` or `"object"`.

Perryvw:
I don't think array is a thing, I can't find any references to MDN, also `typeof([])` gives me "object" in Chrome. I was discussing with @lolleko wether we should be returning "table" or "object", I think table should be fine as long as we document it properly.

apemanzilla:
I might have misremembered that with `"array"` then. My concern was mostly that if you have `const something: object | number`, `if (typeof something === "table")` won't successfully perform type narrowing. In fact, in recent versions of TS, I think it even produces a compiler error if you try to compare an unknown constant to a `typeof` result.

Perryvw:
Ah yes that is a good point, I will change that, we will probably translate to `type(...) == "table" and "object" or type(...)`

PR #139, opened 2018-07-10
Allow string initializer for enums

zengjie:
String enum is a useful language feature, I think we should support it.

### DIFF
--- a/test/translation/lua/enum.lua
+++ b/test/translation/lua/enum.lua
@@ -2,3 +2,4 @@ TestEnum={}
 TestEnum.val1=0
 TestEnum.val2=2
 TestEnum.val3=3
+TestEnum.val4="bye"
--- a/test/translation/lua/enum.lua
+++ b/test/translation/lua/enum.lua
@@ -2,4 +2,3 @@ TestEnum={}
 TestEnum.val1=0
 TestEnum.val2=2
 TestEnum.val3=3
-TestEnum.val4="bye"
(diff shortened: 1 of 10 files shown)


Perryvw:
What about testcase:
```ts
enum abc {
    a,
    b = "def",
    c,
}

**Comment by zengjie** [comment]
Did you mean create a new test case with this input? or just modify the current one?

**Comment by Perryvw** [comment]
I meant adding a new testcase with that specific enum, it is not allowed in typescript and should error out (which you can test)

**Comment by Perryvw** [comment]
Also you probably shouldn't change old tests, just add some new ones.

**Comment by zengjie** [comment]
Yes. You are right, this test case will result an error.

**Comment by zengjie** [comment]
@Perryvw I've updated the test cases.

File: test/translation/lua/enum.lua [hunk 1]
```diff
@@ -1,4 +1,5 @@
 TestEnum={}
 TestEnum.val1=0
 TestEnum.val2=2
-TestEnum.val3=3
\ No newline at end of file
+TestEnum.val3=3
+TestEnum.val4="bye"
```
Perryvw: Revert this file back to just a generic enum as it was before, this case is covered by enumHeterogeneous.lua

test/translation/ts/enum.ts
```diff
@@ -1,5 +1,6 @@
 enum TestEnum {
     val1 = 0,
     val2 = 2,
-    val3
-}
\ No newline at end of file
+    val3,
+    val4 = "bye"
+}
```
Perryvw: Also revert this file

zengjie:
@Perryvw I've reverted the test case. 😄

PR #141, opened 2018-07-10
Fixed bug where table.unpack was called in Lua 5.1

osyrisrblx:


.vscode/settings.json
```diff
@@ -0,0 +1,15 @@
+{
```
Perryvw: Don't commit this file.

osyrisrblx:
@Perryvw Looks like the current test runner just targets the default Lua. Is there a way to specify a test should be Lua5.1?

Perryvw:
Yes, if you look at expressions.spec.ts you will find a whole bunch of version specific tests:
https://github.com/Perryvw/TypescriptToLua/blob/master/test/unit/expressions.spec.ts#L108-L172

PR #145, opened 2018-07-15
Threaded tests

lolleko:
Added `npm run threaded-tests` will run tests faster, but the test reporting is less detailed.
Also improved CI config to not run tests twice.


test/threaded_runner.ts
```diff
@@ -0,0 +1,53 @@
...
+
+config.set({
+    basepath : {
+        node : __dirname,
+    }
+});
+
+let cpuCount = os.cpus().length + 1;
+if ("TRAVIS" in process.env && "CI" in process.env) {
+    // fixed thread coutn for CI
```
Perryvw: typo

PR #146, opened 2018-07-16
Fixed exports if exported value was changed after declaration

### LINKED ISSUE #143: export little problems..
a.ts
```typescript
export let a;
a = 1;
```
----------------------
a.lua
```lua
local exports = exports or {}
local a = nil
exports.a = a
a = 1
return exports
```
After translate to Lua
a value forever is nil

lolleko:
Fixes #143

### DIFF
--- a/src/Transpiler.ts
+++ b/src/Transpiler.ts
@@ -24,6 +24,12 @@ export enum LuaTarget {
     LuaJIT = "JIT",
 }
 
+interface ExportInfo {
+    name: string | ts.__String;
+    node: ts.Node;
+    dummy: boolean;
+}
+
 export abstract class LuaTranspiler {
     public static AvailableLuaTargets = [LuaTarget.LuaJIT, LuaTarget.Lua53];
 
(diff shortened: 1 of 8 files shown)


osyrisrblx:
I haven't gotten a chance to read over this carefully but I don't think moving the exports to the bottom of the file fixes the problem.

As an example, asynchronous code can change the value of the exported variable but that won't get reflected where the module is imported. Perhaps a better solution would be to change all references of `var` to `exports.var`.

lolleko:
I want to avoid replacing every occurrence with `export.var` at all cost, because that requires one additional hash for each var access and looses the great performance benefit Lua locals have.

I am not even sure this works like this in JS:
Pretty sure, In JS all exported variables (let, var) are copied on import.
For objects and functions a module could indeed update the value in an async way.
I have never seen any code like this, that feels like errors waiting to happen.

Also Lua uses a module cache, I'm not sure if updating an exported value after it is in that cache will update the cache. Therefor not sure if what you are proposing is even possible.

So unless anyone encounters this exact issue and gives me an example of why this would be needed, which I doubt. I am fine with having this limitation.

src/Transpiler.ts
```diff
@@ -145,7 +160,12 @@ export abstract class LuaTranspiler {
 
     // Transpile a block
     public transpileBlock(block: ts.Block): string {
-        return block.statements.map(statement => this.transpileNode(statement)).join("");
+        this.exportStack.push([]);
+        let result = block.statements.map(statement => this.transpileNode(statement)).join("");
+        const exps = this.exportStack.pop();
```
lolleko: Todo change makeExport to makeExport`s`and  pop and create result inside there

PR #158, opened 2018-07-27
Fix variable destructuring compatibility issue with lua 5.1

Janne252:
Fixes:
 - https://github.com/Perryvw/TypescriptToLua/issues/157

Replaces:
- https://github.com/Perryvw/TypescriptToLua/pull/141

I added tests for `luaTarget` **5.1**, **5.2**, **5.3**, and **JIT**. Technically only the tests for 5.1 and 5.2 matter as Transpilers 5.3 and JIT don't override `transpileVariableDestructuring`, so they might be a bit excessive.

Edit: There seems to be an issue reported by `continuous-integration/appveyor/p`, that error didn't occur while running `npm run test-threaded`, only with `npm run test`. Investigating...

### DIFF
--- a/src/Transpiler.ts
+++ b/src/Transpiler.ts
@@ -1287,7 +1287,12 @@ export abstract class LuaTranspiler {
     }
 
     // Implemented in 5.1 and overridden in 5.2 (and onwards)
-    public abstract transpileVariableDestructuring(value: string): string;
+    public transpileVariableDestructuring(value: string): string {
+        throw new TranspileError(
+            `transpileVariableDestructuring must be implemented!`,
+            null
+        );
+    }
 
     public transpileVariableDeclaration(node: ts.VariableDeclaration): string {
         if (ts.isIdentifier(node.name)) {
--- a/src/Transpiler.ts
+++ b/src/Transpiler.ts
@@ -1286,6 +1286,9 @@ export abstract class LuaTranspiler {
         return result;
     }
 
+    // Implemented in 5.1 and overridden in 5.2 (and onwards)
+    public abstract transpileVariableDestructuring(value: string): string;
+
     public transpileVariableDeclaration(node: ts.VariableDeclaration): string {
         if (ts.isIdentifier(node.name)) {
             // Find variable identifier
@@ -1298,7 +1301,6 @@ export abstract class LuaTranspiler {
             }
         } else if (ts.isArrayBindingPattern(node.name)) {
             // Destructuring type
-            const value = this.transpileExpression(node.initializer);
 
             // Disallow ellipsis destruction
             if (node.name.elements.some(elem => !ts.isBindingElement(elem) || elem.dotDotDotToken !== undefined)) {
@@ -1306,12 +1308,13 @@ export abstract class LuaTranspiler {
             }
 
             const vars = node.name.elements.map(e => this.transpileArrayBindingElement(e)).join(",");
+            const value = this.transpileExpression(node.initializer);
 
             // Don't unpack TupleReturn decorated functions
             if (tsHelper.isTupleReturnCall(node.initializer, this.checker)) {
                 return `local ${vars}=${value}\n`;
             } else {
-                return `local ${vars}=table.unpack(${value})\n`;
+                return `local ${vars}=${this.transpileVariableDestructuring(value)}\n`;
             }
         } else {
             throw new TranspileError(
--- a/src/targets/Transpiler.52.ts
+++ b/src/targets/Transpiler.52.ts
@@ -53,4 +53,8 @@ export class LuaTranspiler52 extends LuaTranspiler51 {
                 return `bit32.arshift(${lhs},${rhs})`;
         }
     }
+    /** @override */
+    public transpileVariableDestructuring(value: string): string {
+        return `table.unpack(${value})`;
+    }
 }
--- a/src/targets/Transpiler.52.ts
+++ b/src/targets/Transpiler.52.ts
@@ -53,6 +53,7 @@ export class LuaTranspiler52 extends LuaTranspiler51 {
                 return `bit32.arshift(${lhs},${rhs})`;
         }
     }
+
     /** @override */
     public transpileVariableDestructuring(value: string): string {
         return `table.unpack(${value})`;
--- a/test/src/util.ts
+++ b/test/src/util.ts
@@ -12,6 +12,7 @@ import {lauxlib, lua, lualib, to_jsstring, to_luastring } from "fengari";
 const fs = require("fs");
 
 const libSource = fs.readFileSync(path.join(path.dirname(require.resolve("typescript")), "lib.es6.d.ts")).toString();
+const backport51 = fs.readFileSync("test/src/backport.51.lua") + "\n";
 
 export function transpileString(str: string, options: CompilerOptions = { luaLibImport: "none", luaTarget: LuaTarget.Lua53 }): string {
     const compilerHost = {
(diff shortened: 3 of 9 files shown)


Janne252:
Looks like [Fengari](https://github.com/fengari-lua/fengari) only implements lua 5.3, which causes an issue when testing 5.1 specific features that have been deprecated in 5.3. 

I added `with51Backport` to `executeLua` which includes the necessary backport
```
-- Lua 5.1 unpack was moved to table.unpack in Lua 5.3, make it available as a backport
unpack = table.unpack
```

However this pollutes the API of `executeLua` with a new boolean param. Thoughts?

Janne252:
Implemented the requested changes (removed duplicate tests).

src/targets/Transpiler.52.ts
```diff
@@ -53,4 +53,8 @@ export class LuaTranspiler52 extends LuaTranspiler51 {
                 return `bit32.arshift(${lhs},${rhs})`;
         }
     }
+    /** @override */
```
Perryvw: Add a whiteline between methods.

src/Transpiler.ts
```diff
@@ -1286,6 +1286,9 @@ export abstract class LuaTranspiler {
         return result;
     }
 
+    // Implemented in 5.1 and overridden in 5.2 (and onwards)
+    public abstract transpileVariableDestructuring(value: string): string;
```
Perryvw: By convention we throw a TranspileError here, see `transpileContinue` in Transpiler.ts.

test/unit/assignmentDestructuring.spec.ts
```diff
@@ -0,0 +1,31 @@
+import { Expect, Test, TestCase } from "alsatian";
+import * as util from "../src/util";
+import { LuaTarget } from "../../src/Transpiler";
+
+export class AssignmentDestructuringTests {
+    
+    private readonly assignmentDestruturingTs = `
+    declare function myFunc(): [number, string];
```
Perryvw: Indent line 8 and 9 please.

Janne252: Almost all test files have issues reported by `tslint`. I'm using VS Code and the extension seems to also apply the lint to the test files. I think `./tests/**/*.ts` should be ignored by `tslint`?

**tslint.json**
```js
{
/// ...
    "linterOptions": {
        "exclude": [
            "./test/**/*.ts"
        ]
    },
// ...
}
```

Perryvw: Eventually they should adhere to tslint I think, but for now I am treating tslint as "suggestions" for test files (which is why it doesn't fail builds).

test/src/util.ts
```diff
@@ -60,11 +61,15 @@ export function transpileFile(filePath: string): string {
     return result.trim();
 }
 
-export function executeLua(luaStr: string, withLib = true): any {
+export function executeLua(luaStr: string, withLib = true, with51Backport = true): any {
     if (withLib) {
         luaStr = minimalTestLib + luaStr;
     }
 
+    if (with51Backport) {
```
Perryvw: Can this backport stuff be removed since no 5.1 lua is executed?

test/unit/assignmentDestructuring.spec.ts
Janne252: Is it okay to disable `tslint` for `./test` for now (as demonstrated above)? I'm using VS Code and it's rather obtrusive to have the test files full of errors. Unless a path's severity can be forced to "suggestion"?

Perryvw: As long as you don't push it you can change it however you want :D

test/src/util.ts
Janne252: When luaTarget is 5.1 and a variable destructuring is performed, e.g.
```ts
declare function myFunc(): [number, string];
let [a, b] = myFunc();
```
It has to rersult in
```lua
local a,b=unpack(myFunc())
```
Because Lua 5.1 doesn't have `table.unpack`, which is generated by 5.2 and onwards.

The tests that run Lua code rely on [Fengari](https://github.com/fengari-lua/fengari) which only implements Lua 5.3 runtime. `unpack` (that was generated by luaTarget 5.1) is not defined in Lua 5.3 and thus has to be backported in order to be able to test Lua 5.1 code in Lua 5.3 VM.

Perryvw: Which tests are executed in Fengari that are compiled with target 5.1?

Janne252: - [@Test("loop versions")](https://github.com/Janne252/TypescriptToLua/blob/3da4f7e96e53fadbe328a5cc36af1bc7a291dd91/test/unit/loops.spec.ts#L324)
- [ @Test("Bitop [5.1]")](https://github.com/Janne252/TypescriptToLua/blob/3da4f7e96e53fadbe328a5cc36af1bc7a291dd91/test/unit/expressions.spec.ts#L107)
- [@Test("Assignment destructuring [5.1]")](https://github.com/Janne252/TypescriptToLua/blob/3da4f7e96e53fadbe328a5cc36af1bc7a291dd91/test/unit/assignmentDestructuring.spec.ts#L11)

Perryvw: Those are only transpiled, not executed.

Janne252: My bad. Total of 14 tests fail without the backport:

```
FAIL: map iterable constructor
FAIL: map iterable constructor map
FAIL: map clear
FAIL: map delete
FAIL: map entries
FAIL: map foreach
FAIL: map foreach keys
FAIL: map get
FAIL: map get missing
FAIL: map has
FAIL: map has null
FAIL: map keys
FAIL: map values
FAIL: map size
```

They all fail for the same reason:
```
Error: LUA ERROR: [string "function __TS__ArrayEvery(arr,callbackfn)..."]:268: attempt to call a nil value (global 'unpack')
```

Perryvw: Hmm those are not transpiled as 5.1 either... Sounds like something is off here, I will investigate.

Janne252: It's caused by build_lualib.ts, the lualib is transpiled as 5.1. For example the generated map.lua uses unpack call:
````lua
function Map.constructor(self,other)
    self.items = {}
    self.size = 0
    if __TS__InstanceOf(other, Map) then
        self.size = other.size
        for _, __forOfValue0 in ipairs(other:entries()) do
        local key,value=unpack(__forOfValue0)
            do
                self.items[key] = value
            end
        end
    else
        if other~=nil then
            self.size = #other
            for _, __forOfValue1 in ipairs(other) do
            local key,value=unpack(__forOfValue1)
                do
                    self.items[key] = value
                end
            end
        end
    end
end
```

Perryvw: Does changing the lua version to 5.3 in build_lualib.ts fix that? I am fine with you changing that for now so we can get rid of this backport stuff, since that is just treating symptoms instead of the actual cause.

We will have to come up with a better fix later, probably changing the map implementation so it does not use unpack (or any version-sensitive stuff) would be the best answer, but we can do that in a different PR.

Janne252: If the luaTarget for build_lualib.ts is changed to 5.3, then lualib would not work with Lua 5.1, as it would contain `table.unpack` (at the moment). If we don't want to use backports, the other option is to add a polyfill so that transpiled 5.1 code supports `table.unpack`, e.g.

```lua
if not table.unpack then
    function table.unpack(...)
        return unpack(...)
    end
end
```

This polyfill should probably be included in the lualib, but ideally it would only be emitted when the project luaLib target is set to 5.1.

Perryvw: I acutally just hotfixed unpack out of the map implementation, if you merge master into your branch that should hopefully fix the failing tests.

Janne252:
Yep, tests passed without the backport (removed it). Also applied all the requested changes.

Issue #155, opened 2018-07-23
Extend lualib import options


lolleko:
Additional options:

- `inlineLocal` inlines the function add the top of a file as local variables (maybe change inline to inlineGlobal when adding this)
- `requireSingle` (maybe use a better name for this): requires each feature via a separate file.

lolleko:
Add manual import option (comment in #71)

Janne252:
I think the inlined functions should be placed at the top of the file in case the file uses the features (e.g. ternary operation) on module level (not inside a scope), as demonstrated below:

```lua
local value = SomeFancyInlinedFunction();

-- stuff

function SomeFancyInlinedFunction()
    
end
```

Results in `attempt to call global 'SomeFancyInlinedFunction' (a nil value)` because `SomeFancyInlinedFunction` is referenced before its declaration. Lua doesn't hoist variables like JavaScript does!

lolleko:
Superseded  by #653

PR #159, opened 2018-07-28
Inlining of map and set and detection when they are used

### LINKED ISSUE #154: Add Map/Set to lualibfeatureset if used
Possible places to check if map/set is used: 
* `transpileNew`: this might not work if map/set is used in an external library that doesn't do inlining.
* `transpileVaraibleDeclaration` will this cover all cases?
* `transpileNew` + object access: (function call or property access to Maps/Sets) should guarantee that Map/Set is inlined when used somewhere

We should probably also implement it in a way we can easily add other JS internal classes.

Perryvw:
Detect when Map and Set are used and inline them.

Set test lualibimport to "require" by default.

Require now only adds the require statement if any lualib features are detected.

Added an "always" option for luaLibImport to always add the require to the lualib bundle.

Also some general fixes to the lualib imports and inlining.
Closes #154.

Fixes an issue with inlining order as mentioned in #155.

### DIFF
--- a/src/Transpiler.ts
+++ b/src/Transpiler.ts
@@ -44,9 +44,10 @@ export enum LuaLibFeature {
 }
 
 export enum LuaLibImportKind {
+    None = "none",
+    Always = "always",
     Inline = "inline",
     Require = "require",
-    None = "none",
 }
 
 interface ExportInfo {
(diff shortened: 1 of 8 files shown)


src/Transpiler.ts
```diff
@@ -1673,4 +1690,17 @@ export abstract class LuaTranspiler {
 
         return result;
     }
+
+    public checkForLuaLibType(type: ts.Type): void {
```
lolleko: The naming of this function seems a bit of to me. As it checks for the time, but also adds the feature.

PR #160, opened 2018-07-28
Support for single-character escape sequences

Janne252:
With this update single-character escape sequences will be preserved, e.g.
```ts
let escapedCharsInQuotes = '\\ \0 \b \t \n \v \f \" \' \`';
```
Is transpiled to
```lua
local escapedCharsInQuotes = "\\ \0 \b \t \n \v \f \" \' \`"
```

Previously you had to double escape them, for example `"\n"` would have resulted in an actual linebreak in the transpiled lua code.

### DIFF
--- a/src/Transpiler.ts
+++ b/src/Transpiler.ts
@@ -678,6 +678,30 @@ export abstract class LuaTranspiler {
         }
     }
 
+    public transpileStringLiteralExpression(node: ts.Node): string {
+        // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String
+        const escapeSequences: Array<[RegExp, string]> = [
+            [/[\\]/g, "\\\\"],
+            [/[\']/g, "\\\'"],
+            [/[\`]/g, "\\\`"],
+            [/[\"]/g, "\\\""],
+            [/[\n]/g, "\\n"],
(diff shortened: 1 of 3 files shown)


src/Transpiler.ts
```diff
@@ -691,6 +691,30 @@ export abstract class LuaTranspiler {
         }
     }
 
+    public transpileStringLiteralExpression(node: ts.Node): string {
```
Perryvw: I would suggest refactoring to `escapeString(str: string): string` and just calling this wherever suitable.

src/Transpiler.ts
```diff
@@ -714,8 +738,7 @@ export abstract class LuaTranspiler {
                 return this.transpileIdentifier(node as ts.Identifier);
             case ts.SyntaxKind.StringLiteral:
             case ts.SyntaxKind.NoSubstitutionTemplateLiteral:
-                const text = (node as ts.StringLiteral).text;
```
Perryvw: Then this would just be something like `const text = this.escapeString((node as ts.StringLiteral).text);`

Janne252:
I realized the missing support for template strings as well, wasn't quick enough to ninja-commit it though.

Support & tests added for template strings and requested changes applied.

Issue #153, opened 2018-07-20
Unused code after return causes syntax error


lolleko:
TS:
```
function test(): number
  return 12;
  print("test");
end
```
Lua:
```
function test()
  return 12
  print("test") <----- "Expected end"
end
```

PR #161, opened 2018-07-28
Stop block transpilation after encountering a return statement

Perryvw:
Addresses #153

### DIFF
--- a/src/Transpiler.ts
+++ b/src/Transpiler.ts
@@ -597,13 +597,14 @@ export abstract class LuaTranspiler {
             this.pushIndent();
 
             this.transpilingSwitch++;
-            clause.statements.forEach(statement => {
-                result += this.transpileNode(statement);
-            });
+            result += this.transpileBlock(ts.createBlock(clause.statements));
             this.transpilingSwitch--;
 
             let i = index + 1;
-            if (i < clauses.length && !tsHelper.containsStatement(clause.statements, ts.SyntaxKind.BreakStatement)) {
+            if (i < clauses.length
+                && !tsHelper.containsStatement(clause.statements, ts.SyntaxKind.BreakStatement)
+                && !tsHelper.containsStatement(clause.statements, ts.SyntaxKind.ReturnStatement)
+            ) {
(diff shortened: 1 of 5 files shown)


src/Transpiler.ts
```diff
@@ -224,7 +224,17 @@ export abstract class LuaTranspiler {
...
         this.exportStack.push([]);
-        let result = block.statements.map(statement => this.transpileNode(statement)).join("");
+
+        let result = "";
+        for (const statement of block.statements) {
+            result += this.transpileNode(statement);
+
+            // Don't transpile any dead code after a return
+            if (ts.isReturnStatement(statement)) {
+                break;
```
lolleko: Please no breaks.

test/unit/functions.spec.ts
```diff
@@ -204,4 +204,20 @@ export class FunctionTests {
...
+        Expect(result).toBe(3);
+    }
+
+    @Test("Method dead code after return")
+    public methodDeadCodeAfterReturn(): void {
+        const result = util.transpileAndExecute(
+            `class def { public static abc() { return 3; const a = 5; } } return def.abc();`);
+
+        Expect(result).toBe(3);
+    }
```
lolleko: namespaces also use transpire block a test would be nice.

test/unit/conditionals.spec.ts
```diff
@@ -251,4 +251,12 @@ export class LuaConditionalsTests {
...
         Expect(result).toBe(expected);
     }
+
+    @Test("If dead code after return")
+    public ifDeadCodeAfterReturn(): void {
+        const result = util.transpileAndExecute(
+            `if (true) { return 3; const b = 8; }`);
+
+        Expect(result).toBe(3);
+    }
```
lolleko: switch also use transpileblock irrc, but I think this is less important than the namespace test and probably not needed.

test/unit/functions.spec.ts
Perryvw: You can not return from a namespace since it is not a function body.

PR #164, opened 2018-07-29
Support for unary expressions in assignments

Janne252:
**TS**
```ts
let count = 0;
count++;
--count;

let count01 = count++;
let count02 = --count;

// Self-referencing operations
let count03 = count02 + count02++;
let count04 = count02 + --count02;
// Non-self-referencing
let count05 = count02 + ++count03;
```
**Lua**
```lua
local count = 0

count = count+1
count = count-1
local count01 = (function() local count_before = count; count = count+1; return count_before end)()

local count02 = (function() count = count-1; return count end)()

-- Self-referencing operations
local count03 = (function() return count02 end)()+(function() local count02_before = count02; count02 = count02+1; return count02_before end)()
local count04 = (function() return count02 end)()+(function() count02 = count02-1; return count02 end)()
--  Non-self-referencing
local count05 = count02+(function() count03 = count03+1; return count03 end)()
```

### Note!

Does not add support for chained assignments, e.g. `let count = count03 += 3` . (has to be implemented at some point, hopefully)

### DIFF
--- a/src/TSHelper.ts
+++ b/src/TSHelper.ts
@@ -190,4 +190,12 @@ export class TSHelper {
 
         return [false, null];
     }
+
+    public static isUnaryAssignmentExpression(node: ts.Node): boolean {
+        return (
+            (ts.isPrefixUnaryExpression(node) || ts.isPostfixUnaryExpression(node)) &&
+            node.parent &&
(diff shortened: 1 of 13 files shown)


Perryvw:
Replaced by #166

PR #166, opened 2018-08-05
Added support for assignments inside expressions

Perryvw:


src/Transpiler.ts
```diff
@@ -952,10 +949,36 @@ export abstract class LuaTranspiler {
     public transpilePostfixUnaryExpression(node: ts.PostfixUnaryExpression): string {
         const operand = this.transpileExpression(node.operand, true);
         switch (node.operator) {
-            case ts.SyntaxKind.PlusPlusToken:
-                return `${operand}=${operand}+1`;
-            case ts.SyntaxKind.MinusMinusToken:
-                return `${operand}=${operand}-1`;
+            case ts.SyntaxKind.PlusPlusToken: {
+                const increment = ts.createBinary(node.operand, ts.SyntaxKind.PlusToken, ts.createLiteral(1));
+                const oldValueName = `__originalValue${this.genVarCounter++}`;
```
Janne252: Technically this variable name can still collide with the operand name because it doesn't depend on its name:

```lua
local __originalValue1 = 1

-- Transpiled from postifx ++
local c = (function() local __originalValue1 = __originalValue1; __originalValue1 = __originalValue1 + 1; return __originalValue1  end)()

-- Expected to be 2, 1
print(__originalValue1, c)

-- vs

local __originalValue1 = 1

-- Transpiled from postifx ++
local c = (function() local __originalValue1_before = __originalValue1; __originalValue1 = __originalValue1 + 1; return __originalValue1_before end)()

-- Expected to be 2, 1
print(__originalValue1, c)
```
Yes, it's a rare edge case but since it can be properly protected against, I think it should be be based on the name of the operand, e.g. `` `${operandName}_before` ``. Feel free to execute the lua above to see it in action (if it collides, the temp var re-declares it locally and thus overrides it)

PR #168, opened 2018-08-14
Add support for block statements

nucleartide:
I hacked in a quick change to support block statements. Lemme know if I should add tests!

TS:

```ts
{
  const a = 1 + 2
  do_something(a)
}
```

Lua:

```lua
do
  local a = 1 + 2
  do_something(a)
end
```

### DIFF
--- a/src/Transpiler.ts
+++ b/src/Transpiler.ts
@@ -250,7 +250,7 @@ export abstract class LuaTranspiler {
             }
         }
 
-        result += "end";
+        result += "end\n";
         return result;
     }
 
--- a/src/Transpiler.ts
+++ b/src/Transpiler.ts
@@ -240,6 +240,20 @@ export abstract class LuaTranspiler {
         return result;
     }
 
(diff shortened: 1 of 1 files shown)


src/Transpiler.ts
```diff
@@ -240,6 +240,20 @@ export abstract class LuaTranspiler {
         return result;
     }
 
+    // Transpile a block statement
+    public transpileBlockStatement(block: ts.Block): string {
```
lolleko: Why not use the existing `transpileBlock(block: ts.Block): string`?

lolleko: To clarify you could replace line 246 - 256 with `this.transpileblock(block);`

nucleartide:
@lolleko Awesome, thanks for the quick turnaround! I'll take a look at refactoring & adding tests later this week – currently focused on building out this game. :)

Perryvw:
No longer needed after #199

Issue #174, opened 2018-08-19
Watch mode is only partially incremental


lolleko:
The current watch mode is only partially incremental.

Typechecking handled by the ts compiler API is incremental (there is a small chance that this is not incremental aswell)

Emitting files (walking and transpiling the AST) is not incremental. Currently all files get reemitted on a file change.

This is kind of hard to fix because the compiler API implements incremental builds by hashing the emitted `.js` files. We don't generate those, so we either need to roll our own solution in the future or hack it.

ark120202:
Resolved in #488

PR #175, opened 2018-08-19
Added experimental watch mode

lolleko:
Added support for tsc flags `-w`/`--watch`

The watch mode is currently not really incremental and will emit the whole project if a file is changed.
See #174


src/Compiler.ts
```diff
@@ -11,42 +11,96 @@ import { LuaLibImportKind, LuaTarget, LuaTranspiler, TranspileError } from "./Tr
...
+    } else {
+        host = ts.createWatchCompilerHost(
+            fileNames,
+            options,
+            ts.sys,
+            ts.createSemanticDiagnosticsBuilderProgram
+        );
+    }
+
+    const errorDiagnostic: ts.Diagnostic = {
```
Perryvw: I would prefer you create this and immediately fill in the correct values in `afterProgramCreate` since it's not used anywhere else.

src/Compiler.ts
```diff
@@ -57,7 +111,7 @@ export function compileFilesWithOptions(fileNames: string[], options: CompilerOp
                 let outPath = sourceFile.fileName;
                 if (options.outDir !== options.rootDir) {
                     const relativeSourcePath = path.resolve(sourceFile.fileName)
-                                                   .replace(path.resolve(rootDir), "");
+                                                .replace(path.resolve(rootDir), "");
```
Perryvw: You sure this should be gone? I like the old version better.

lolleko: auto format /shrug

PR #181, opened 2018-08-23
Casting ambiguous syntax

### LINKED ISSUE #191: ambiguous syntax (function call x new statement) near '('
```
(this.something as any).Initialize();
```

=>

```
(self.something ):Initialize()
```

Throws `ambiguous syntax (function call x new statement) near '('` at lua runtime.

andreiradu:
using extra paranthesis in ts causes code like `dostuff() (self.member):method();` calls, which can be ambiguous depending on context. I've added ; at the end of statements to help disambiguate. 

Closes #191


src/Transpiler.ts
```diff
@@ -1421,9 +1421,9 @@ export abstract class LuaTranspiler {
             const identifierName = this.transpileIdentifier(node.name);
             if (node.initializer) {
                 const value = this.transpileExpression(node.initializer);
-                return `local ${identifierName} = ${value}\n`;
+                return `local ${identifierName} = ${value};\n`;
```
Perryvw: You pushed these too far down, I suggest adding the ';' in the `transpileNode` function, like you did for ExpressionStatement.

andreiradu: That was my first approach, but `transpileVariableDeclaration` returns a newline, so the transpiled code would look like
```
      local a = 0
;
      foo()
```

Perryvw: I think that's a bug, if removing the `\n` from transpileVariableDeclaration does not break any (non-whitespace) tests I am fine with you removing it there.

lolleko: That new line after variable declaration is definitely unintentional and should be removed.

src/Transpiler.ts
```diff
@@ -497,7 +497,7 @@ export abstract class LuaTranspiler {
         // Add header
         let result = "";
         for (const variableDeclaration of (node.initializer as ts.VariableDeclarationList).declarations) {
-            result += this.indent + this.transpileVariableDeclaration(variableDeclaration);
+            result += this.indent + this.transpileVariableDeclaration(variableDeclaration) + "\n";
```
Perryvw: Should there also be a `;` here?

andreiradu: Not really, you know a while will always follow, so there's no ambiguity. I've tried not to add ; where they weren't needed.

andreiradu:
Merged master and fixed tests

PR #182, opened 2018-08-23
-added support for array.pop

andreiradu:
-only issue I see is that this is an inline implementation. So you can't hold Array.Pop in a function. Not sure if that matters to anyone.


lolleko:
I could have sworn we already had pop implemented. Looks fine to me. We can always move it to an extra function later if needed.

PR #183, opened 2018-08-26
Support shorthand property assignment in object literals

gakada:


gakada:
Seems like travis is happy, though locally I have
```
Pass: 464/578
Fail: 114/578
```
because of https://github.com/Perryvw/TypescriptToLua/issues/186
```
Error: LUA ERROR: error loading module 'lualib_bundle' from file './lualib_bundle.lua':./lualib_bundle.lua:436: <eof> expected near 'function'
```

Perryvw:
I would still like a test `const myfield = 3; const myobject = {myfield}; return myobject.myfield` expecting 3. After that I am fine with merging this.

gakada:
Oh there is `executeLua` as well. Where should I add it?

gakada:
Added in ObjectLiteralTests as well since it is not exactly an assignment.

Perryvw:
`test/unit/objectLiteral.spec.ts` is fine.

PR #189, opened 2018-08-26
Escape reserved identifier instead of throwing an exception

### LINKED ISSUE #188: Mixing typeof and parameters named type
`typeof` is translated to `type`, if there is parameter named `type` it is not renamed. For example
```ts
const f = (type: number | ((x: number) => number)): number => (typeof type === 'number' ? type : type(42))
```
gives
```lua
local f = function(type) return (__TS__Ternary((type(type) == "table" and "object" or type(type))=="number", function() return type end, function() return type(42) end)) end
```
which won't work.

Perryvw:
Instead of throwing an error when using lua keywords as identifiers, we just escape them with some underscores so code does not unnecessarily breaks.

Should fix #188

### DIFF
--- a/src/Errors.ts
+++ b/src/Errors.ts
@@ -40,9 +40,6 @@ export class TSTLErrors {
     public static InvalidThrowExpression = (node: ts.Node) =>
         new TranspileError(`Invalid throw expression, only strings can be thrown.`, node)
 
-    public static KeywordIdentifier = (node: ts.Identifier) =>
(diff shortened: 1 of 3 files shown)


andreiradu:
Something worth noting is that changing variable names when transpiling makes debugging harder. Right now the only thing you need to have a lua debugger work on the ts code is line mappings.

Perryvw:
After discussing this with @lolleko we decided this is probably not the right solution here, closing this PR for now.

Issue #196, opened 2018-09-20
/** !TupleReturn */ does not destructure class return parameters on lua implementation


hazzard993:
Code
```ts
class X {
	
	/** !TupleReturn */
	tuple(): [number, number, number] {
		return [3, 4, 5]
	}

	/** !TupleReturn */
	static staticTuple(): [number, number, number] {
		return [1, 2, 3]
	}

}

/** !TupleReturn */
function tuple() {
	return [1, 2, 3]
}

let x = new X()
let [a, b, c] = X.staticTuple()
let [d, e, f] = x.tuple()
let [g, h, i] = tuple()
print(a, b, c)
print(d, e, f)
print(g, h, i)
```
Produces
```lua
X = X or {}
X.__index = X
function X.new(construct, ...)
    local instance = setmetatable({}, X)
    if construct and X.constructor then X.constructor(instance, ...) end
    return instance
end
function X.constructor(self)
end
function X.tuple(self)
    return {3,4,5} -- UNEXPECTED
end
function X.staticTuple(self)
    return {1,2,3} -- UNEXPECTED
end
function tuple()
    return 1,2,3 -- EXPECTED
end
local x = X.new(true)

local a,b,c=X:staticTuple() -- EXPECTED

local d,e,f=x:tuple() -- EXPECTED

local g,h,i=tuple() -- EXPECTED

print(a,b,c)
print(d,e,f)
print(g,h,i)
```
Note the `-- comments`

The values of a `!TupleReturn` method are received correctly but they are not sent correctly from class functions.

`class A { tuple() { return [1, 2, 3] } ==>  ...tuple() { return {1,2,3} -- should be 1, 2, 3 }...`

`tstl --version ==> 0.7.0`

tsconfig.json = `{}`

lolleko:
Seems like `https://github.com/Perryvw/TypescriptToLua/blob/master/src/Transpiler.ts#L684` doesn't work for class functions?

PR #197, opened 2018-09-21
Method declarations now use custom decorators

hazzard993:
For #196 

The following code
https://github.com/Perryvw/TypescriptToLua/blob/50e6cb59691daff5dad40613fca332e78eaad754/src/Transpiler.ts#L688
Only picks up function declarations, not method declarations.

Adding `|| tsHelper.findFirstNodeAbove(node, ts.isMethodDeclaration)` to the end allows it to pick up static and dynamic methods as well.

This allows the `!TupleReturn` decorator to work on class methods

### DIFF
--- a/src/Transpiler.ts
+++ b/src/Transpiler.ts
@@ -685,7 +685,8 @@ export abstract class LuaTranspiler {
         if (node.expression) {
             // If parent function is a TupleReturn function
             // and return expression is an array literal, leave out brackets.
-            const declaration = tsHelper.findFirstNodeAbove(node, ts.isFunctionDeclaration);
+            const declaration = tsHelper.findFirstNodeAbove(node, ts.isFunctionDeclaration)
+                || tsHelper.findFirstNodeAbove(node, ts.isMethodDeclaration);
             let isTupleReturn = false;
             if (declaration) {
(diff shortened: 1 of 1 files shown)


hazzard993:
Thanks for the info @lolleko! I've added some tests to https://github.com/Perryvw/TypescriptToLua/blob/master/test/unit/tuples.spec.ts similar to the example in #196

src/Transpiler.ts
```diff
@@ -685,7 +685,8 @@ export abstract class LuaTranspiler {
         if (node.expression) {
             // If parent function is a TupleReturn function
             // and return expression is an array literal, leave out brackets.
-            const declaration = tsHelper.findFirstNodeAbove(node, ts.isFunctionDeclaration);
+            const declaration = tsHelper.findFirstNodeAbove(node, ts.isFunctionDeclaration)
```
Perryvw: I would prefer if we only did the search once, I think `tsHelper.findFirstNodeAbove(node, n => ts.isFunctionDeclarations(n) || ts.isMethodDeclaration(n))` should be fine.

hazzard993: Thanks and good idea, added another commit to do this search once

Perryvw: Looks good!

PR #198, opened 2018-09-22
-changed toString calls to transpile to tostring()

andreiradu:
Also changed the method declaration to transpile to __tostring

### DIFF
--- a/src/Transpiler.ts
+++ b/src/Transpiler.ts
@@ -1127,10 +1127,14 @@ export abstract class LuaTranspiler {
         } else {
             // Replace last . with : here
             const name = this.transpileIdentifier(node.expression.name);
-            callPath =
+            if (name === "toString") {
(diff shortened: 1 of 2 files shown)


src/Transpiler.ts
```diff
@@ -1513,7 +1517,10 @@ export abstract class LuaTranspiler {
 
         let result = "";
         const identifier = node.name as ts.Identifier;
-        const methodName = this.transpileIdentifier(identifier);
+        let methodName = this.transpileIdentifier(identifier);
+        if (methodName === "toString") {
+            methodName = "__tostring";
```
Perryvw: I must be blind, but how does it know to escape when calling this method?

andreiradu: I don't really understand the question. When calling tostring(table), the __tostring metamethod of the table's metatable is invoked.

Perryvw: Aah gotcha

PR #210, opened 2018-10-03
Declaration Merging Fix

### LINKED ISSUE #207: Declaration merging breaks colon syntax
```ts
class Foo
{
    public method() {}
}

namespace Foo
{
    export interface Bar {}
}

let foo = new Foo();
foo.method();
```
Produces:
```lua
Foo = Foo or {}
Foo.__index = Foo
function Foo.new(construct, ...)
    local instance = setmetatable({}, Foo)
    if construct and Foo.constructor then Foo.constructor(instance, ...) end
    return instance
end
function Foo.constructor(self)
end
function Foo.method(self)
end
Foo = Foo or {}
do
end
local foo = Foo.new(true);
foo.method(); --Should use ":"
```

Without the namespace, the call to "method" will use a colon, but with it, only the dot will be used, so self is not passed.

tomblind:
Modified logic determining if a method call should use dot instead of colon so that calls to methods on class types that are merged with a namespace will correctly use a colon.

Note that the separate check for property-style functions is also covered by this alternate logic so I removed it.

fixes #207

### DIFF
--- a/src/Transpiler.ts
+++ b/src/Transpiler.ts
@@ -1131,7 +1131,9 @@ export abstract class LuaTranspiler {
         // Get the type of the function
         const functionType = this.checker.getTypeAtLocation(node.expression);
         // Don't replace . with : for namespaces or functions defined as properties with lambdas
-        if (functionType.symbol && !(functionType.symbol.flags & ts.SymbolFlags.Method)) {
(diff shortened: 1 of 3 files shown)


tomblind:
No problem - test added.

tomblind:
Turns out this fix missed a case: methods called on 'this' inside other methods would not have the 'Method' flag set, so I added an additional explicit check for this and updated the test to be more exhaustive.

PR #211, opened 2018-10-04
Super calls use self

### LINKED ISSUE #204: Super function being invoked with class, not instance
The following code
```ts
class Parent {
	protected value: number
	constructor() {
		this.value = 10
	}
	func() {
		console.log(this.value)
	}
}
class Child extends Parent {
	constructor() {
		super()
	}
	func() {
		super.func()
		console.log(this.value)
	}
}
let c = new Child()
c.func()
```
Should print `10` twice. But instead prints `nil` followed by `10`

When using `tsc` (and changing `print` to `console.log`) the resulting script outputs `10` twice as expected

This is because the lua it is transpiled to turns `super.func()` into `self.__base:func()`

This means that `Parent#func()` is called with `this == Parent` instead of `this == c`

This happens for all `super` calls except for `super()`

A change in transpilation behavior might have to be made specifically for `super` calls

If `super.func()` became `self.__base.func(self)` the script would work as expected

hazzard993:
Fixes #204 
Added some test cases to `class.spec.ts` testing `super` invocation.

### DIFF
--- a/src/Transpiler.ts
+++ b/src/Transpiler.ts
@@ -1137,6 +1137,10 @@ export abstract class LuaTranspiler {
             callPath = this.transpileExpression(node.expression);
             params = this.transpileArguments(node.arguments);
             return `${callPath}(${params})`;
+        } else if (node.expression.expression.kind === ts.SyntaxKind.SuperKeyword) {
+            // Super calls take the format of super.call(self,...)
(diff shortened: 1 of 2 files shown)


Perryvw:
Thanks, looks good!

PR #220, opened 2018-10-07
Recursion fix

### LINKED ISSUE #187: Recursion doesn't work with arrow functions
Input
```ts
const f = (i: number): number => (i <= 1 ? i : i * f(i - 1))
```
Output
```lua
local f = function(i) return (__TS__Ternary(i<=1, function() return i end, function() return i*f(i-1) end)) end
```

tomblind:
Separated variable assignment from declaration when assigning to function expression or arrow function to allow recursion.

fixes #187


Perryvw:
Could you add a functional test (test/unit) that executes a recursive function?

tomblind:
Good idea - I just added a couple.

Issue #221, opened 2018-10-07
Issues with non-lua operators on properties of objects


tomblind:
Operators such as '+=', '++', etc... can be broken easily right now.

Example 1:
```ts
declare function foo(): number;
declare const a: number[];
a[foo()]++;
```
=>
```lua
a[(foo())+1] = (a[(foo())+1]+1);
```
Here foo() will be called multiple times, producing possibly different values each time or having other undesired side-effects.

Example 2:
```ts
declare function bar(): number[];
let f = bar()[0]++;
```
=>
```lua
local f = (function()
    local __originalValuebar()[(0)+1]0 = bar()[(0)+1];
    bar()[(0)+1] = (bar()[(0)+1]+1);
    return __originalValuebar()[(0)+1]0
end
)();
```
This one won't even compile in lua.

These kinds of issues could be solved by detecting if the value being operated on is a property and caching the object and index parts before performing the operation. This could be wrapped in specialized polyfill functions like the ternary operator is:
```lua
function __TS__PropertyPostInc(o, i)
    local original = o[i]
    o[i] = o[i] + 1
    return original
end

--Example 1
__TS__PropertyPostInc(a, foo())

--Example 2
local f = __TS__PropertyPostInc(bar(), 0)
```

Perryvw:
Looks like a tricky issue indeed. We would have to polyfill all(?) unary and [binary assignment operators](https://github.com/Perryvw/TypescriptToLua/blob/master/src/TSHelper.ts#L142) to have this work properly? I feel like it might be nicer to find an inline solution if possible, but I will have to think about this for a bit.

tomblind:
Yes, all of them would have to be covered. Using polyfills is more efficient however, since the inline solution requires generating a temp lambda every time (at least for expressions).

tomblind:
Of course, to be clear, it is possible to do this inline as well (just less efficient). I definitely can understand the desire to avoid excessive polyfills.

```lua
(function(o, i) local original = o[i]; o[i] = original + 1; return original)(a, foo())

local f = (function(o, i) local original = o[i]; o[i] = original + 1; return original)(bar(), 0)
```

lolleko:
why do you need that extra function call? couldnt we just use:
```
a[foo()]++;
````

```
local ___propertyIndex1 = foo(); a[___propertyIndex1 ] = a[___propertyIndex1 1] + 1;
```
and to make `let f = bar()[0]++;` work extend it further:
```
[...]

PR #230, opened 2018-10-10
fix for translation of assignment expressions

### LINKED ISSUE #225: Problems with assignments as expressions
Similar to #221, evaluating assignments as expression has some issues.

Example:
```ts
declare function a(): number[];
declare function b(): number;
declare function c(): number;
declare const d: number[];
let e = ( a()[b()] = c() );
```
Currently produces:
```lua
local e = ((function() a()[(b())+1] = c(); return a()[(b())+1] end)());
```
The problem here being that a and b are called twice having potentially undesired effects.

A solution could be to cache with a temp:
```lua
local e = ((function() local __temp = c(); a()[(b())+1] = __temp; return __temp end)());
```
But there's a subtle problem with this. Now, c is called before a and b. This would cause issues if c depends on a or b. _shudder_

The only way I can think of to execute it all correctly would be to use another lambda:
```lua
local e = ((function() local __temp; a()[(b())+1] = (function() __temp = c(); return __temp end)(); return __temp end)());
```
At this point things should work correctly but it's very ugly and not great performance-wise.

tomblind:
fixes #225

Non-expression assignments still translate directly:
```ts
x = y;
```

Stand-alone variable assignment expressions:
```ts
z = (x = y);
```
```lua
z = ((function(v) x = v; return v end)(y));
```

Property/element assignment expressions:
```ts
x = (a[0] = "foo");
x = (t.x = "foo");
```
```lua
x = ((function(o, i, v) o[i] = v; return v end)(a, (0)+1, "foo"));
x = ((function(o, i, v) o[i] = v; return v end)(t, "x", "foo"));
```

Destructuring assignment expressions:
```ts
t = ([x, y] = tuple);
t = ([x, y] = tupleReturn());
```
```lua
t = ((function(...) x,y = ...; return {...} end)(table.unpack(tuple)));
t = ((function(...) x,y = ...; return {...} end)(tupleReturn()));
```

There's a lot of possible various combinations so I've tried to cover them with tests.


tomblind:
I removing this PR as I've discovered a few edge cases and there is a lot of duplicate logic with what's needed for #221 so I think it would be better to address both in one PR.

PR #232, opened 2018-10-10
Fixes to return for TupleReturn functions

tomblind:
Currently, TupleReturn functions will only correctly unpack values if returning an array literal. Returning a variable containing a tuple or forwarding a tuple from another TupleReturn call don't work:
```ts
/** !TupleReturn */
function foo(): [number, string]
{
    const r: [number, string] = [0, "foo"];
    return r;
}
/** !TupleReturn */
function bar(): [number, string]
{
    return foo();
}
```
```lua
function foo()
    local r = {0,"foo"};
    return r --Should be unpacked
end
function bar()
    return ({ foo() }) --Should not be wrapped in table
end
```

This PR adds handling for both of these cases.

### DIFF
--- a/src/TSHelper.ts
+++ b/src/TSHelper.ts
@@ -82,6 +82,20 @@ export class TSHelper {
         }
     }
 
+    public static isInTupleReturnFunction(node: ts.Node, checker: ts.TypeChecker): boolean {
+        const declaration = this.findFirstNodeAbove(node, (n): n is ts.Node =>
+        ts.isFunctionDeclaration(n) || ts.isMethodDeclaration(n));
+        if (declaration) {
+            const decorators = this.getCustomDecorators(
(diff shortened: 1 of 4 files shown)


src/TSHelper.ts
```diff
@@ -82,6 +82,20 @@ export class TSHelper {
         }
     }
 
+    public static isInTupleReturnFunction(node: ts.Node, checker: ts.TypeChecker): boolean {
+        const declaration = this.findFirstNodeAbove(node, (n): n is ts.Node =>
+        ts.isFunctionDeclaration(n) || ts.isMethodDeclaration(n));
```
Perryvw: Strange formatting

tomblind: Oops - copy/paste bug

PR #234, opened 2018-10-12
-added support for standard array functions

andreiradu:
array.shift, array.unshift, array.sort and array.reverse

sort uses table.sort which has slightly different behavior than js. It works by default with items other than string(which I think is a good thing). It's also not guaranteed to be stable.

### DIFF
--- a/src/Transpiler.ts
+++ b/src/Transpiler.ts
@@ -26,6 +26,10 @@ export enum LuaLibFeature {
     ArrayIndexOf = "ArrayIndexOf",
     ArrayMap = "ArrayMap",
     ArrayPush = "ArrayPush",
+    ArrayReverse = "ArrayReverse",
+    ArrayShift = "ArrayShift",
+    ArrayUnshift = "ArrayUnshift",
+    ArraySort = "ArraySort",
     ArraySlice = "ArraySlice",
     ArraySome = "ArraySome",
     ArraySplice = "ArraySplice",
@@ -305,7 +309,11 @@ export abstract class LuaTranspiler {
 
     public transpileLuaLibFunction(func: LuaLibFeature, ...params: string[]): string {
         this.importLuaLibFeature(func);
-        return `__TS__${func}(${params.join(", ")})`;
+        if (params.length > 1) {
+            return `__TS__${func}(${params.join(", ")})`;
+        } else {
+          return `__TS__${func}(${params[0]})`;
+        }
     }
 
     public transpileImport(node: ts.ImportDeclaration): string {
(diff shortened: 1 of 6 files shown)


src/Transpiler.ts
```diff
@@ -305,7 +309,11 @@ export abstract class LuaTranspiler {
 
     public transpileLuaLibFunction(func: LuaLibFeature, ...params: string[]): string {
         this.importLuaLibFeature(func);
-        return `__TS__${func}(${params.join(", ")})`;
+        if (params.length > 1) {
+            return `__TS__${func}(${params.join(", ")})`;
+        } else {
+          return `__TS__${func}(${params[0]})`;
```
Perryvw: What is the purpose of this if/else? `params[0]` and `params.join(", ")` look like they would produce exactly the same result?

test/unit/lualib/lualib.spec.ts
```diff
@@ -320,7 +320,73 @@ export class LuaLibArrayTests {
             Expect(result).toBe(expected[1]);
         }
     }
+    @TestCase("[1, 2, 3]", [3, 2, 1])
+    @TestCase("[1, 2, 3, null]", [3, 2, 1])
+    @TestCase("[1, 2, 3, 4]", [4, 3, 2, 1])
```
Perryvw: Please add edge cases empty array and array with 1 element.

test/unit/lualib/lualib.spec.ts
```diff
@@ -320,7 +320,73 @@ export class LuaLibArrayTests {
...
+        }
+    }
+    @TestCase("[1, 2, 3]", [2, 3])
+    @Test("array.shift")
+    public arrayShift(array: string, expected): void {
+        {
+            // Transpile
+            const lua = util.transpileString(
+                `let testArray = ${array};
+                let val = testArray.shift();
```
Perryvw: I would like to also have a test that checks the value of `val`

test/unit/lualib/lualib.spec.ts
```diff
@@ -320,7 +320,73 @@ export class LuaLibArrayTests {
...
+                return JSONStringify(testArray)`);
+
+            // Execute
+            const result = util.executeLua(lua);
+            // Assert
+            Expect(result).toBe(JSON.stringify(expected));
+        }
+    }
+    @TestCase("[1, 2, 3]", [2, 3])
+    @Test("array.shift")
```
Perryvw: Here also singleton array and empty array would be nice test cases.

test/unit/lualib/lualib.spec.ts
```diff
@@ -320,7 +320,73 @@ export class LuaLibArrayTests {
...
+        }
+    }
+    @TestCase("[3, 4, 5]", [1, 2], [1, 2, 3, 4, 5])
+    @Test("array.unshift")
+    public arrayUnshift(array: string, toUnshift, expected): void {
+        {
+            // Transpile
+            const lua = util.transpileString(
+                `let testArray = ${array};
+                testArray.unshift(${toUnshift});
```
Perryvw: Please also add a test on the return value of unshift.

test/unit/lualib/lualib.spec.ts
```diff
@@ -320,7 +320,73 @@ export class LuaLibArrayTests {
...
+                return JSONStringify(testArray)`);
+
+            // Execute
+            const result = util.executeLua(lua);
+
+            // Assert
+            Expect(result).toBe(JSON.stringify(expected));
+        }
+    }
+    @TestCase("[4, 5, 3, 2, 1]", [1, 2, 3, 4, 5])
```
Perryvw: Same edge cases as above would be nice here.

src/Transpiler.ts
```diff
@@ -305,6 +309,9 @@ export abstract class LuaTranspiler {
 
     public transpileLuaLibFunction(func: LuaLibFeature, ...params: string[]): string {
         this.importLuaLibFeature(func);
+        params = params.filter(element => {
```
Perryvw: I still don't understand in what situation element.tostring() would be empty string, could you give an example?

andreiradu: sure, I hit this when calling `array.unshift()` with no params

PR #236, opened 2018-10-13
-added support for class expressions

andreiradu:
This is mostly for the sake of completion. I ran into it while trying to implement class decorators

### DIFF
--- a/src/Transpiler.ts
+++ b/src/Transpiler.ts
@@ -1602,12 +1602,8 @@ export abstract class LuaTranspiler {
 
     // Transpile a class declaration
     public transpileClass(node: ts.ClassLikeDeclarationBase, nameOverride?: string): string {
-        let className: string;
-        if (node.name) {
-            className = this.transpileIdentifier(node.name);
-        } else if (nameOverride) {
-            className = nameOverride;
-        } else {
+        let className = node.name ? this.transpileIdentifier(node.name) : nameOverride;
+        if (!className) {
             throw TSTLErrors.MissingClassName(node);
         }
 
--- a/src/Transpiler.ts
+++ b/src/Transpiler.ts
@@ -805,6 +805,11 @@ export abstract class LuaTranspiler {
                 return this.transpileSpreadElement(node as ts.SpreadElement);
             case ts.SyntaxKind.NonNullExpression:
                 return this.transpileExpression((node as ts.NonNullExpression).expression);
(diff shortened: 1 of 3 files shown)


test/unit/class.spec.ts
```diff
@@ -429,4 +429,27 @@ export class ClassTests {
...
+            }
+            let inst = new b(6);
+            return inst.method();`
+        );
+
+        // Execute
+        const result = util.executeLua(lua);
+
+        // Assert
+        Expect(result).toBe("instance of b");
```
Perryvw: I would also like a test where you call a method defined on class a on an instance of class b.

src/Transpiler.ts
```diff
@@ -1596,13 +1601,16 @@ export abstract class LuaTranspiler {
     }
 
     // Transpile a class declaration
-    public transpileClass(node: ts.ClassDeclaration): string {
-        if (!node.name) {
+    public transpileClass(node: ts.ClassLikeDeclarationBase, nameOverride?: string): string {
+        let className: string;
```
Perryvw: `const className = node.name ? this.transpileIdentifier(node.name) : nameOverride;` is much nicer in my opinion.

src/Transpiler.ts
```diff
@@ -805,6 +805,11 @@ export abstract class LuaTranspiler {
                 return this.transpileSpreadElement(node as ts.SpreadElement);
             case ts.SyntaxKind.NonNullExpression:
                 return this.transpileExpression((node as ts.NonNullExpression).expression);
+            case ts.SyntaxKind.ClassExpression:
+                this.namespace.push("");
```
Perryvw: Any specific reason for this line?

andreiradu: It scopes the class declaration. Otherwise the class would be in global scope. Not sure if this is the best approach - I see it as sort of an anonymous namespace in C++.

Perryvw: Could you provide an input/ouput example?

andreiradu: `let f = function()
{
    let a = class{}
}`
would generate 
`local f; f = function()
    local a = (function()   --[[ should be local ]]  _ = _ or {}
    _.__index = _
    function _.new(construct, ...)
        local self = setmetatable({}, _)
        if construct and _.constructor then _.constructor(self, ...) end
        return self
    end
    function _.constructor(self)
    end
; return _ end)();
end
;`

PR #239, opened 2018-10-13
Public API and better webpack compatibility  

lolleko:


src/TSHelper.ts
```diff
@@ -48,7 +48,7 @@ export class TSHelper {
             // Vanilla ts flags files as external module if they have an import or
             // export statement, we only check for export statements
             return sourceFile.statements.some(statement =>
-                (ts.getCombinedModifierFlags(statement) & ts.ModifierFlags.Export) !== 0
+                (ts.getCombinedModifierFlags(statement as any as ts.Declaration) & ts.ModifierFlags.Export) !== 0
```
Perryvw: Is this because of TS3? Might be nicer to do `sourceFile.statements.filter(ts.isDeclaration).some(ts.getCombinedModifierFlags....` if that avoids this hacky casting.

lolleko: Yes, even though I didn't update to TS3 I left that in there, so it's easier in the future.
Since we are only looking for exports I guess its fine to only check the declarations like you recommended.

Perryvw:
There's an empty src/LuaLibFeatures/ts file in changes I can't comment on.

PR #244, opened 2018-10-15
allowing all strings to be thrown, not just literals

tomblind:
This PR enables other strings, such as template strings, to be thrown. It also fixes indentation on the transpiled code.


src/Transpiler.ts
```diff
@@ -294,7 +294,7 @@ export abstract class LuaTranspiler {
             case ts.SyntaxKind.TryStatement:
                 return this.transpileTry(node as ts.TryStatement);
             case ts.SyntaxKind.ThrowStatement:
-                return this.transpileThrow(node as ts.ThrowStatement);
+                return this.indent + this.transpileThrow(node as ts.ThrowStatement) + "\n";
```
Perryvw: Should there be a `;` after this line?

PR #247, opened 2018-10-17
WIP Upgraded to TS 3.1.3

lolleko:
Fixed issue with es6 lib (Microsoft/TypeScript/issues/27874) for now by using es5 combined with es6 collections

We still should  wait until MS resolves that bug.
Because tstl projects using ES6 will break.


lolleko:
Will reopen with an updated version for TS `3.3`.

PR #251, opened 2018-10-18
refactored try block to allow rethrow

### LINKED ISSUE #245: Cannot re-throw in catch clause
Example:
```ts
function broken()
{
    throw "foobar";
}
try
{
    broken();
}
catch(e)
{
    throw `${e}`;
}
print("everything is fine");
```
```lua
function broken()
    error("foobar")
end
xpcall(function()
    broken();
end,
function(e)
    error(""..tostring(e).."")
end)
print("everything is fine");
```

The problem with this is the error will be silently eaten and the code after will be executed (whereas, in JS, execution would stop at the re-throw).

Something like this might be a better way to go than using xpcall:
```lua
function broken()
    error("foobar")
end
local __TS_try, e = pcall(function()
    broken();
end)
if not __TS_try then
    error(""..tostring(e).."")
end
print("everything is fine");
```

tomblind:
fixes #245 

Also adjusts formatting and ensure everything is wrapped in do..end to keep variables properly scoped.

### DIFF
--- a/test/unit/error.spec.ts
+++ b/test/unit/error.spec.ts
@@ -24,4 +24,45 @@ export class LuaErrorTests {
             );
         }).toThrowError(TranspileError, "Invalid throw expression, only strings can be thrown.");
     }
+
+    @TestCase(1, "ad")
+    @TestCase(2, "bcd")
+    @TestCase(3, "e")
+    @TestCase(4, "g")
+    @TestCase(5, "h")
+    @TestCase(6, "i")
+    @Test("re-throw")
+    public reThrow(count: number, expected: any): void {
+        const source =
+            `declare namespace string { export function match(s: string, p: string): string | null; }
+            let i = ${count};
+            function foo(s: string) { if (--i <= 0) { throw s; } }
+            try {
+                try {
+                    foo("a");
+                    try {
+                        foo("b");
+                    } catch (e) {
+                        foo(e + "c");
+                    }
+                } catch (e) {
+                    foo(e + "d");
+                } finally {
+                    foo("e");
+                }
+                try {
+                    foo("f");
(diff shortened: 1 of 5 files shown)


test/unit/error.spec.ts
```diff
@@ -24,4 +24,45 @@ export class LuaErrorTests {
...
+    @TestCase(3, "e")
+    @TestCase(4, "g")
+    @TestCase(5, "h")
+    @TestCase(6, "i")
+    @Test("re-throw")
+    public reThrow(count: number, expected: any): void {
+        const source =
+            `declare namespace string { export function match(s: string, p: string): string | null; }
+            let i = ${count};
+            function foo(s: string) { if (--i <= 0) { throw s; } }
```
Perryvw: This mechanism actually makes it really hard to see what this test does, I would suggest splitting this up into multiple tests or maybe even just doing a `if (count == 1) { throw "a" }` instead of `foo(`

lolleko:
Okay apparently  we don't use 4 underscores everywhere but we really should. 2 won't work in my opinion because Lua uses double underscore for `__tostring` etc.
3 are escaped by typescript into 4.
At some point I was going to add our own escaping for 4 underscores, so there is absolutely no chance of naming conflicts.

So disregard my previous comment, you don't have to change it here since its already inconsistent, but we should address this in another issue.

test/unit/error.spec.ts
tomblind: Yes, it's pretty convoluted. I'll come up with something better.

tomblind:
Happy to rename the temps to whatever is decided to be best. I don't have a strong opinion, but I agree the double underscore isn't best since it's a lua convention.

PR #253, opened 2018-10-18
Const enum support

DoctorGester:


src/Transpiler.ts
```diff
@@ -1398,6 +1414,33 @@ export abstract class LuaTranspiler {
...
+                    const memberPosition = enumMember.parent.members.indexOf(enumMember);
+
+                    if (memberPosition === -1) {
+                        throw TSTLErrors.UnsupportedProperty(type.symbol.name, property, node);
+                    }
+
+                    const value = enumMembers[memberPosition].value;
+
+                    if (typeof value === "string") {
+                        return "\"" + value + "\"";
```
Perryvw: I personally prefer `` `"${value}"` ``

PR #258, opened 2018-10-20
Fixed JIT using table.unpack

### LINKED ISSUE #238: luajit target should not use table.unpack
LuaJIT still uses 'unpack' from lua5.1; it does not support 'table.unpack', which is what tstl uses for the 'jit' target right now.

lolleko:
Closes #238 for real this time.

sry

### DIFF
--- a/src/Transpiler.ts
+++ b/src/Transpiler.ts
@@ -1623,7 +1623,7 @@ export abstract class LuaTranspiler {
     }
 
     public transpileDestructingAssignmentValue(node: ts.Expression): string {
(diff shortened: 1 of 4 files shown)


adriweb:
The test also has to be updated, btw.

hazzard993:
The LuaJIT transpiler is still using `table.unpack` for its `transpileSpreadElement` function inherited from the 5.2 transpiler

```ts
// main.ts
function x(...args) {
	print(args);
}
let a = [0, 1, 2];
x(...a); // transpiles to table.unpack(a) instead of unpack(a)
```
Compiled with `tstl -p main.ts`
tstl is at version `0.11.0`

PR #259, opened 2018-10-20
Moved transpile switch to Lua 5.2

### LINKED ISSUE #194: Switch statements not allowed in Lua 5.1
Since we use `gotos` to transpile switch statements we need to throw if used in 5.1

lolleko:
Closes #194

### DIFF
--- a/test/unit/conditionals.spec.ts
+++ b/test/unit/conditionals.spec.ts
@@ -330,7 +330,7 @@ export class LuaConditionalsTests {
     }
 
     @Test("switch not allowed in 5.1")
-    public switchThrow50(): void {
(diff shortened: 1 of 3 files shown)


test/unit/conditionals.spec.ts
```diff
@@ -326,4 +328,10 @@ export class LuaConditionalsTests {
 
         Expect(result).toBe(5);
     }
+
+    @Test("switch not allowed in 5.1")
+    public switchThrow50(): void {
```
Perryvw: Is this 5.0 or 5.1? Should be switchThrow51?

PR #264, opened 2018-10-23
New functions

tomblind:
This PR gets the new system working. All functions have an initial context parameter and are called with either a colon or '_G' as the first parameter. Lib functions for bind(), apply() and call() have been added. All tests are updated so they pass.

Notes:
- The new decorator is called 'NoContext'. It prevents functions from transpiling the context parameter or being called with colon/'_G'. Happy to rename if it doesn't make sense.
- transpileFunctionExpression() had a custom code path for transpiling functions that actually had a number of issues (I was even able to crash the compiler with certain function expressions). I replaced it with a version that utilizes the main code path to avoid these issues. The current downside is that it prevents simple arrow functions from being kept on one line (these will be broken down into multiple lines like any other function).

Issues:
- bind(), call() and apply() all require the unpack function. Since different lua versions use different calls for unpack, I had to hack around this for now. It might be worth considering compiling lib functions for each lua target and including the correct version when transpiling.
- JSONStringify is used in a lot of tests, but requires the new decorator in order to transpile correctly. My ugly quick fix was to declare it in each test that uses it but I suspect there's a cleaner way to handle this. I'll look into it.
- ~~I had to disable a tslint rule in the new lib functions because I needed to declare the passed-in function with the new decorator so 'this' can be swapped. But decorators only work on function decls that are interfaces, not type aliases. Allowing decorators on aliases should be investigated.~~ <-Turned out to be an easy fix.
- Need to go through tests and remove some that aren't really relevant now and add some specifically for the new behavior.

### DIFF
--- a/src/Decorator.ts
+++ b/src/Decorator.ts
@@ -21,4 +21,5 @@ export enum DecoratorKind {
     Phantom = "Phantom",
     TupleReturn = "TupleReturn",
     NoClassOr = "NoClassOr",
+    NoContext = "NoContext",
 }
--- a/src/TSHelper.ts
+++ b/src/TSHelper.ts
@@ -109,22 +109,29 @@ export class TSHelper {
         }
     }
 
+    public static collectCustomDecorators(symbol: ts.Symbol, checker: ts.TypeChecker,
+                                          decMap: Map<DecoratorKind, Decorator>): void {
+        const comments = symbol.getDocumentationComment(checker);
+        const decorators =
+            comments.filter(comment => comment.kind === "text")
+                    .map(comment => comment.text.trim().split("\n"))
+                    .reduce((a, b) => a.concat(b), [])
+                    .filter(comment => comment[0] === "!");
+        decorators.forEach(decStr => {
+            const dec = new Decorator(decStr);
+            decMap.set(dec.kind, dec);
+        });
+    }
+
     public static getCustomDecorators(type: ts.Type, checker: ts.TypeChecker): Map<DecoratorKind, Decorator> {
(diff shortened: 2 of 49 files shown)


Perryvw:
I think we should probably disable translation tests in this branch until we are ready to finish this branch. There are a couple clean-up and refactorings I can see would be nice here, but let's also do that once we actually settle on a nice solution.

tomblind:
Personally I don't mind having them on because it gives me something to look and and ensure the output is what I expect to see in a lot of situations, but I won't worry about updating them until it's done.

For JSONStringify, how would you feel about "injecting" the declaration in tranpsileString, in util.ts? It could either be baked into the current version, or with a custom alternate version:
```ts

export function transpileStringWithJSONUtil(str: string, options?: CompilerOptions): string {
    return transpileString("/** !NoContext */ declare function JSONStringify(t: any): string;\n" + str, options);
}
```

Perryvw:
I think injecting it inside transpileString is a fine solution, that function is only for testing anyway.

lolleko:
Dont do it in transpileString, that function is used in the online compiler and is part of the public API.

Perryvw:
Oh it was moved. Should be no big deal to wrap the compiler function in `test/src/util.ts`.

tomblind:
Yes, I saw that - currently I've overridden the one exposed via util so it only affects tests:
```ts
//util.ts
import { transpileString as _transpileString } from "../../src/Compiler";

export function transpileString(str: string, options?: CompilerOptions): string {
    return _transpileString("/** !NoContext */ declare function JSONStringify(t: any): string;\n" + str, options);
}
```

PR #271, opened 2018-10-25
passing nil instead of _G as context for global functions when in ES …

tomblind:
As discussed here: https://github.com/Perryvw/TypescriptToLua/issues/250#issuecomment-432907012

### DIFF
--- a/src/Transpiler.ts
+++ b/src/Transpiler.ts
@@ -92,8 +92,9 @@ export abstract class LuaTranspiler {
         this.importCount = 0;
         this.sourceFile = sourceFile;
         this.isModule = tsHelper.isFileModule(sourceFile);
-        this.isStrict = options.alwaysStrict || options.strict
-            || (options.target && options.target > ts.ScriptTarget.ES5);
+        this.isStrict = options.alwaysStrict
+            || (options.strict && options.alwaysStrict !== false)
+            || (this.isModule && options.target && options.target >= ts.ScriptTarget.ES2015);
         this.loopStack = [];
         this.classStack = [];
         this.exportStack = [];
--- a/src/Transpiler.ts
+++ b/src/Transpiler.ts
@@ -1162,7 +1162,8 @@ export abstract class LuaTranspiler {
             if (!ts.isPropertyAccessExpression(node.expression)
                 && !ts.isElementAccessExpression(node.expression)
                 && !tsHelper.getCustomDecorators(type, this.checker).has(DecoratorKind.NoContext)) {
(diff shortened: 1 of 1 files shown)


src/Transpiler.ts
```diff
@@ -91,6 +92,8 @@ export abstract class LuaTranspiler {
         this.importCount = 0;
         this.sourceFile = sourceFile;
         this.isModule = tsHelper.isFileModule(sourceFile);
+        this.isStrict = options.alwaysStrict || options.strict
+            || (options.target && options.target > ts.ScriptTarget.ES5);
```
Perryvw: Is ES5+ always strict? Or (if it's default true) did you forget to check if it is not explicitly set to false.

src/Transpiler.ts
```diff
@@ -1197,7 +1200,7 @@ export abstract class LuaTranspiler {
         if (!ts.isPropertyAccessExpression(node.expression)
             && !ts.isElementAccessExpression(node.expression)
             && !tsHelper.getCustomDecorators(type, this.checker).has(DecoratorKind.NoContext)) {
-            params = this.transpileArguments(node.arguments, ts.createIdentifier("_G"));
+            params = this.transpileArguments(node.arguments, ts.createIdentifier(this.isStrict ? "nil" : "_G"));
```
Perryvw: Maybe instead of creating an identifier you could just create a null or undefined keyword? I don't know how big that change would be

src/Transpiler.ts
tomblind: Ok, I just looked this up in the ecma specs. In ES5, strict mode has to be explicitly enabled (with "use strict" which is what alwaysStrict produces). In ES6 (aka ES2015) and beyond, modules are always strict (I don't believe it can be disabled), but non-modules still require the explicit enabling. I'll update to match that logic.

src/Transpiler.ts
tomblind: The context paramater of transpileArguments is optional and skips adding the initial parameter if omitted (or undefined). I could re-purpose null to explicitly convert to nil but having separate logic for null and undefined always feels icky to me.

I could change transpileArguments to take either an expression or a string as the context parameter, and just pass the "nil" instead of a temp identifier.

Perryvw: I don't mean actual null, I meant something like `ts.CreateNullKeyword()` or something similar.

Perryvw: actually nvm that won't work here since it is an abstract argument and not a concrete instantiation of one.

tomblind: Ah - sorry - misunderstood. Yes, that seems like a good idea.

PR #273, opened 2018-11-08
Luajit unpack fix

hazzard993:
For #258 

Fixes up transpiled spread operators

`[...[0, 1, 2]]` transpiled using the LuaJIT transpiler becomes `{unpack({0,1,2})};` instead of `{table.unpack({0,1,2})};`

### DIFF
--- a/src/targets/Transpiler.JIT.ts
+++ b/src/targets/Transpiler.JIT.ts
@@ -1,5 +1,4 @@
 import { TSTLErrors } from "../Errors";
-import { LuaTranspiler } from "../Transpiler";
 import { TSHelper as tsHelper } from "../TSHelper";
 import { LuaTranspiler52 } from "./Transpiler.52";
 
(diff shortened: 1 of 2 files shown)


src/targets/Transpiler.JIT.ts
```diff
@@ -39,4 +40,9 @@ export class LuaTranspilerJIT extends LuaTranspiler52 {
     public transpileDestructingAssignmentValue(node: ts.Expression): string {
         return `unpack(${this.transpileExpression(node)})`;
     }
+
+     /** @override */
+    public transpileSpreadElement(node: ts.SpreadElement): string {
+        return LuaTranspiler.prototype.transpileSpreadElement(node);
```
Perryvw: I would prefer you just copy paste the entire function from the transpiler (only one line anyway) here for clarity. I understand the idea behind prototype but I'm not sure that is the best way to do this.

PR #275, opened 2018-11-08
Fixed constructor overloads

### LINKED ISSUE #274: Constructor overloading doesn't work
`class OverloadConstructors {

	constructor(a : string);
	constructor(a : number);
	constructor(a : any){
		/* */
	}
}`

This code causes the transpiler to crash, whereas the same code works just fine when overloading functions.

lolleko:
Fixes #274

### DIFF
--- a/src/Transpiler.ts
+++ b/src/Transpiler.ts
@@ -1777,9 +1777,8 @@ export abstract class LuaTranspiler {
         }
 
         // Find first constructor with body
-        const constructor = node.members.filter(n => {
-            return ts.isConstructorDeclaration(n) && n.body;
-        })[0] as ts.ConstructorDeclaration;
+        const constructor =
+            node.members.filter(n => ts.isConstructorDeclaration(n) && n.body)[0] as ts.ConstructorDeclaration;
         if (constructor) {
             // Add constructor plus initialization of instance fields
             result += this.transpileConstructor(constructor, className);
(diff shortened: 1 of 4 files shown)


src/Transpiler.ts
```diff
@@ -339,7 +339,7 @@ export abstract class LuaTranspiler {
         this.importLuaLibFeature(func);
         params = params.filter(element => {
             return element.toString() !== "";
```
Perryvw: Since there are changes here anyway, could you change this to `element => element.toString() !== ""`, it bothers me more than it should :D

src/Transpiler.ts
```diff
@@ -1778,8 +1778,10 @@ export abstract class LuaTranspiler {
             result += this.indent + `${className}.${fieldName} = ${value}\n`;
         }
 
-        // Try to find constructor
-        const constructor = node.members.filter(ts.isConstructorDeclaration)[0];
+        // Find first constructor with body
+        const constructor = node.members.filter(n => {
+            return ts.isConstructorDeclaration(n) && n.body;
```
Perryvw: Also no reason here to have a `{ return }` lambda if you can just do `m => ts.isConstructorDeclaration(m) && m.body`

lolleko: I did use that but the line was to long with the typecast. Thought if I have to split the lambda it would be more readable to add a proper body.

Perryvw: I prefer breaking expressions at periods instead of in the middle of lambdas, something like
```ts
node.members
    .filter(m => ts.IsConstructorDeclaration(m) && m.body)[0]
    as ts.ConstructorDeclaration
```

PR #279, opened 2018-11-10
Initial Transformer Setup

lolleko:
Added Transformer setup that allows us to incrementally port our transpile functions to transform functions.

We have to prot all transpile functions that transpile Nodes that are not allowed in the 2nd AST.

Once thats done we should restrucutre the whole project, because right now the transform process is just hacked before the transpilation.

### DIFF
--- a/src/Transformer.ts
+++ b/src/Transformer.ts
@@ -287,7 +287,7 @@ export class LuaTransformer {
         return node;
     }
     public visitEmptyStatement(node: ts.EmptyStatement): ts.VisitResult<ts.EmptyStatement> {
-        throw undefined;
+        return undefined;
     }
     public visitBinaryExpression(node: ts.BinaryExpression): ts.VisitResult<ts.BinaryExpression> {
         return node;
--- a/src/Transformer.ts
+++ b/src/Transformer.ts
@@ -1,0 +1,399 @@
+import * as path from "path";
+import * as ts from "typescript";
+
+import {CompilerOptions} from "./CompilerOptions";
+import {DecoratorKind} from "./Decorator";
+import {TSTLErrors} from "./Errors";
+import {TransformHelper as transformHelper} from "./TransformHelper";
+import {TSHelper as tsHelper} from "./TSHelper";
+
+export class LuaTransformer {
+    private checker: ts.TypeChecker;
+    private options: CompilerOptions;
+    private context: ts.TransformationContext;
+
+    constructor(checker: ts.TypeChecker, options: CompilerOptions) {
+        this.checker = checker;
+        this.options = options;
+    }
+
+    public transform(node: ts.SourceFile): ts.SourceFile {
+        return ts
+            .transform(node,
+                       [(ctx: ts.TransformationContext) => {
+                           this.context = ctx;
+                           return this.transformSourceFile.bind(this);
+                       }],
+                       this.options)
+            .transformed[0];
+    }
+
+    public visitEachChild<T extends ts.Node>(node: T): T {
+        return ts.visitEachChild(node, child => this.visitor(child), this.context);
+    }
+
+    public transformSourceFile(node: ts.SourceFile): ts.SourceFile {
+        if (node.isDeclarationFile) {
+            return node;
+        }
+        return this.visitEachChild(node);
+    }
+
+    public visitor(node: ts.Node): ts.VisitResult<ts.Node> {
+        // Ignore declarations
+        if (node.modifiers && node.modifiers.some(modifier => modifier.kind === ts.SyntaxKind.DeclareKeyword)) {
+            return undefined;
+        }
+        switch (node.kind) {
+            // Declarations
+            case ts.SyntaxKind.ImportDeclaration:
+                return this.visitImportDeclaration(node as ts.ImportDeclaration);
+            case ts.SyntaxKind.ClassDeclaration:
+                return this.visitClassDeclaration(node as ts.ClassDeclaration);
+            case ts.SyntaxKind.ModuleDeclaration:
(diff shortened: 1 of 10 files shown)


src/Transformer.ts
```diff
@@ -0,0 +1,400 @@
...
+import {TransformHelper as transformHelper} from "./TransformHelper";
+import {TSHelper as tsHelper} from "./TSHelper";
+
+export class LuaTransformer {
+    private checker: ts.TypeChecker;
+    private options: CompilerOptions;
+    private context: ts.TransformationContext;
+
+    constructor(checker: ts.TypeChecker, options: CompilerOptions) {
+        this.checker = checker;
```
Perryvw:
```ts
constructor(
    private checker: ts.TypeChecker,
    private context: ts.TransformationContext,
    private options: CompilerOptions
) { }
```
Is slightly nicer imo

src/Transformer.ts
```diff
@@ -0,0 +1,400 @@
...
+            throw TSTLErrors.UnsupportedImportType(imports);
+        }
+    }
+    public visitClassDeclaration(node: ts.ClassDeclaration): ts.VisitResult<ts.ClassDeclaration> {
+        return node;
+    }
+    // previously transpileNamespace
+    public visitModuleDeclaration(node: ts.ModuleDeclaration): ts.VisitResult<ts.Node> {
+        const decorators = tsHelper.getCustomDecorators(this.checker.getTypeAtLocation(node), this.checker);
+        // If phantom namespace elide the declaration and returnt he body
```
Perryvw: typo

src/Transformer.ts
```diff
@@ -0,0 +1,400 @@
...
+
+    public visitor(node: ts.Node): ts.VisitResult<ts.Node> {
+        // Ignore declarations
+        if (node.modifiers && node.modifiers.some(modifier => modifier.kind === ts.SyntaxKind.DeclareKeyword)) {
+            return undefined;
+        }
+        switch (node.kind) {
+            // Declarations
+            case ts.SyntaxKind.ImportDeclaration:
+                return this.visitImportDeclaration(node as ts.ImportDeclaration);
```
Perryvw: Random idea: Maybe we should have default `this.ignoreNode(node)` (returns undefined) and `this.visitDefault(node)` (returns node)?

src/Transpiler.ts
```diff
@@ -342,50 +347,8 @@ export abstract class LuaTranspiler {
...
-
-            return result;
-        } else if (ts.isNamespaceImport(imports)) {
-            const resolvedImportPath = this.getImportPath(importPathWithoutQuotes);
-            return `local ${this.transpileIdentifier(imports.name)} = ${requireKeyword}(${resolvedImportPath})\n`;
-        } else {
-            throw TSTLErrors.UnsupportedImportType(imports);
-        }
+        // Handled in transform;
+        return "";
```
Perryvw: Maybe we should throw an error if this function is actually still called?

test/src/util.ts
```diff
@@ -10,6 +10,7 @@ import { createTranspiler } from "../../src/TranspilerFactory";
 import {lauxlib, lua, lualib, to_jsstring, to_luastring } from "fengari";
 
 import * as fs from "fs";
+import { LuaTransformer } from "../../src/Transformer";
```
Perryvw: Unused import

test/unit/modules.spec.ts
```diff
@@ -41,20 +41,6 @@ export class LuaModuleTests {
         Expect(lua).toBe(`require("lualib_bundle")`);
     }
 
-    @Test("Import named bindings exception")
-    public namedBindigsException(): void {
```
Perryvw: Maybe ignore this for now instead of removing, since we probably want to still have this test on the transformer instead of transpiler.

src/Transformer.ts
```diff
@@ -0,0 +1,400 @@
...
+    constructor(checker: ts.TypeChecker, options: CompilerOptions) {
+        this.checker = checker;
+        this.options = options;
+    }
+
+    public transform(node: ts.SourceFile): ts.SourceFile {
+        return ts
+            .transform(node,
+                       [(ctx: ts.TransformationContext) => {
+                           this.context = ctx;
```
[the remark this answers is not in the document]
lolleko: Agreed, we should replicate the pattern we use for transpiler targets (TranspilerFactory, createTranspiler...). Since we most likely want to have differentiations transformers for differentiations Lua versions in the future. But I would reorganise that some point later once we are sure how exactly the transformer architecture will be implemented.

src/Transpiler.ts
lolleko: I think we can remove that function completely. Since all ImportDeclarations get replaced, not sure why I didn't do that already.

src/Transformer.ts
lolleko: I think this is only nicer if all members are declared & initialized in the constructor, which is not the case here.

src/Transformer.ts
lolleko: Ideally all those method stubs should contain an actual implementation in the future. If we figure out that we don't need to transform a node (ignore it) we can just delete the case from the switch statement.

test/unit/modules.spec.ts
lolleko: This test was using a custom created AST/Node I don't think that named bindings exception can ever be thrown.

.codecov.yml
```diff
@@ -4,7 +4,7 @@ coverage:
   status:
     project:
       default:
-        target: 90
+        target: 80
```
Perryvw: MonkaS

PR #283, opened 2018-11-22
New functions

tomblind:
This PR brings everything up to what is spec'd out here: https://github.com/Perryvw/TypescriptToLua/issues/250#issuecomment-437894204
- non-static class/interface methods and function properties (lambdas) are given a context parameter
- static methods/lambdas and stand-alone functions have no context parameter
- errors are thrown for attempts to assign methods to non-methods and vice-versa
- stand-alone functions can be given a context by explicitly adding a 'this' parameter
- methods can be declared without a context by explicitly declaring a 'this: void' parameter

A few notes:
- NoContext decorator has been removed
- bind/call/apply currently throw errors when called with non-methods. Are there any good reasons to support this?
- Custom constructor functions are expected to be non-methods. If that needs to be supported, it might need to be an option in the decorator.

Current edge cases:
```ts
interface Func<T> {
    (this: T, s: string): string;
}
let fn: Func<void> = foo.method; // Should Error
```
Right now, Func<void> is detected as having context when it shouldn't. I need to figure out how to resolve the generic parameter when checking assignments.

```ts
interface A {
    method(s: string): string;
}
interface B {
    method(this: void, s: string): string;
}
declare let a: A;
let b: B = a; // Should Error
```
Assigning to interfaces allows the check to be bypassed and calling the method on 'b' will incorrectly use the dot syntax. This should be solvable by recursively checking assignments to interface types, but this is complex and will likely add quite a lot of overhead. Need to think on this one...

### DIFF
--- a/test/unit/functions.spec.ts
+++ b/test/unit/functions.spec.ts
@@ -233,7 +233,7 @@ export class FunctionTests {
 
     @Test("Function bind")
     public functionBind(): void {
-        const source = `const abc = function (a: string, b: string) { return this.a + a + b; }
+        const source = `const abc = function (this: { a: number }, a: string, b: string) { return this.a + a + b; }
                         return abc.bind({ a: 4 }, "b")("c");`;
 
         const result = util.transpileAndExecute(source);
@@ -243,7 +243,7 @@ export class FunctionTests {
 
     @Test("Function apply")
     public functionApply(): void {
-        const source = `const abc = function (a: string) { return this.a + a; }
+        const source = `const abc = function (this: { a: number }, a: string) { return this.a + a; }
                         return abc.apply({ a: 4 }, ["b"]);`;
 
         const result = util.transpileAndExecute(source);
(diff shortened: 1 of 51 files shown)


Perryvw:
Wow you've been busy, sounds very good! I'll have a look at this later today or tomorrow.

tomblind:
No problem - whenever you have time. It's definitely not ready for integration but worth taking a look at.

One more edge case I forgot to mention:
```ts
interface F {
    (this: Foo, s: string): string;
    (s: string): string;
}
let fn: F = foo.method; // Error
```
Overloads that have both methods and non-methods are currently treated as non-methods when doing the assignment checks. We might want to change things to allow either in this situation.

test/unit/functions.spec.ts
```diff
@@ -233,7 +233,7 @@ export class FunctionTests {
 
     @Test("Function bind")
     public functionBind(): void {
-        const source = `const abc = function (a: string, b: string) { return this.a + a + b; }
+        const source = `const abc = function (this: { a: number }, a: string, b: string) { return this.a + a + b; }
                         return abc.bind({ a: 4 }, "b")("c");`;
```
Perryvw: This test looks strange actually, should this not be `abc.bind({a: 4})("b", "c")`?

Perryvw:
You should probably remove your `/* NoContext */` decorators by the way, now they are no longer used.

PR #285, opened 2018-11-25
Feature/jsdoc annotations

Perryvw:
Added the ability to define decorators like JSDoc annotations. Left the old decorators in for now to stay backwards compatible.

I'm planning on updating documentation once this is merged to only mention the `@` decorators. Before v1 releasees we should remove the old decorators, and probably rename to 'annotations' or something similar.

### DIFF
--- a/src/TSHelper.ts
+++ b/src/TSHelper.ts
@@ -120,6 +120,8 @@ export class TSHelper {
                 const dec = new Decorator(decStr.substr(1));
                 if (dec.kind !== undefined) {
                     decMap.set(dec.kind, dec);
+                    console.warn(`[Deprecated] Decorators with ! are being deprecated, `
+                        + `use @${decStr.substr(1)} instead`);
                 } else {
                     console.warn(`Encountered unknown decorator ${decStr}.`);
                 }
--- a/src/TSHelper.ts
+++ b/src/TSHelper.ts
@@ -109,14 +109,29 @@ export class TSHelper {
             const comments = type.symbol.getDocumentationComment(checker);
             const decorators =
                 comments.filter(comment => comment.kind === "text")
-                        .map(comment => comment.text.trim().split("\n"))
+                        .map(comment => comment.text.split("\n"))
                         .reduce((a, b) => a.concat(b), [])
+                        .map(line => line.trim())
                         .filter(comment => comment[0] === "!");
+
             const decMap = new Map<DecoratorKind, Decorator>();
+
             decorators.forEach(decStr => {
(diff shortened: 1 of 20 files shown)


test/src/util.ts
```diff
@@ -67,6 +67,54 @@ export function transpileAndExecute(tsStr: string): any {
     return executeLua(transpileString(tsStr));
 }
 
+export function parseTypeScript(typescript: string, target: LuaTarget = LuaTarget.Lua53)
```
lolleko: This looks a lot like transpileString from `Compiler.ts`, maybe we can move the dummy compilerHost creation to a shared function.

test/unit/tshelper.spec.ts
```diff
@@ -26,14 +29,132 @@ export class TSHelperTests {
...
+        Expect(decorators.has(DecoratorKind.Phantom)).toBeTruthy();
+        Expect(decorators.has(DecoratorKind.CompileMembersOnly)).toBeTruthy();
+    }
+
+    @Test("GetCustomDecorators multiple default jsdoc")
+    public GetCustomDecoratorsMultipleDefaultJSDoc(): void {
+        const source = `/**
+            * @description abc
+            * @Phantom
+            * @CompileMembersOnly */
```
lolleko: We should consider using lowercase names for decorators to fit into the jsdoc style.
We could just make them case insensitive in Decorator.ts

src/TSHelper.ts
```diff
@@ -109,14 +109,31 @@ export class TSHelper {
...
+                        .map(line => line.trim())
                         .filter(comment => comment[0] === "!");
+
             const decMap = new Map<DecoratorKind, Decorator>();
+
             decorators.forEach(decStr => {
-                const dec = new Decorator(decStr);
-                decMap.set(dec.kind, dec);
+                const dec = new Decorator(decStr.substr(1));
+                if (dec.kind !== undefined) {
```
lolleko: Looks a bit weird to create a decorator jsut to see if its valid, maybe we can move this check to a static function `Decorator::IsValid(string)`

test/unit/tshelper.spec.ts
Perryvw: agreed

test/src/util.ts
Perryvw: Maybe we should look at this later. I have tried to figure out a nice way to do this but was not able to come up with a way to neatly refactor this. I don't want to public expose this function from compiler.js, and I'm not sure it would fit at any other location.

PR #289, opened 2018-11-30
Detecting types derived from array

TheLartians:
Allow the transpiler to detect types derived from array. This allows us to easily create interfaces for array-like types implemented in lua and get a coherent typescript-style interface while retaining the original functionality.

**Oversimplified example** 

An array with a clear method:
```
function ClearableArray()
  local data = {}
  return setmetatable({ 
    clear = function() data = {} end
  },{
    __index = function(self,key) return data[key] end,
    __newindex = function(self,key,value) data[key] = value end,
    __len = function() return #data end
  })
end
```

Typescript interface:
```
declare interface ClearableArray extends Array<any>{
  clear():void
}
declare function ClearableArray():ClearableArray
```

Usage:
 ```
let a = ClearableArray()
a[0] = 1
a.push(2)
print(a.length) // 2
a.clear()
print(a.length) // 0
```

### CODE REVIEW

File: src/Transpiler.ts [hunk 1]
```diff
@@ -1203,10 +1203,21 @@ export abstract class LuaTranspiler {
...
             return this.transpileArrayCallExpression(node);
         }
 
+        // if ownerType inherits from an array, use array calls where appropriate
+        if (tsHelper.isArrayType(ownerType, this.checker)) {
+            const transpiledExpression = this.transpileArrayCallExpression(node, true);
+            if (transpiledExpression !== "") {
+                return transpiledExpression;
+            } else {
+                // non-array method
```
Perryvw: No need for an empty clause, just remove the `else`

src/TSHelper.ts
```diff
@@ -74,7 +74,18 @@ export class TSHelper {
                 && (typeNode as ts.UnionOrIntersectionTypeNode).types.some(this.isArrayTypeNode));
     }
 
-    public static isArrayType(type: ts.Type, checker: ts.TypeChecker): boolean {
+    public static isArrayType(type: ts.Type, checker: ts.TypeChecker, dontCheckBases?: boolean): boolean {
+        if (! dontCheckBases) {
```
Perryvw: Remove space between `!` and `dontCheckBases` to stay consistent with the rest of the formatting.

src/TSHelper.ts
```diff
@@ -74,7 +74,18 @@ export class TSHelper {
                 && (typeNode as ts.UnionOrIntersectionTypeNode).types.some(this.isArrayTypeNode));
     }
 
-    public static isArrayType(type: ts.Type, checker: ts.TypeChecker): boolean {
+    public static isArrayType(type: ts.Type, checker: ts.TypeChecker, dontCheckBases?: boolean): boolean {
```
Perryvw: In what case would we **not** want to check base classes and interfaces?

src/Transpiler.ts
```diff
@@ -1303,7 +1314,7 @@ export abstract class LuaTranspiler {
         }
     }
 
-    public transpileArrayCallExpression(node: ts.CallExpression): string {
+    public transpileArrayCallExpression(node: ts.CallExpression, noException?: boolean ): string {
```
Perryvw: I really don't like this boolean hiding an exception. It is probably better to make a function `tsHelper.isDefaultArrayCallExpression(node: ts.CallExpression)` and just not call `transpileArrayCallExpression` if that returns false.

src/Transpiler.ts
Perryvw: Also like I said below it's probably nicer to remove the need for this if statement using `tsHelper.isDefaultArrayCallExpression()`

src/TSHelper.ts
TheLartians: Maybe to avoid confusion the parameter could be renamed to `nonDerived` or we could create a new method `isPureArrayType(type,checker)` altogether. Not sure about the correct terminology, though.

TheLartians:
I added a translation test to see if the array base detection is working. Doing so I noticed that it won't work if the derived class is a generic passing the parameter to the base array class e.g. `declare interface CustomArray<T> extends Array<T>{ };` as `type.getBaseTypes()` will return `undefined`. Not sure how to find generic base types.

Perryvw:
I prefer functional unit tests over translation tests. You can find bases of generic types using type.typeName/type.typeArguments, try putting `const abc : Array<string> = [];` into https://ts-ast-viewer.com/

TheLartians:
> I prefer functional unit tests over translation tests.

I added a unit test for the derived array type.

> You can find bases of generic types using type.typeName/type.typeArguments

Won't type.typeArguments just give me the generic's argument though?

src/TSHelper.ts
```diff
@@ -236,4 +246,47 @@ export class TSHelper {
         }
         return [false, null, null];
     }
+
+    public static isDefaultArrayCallExpression(node: ts.CallExpression,
```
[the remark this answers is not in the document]
TheLartians: Sure, that would work as well. In this case I replicated the implementation of `transpiler.transpileArrayCallExpression` to match the existing coding style.

Perryvw:
Assignments.ts translation test was removed, please make sure your cases are covered by functional tests and remove the file from your branch too to solve the conflict.

src/TSHelper.ts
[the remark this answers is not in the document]
Perryvw: The reason I don't like this is you are effectively moving part of the transpilation process into TSHelper, which is supposed to only contain helper methods for the transpiler.

TheLartians:
I changed `isDefaultArrayCall(node)` to `isDefaultArrayCallMethodName(string)` to keep the transpilation process separated from the helper functions.
The `Assignments.ts` translation test is removed and this case is still covered by the unit test.

PR #291, opened 2018-11-30
New functions - mixed context overloads

tomblind:
This update addresses issues with function overloads that mix signatures with and without context:
```ts
declare class Foo {
    method(s: string): string; // Has context
    method(this: void, s: string, n: number): string; // No context
}
```

Now, calling the overload will correctly use dot or colon based on the resolved signature:
```ts
declare const foo: Foo;
foo.method("foo");
foo.method("bar", 17);
```
```lua
foo:method("foo"); -- Uses colon
foo.method("bar", 17); -- Uses dot
```

Additionally, assignments to and from function types that have mixed overloads are now banned. Properly validating these assignments would require manually signature matching the overloads and would add a lot of complexity for a fairly fringe-y case. The error advises using only overloads that are all functions or all methods, but not both.
```ts
interface Func { (s: string, n: number): string };
let fn: Func = foo.method; // Error: Unsupported assignment of mixed function/method overload. All overloads should either be functions or methods.
```

### DIFF
--- a/src/Errors.ts
+++ b/src/Errors.ts
@@ -68,4 +68,10 @@ export class TSTLErrors {
 
     public static UnsupportedObjectLiteralElement = (elementKind: ts.SyntaxKind, node: ts.Node) =>
         new TranspileError(`Unsupported object literal element: ${elementKind}.`, node)
+
+    public static UnsupportedFunctionConversion = (node: ts.Node) =>
+        new TranspileError(`Unsupported conversion from method to function.`, node)
+
+    public static UnsupportedMethodConversion = (node: ts.Node) =>
+        new TranspileError(`Unsupported conversion from function to method.`, node)
 }
--- a/src/Errors.ts
+++ b/src/Errors.ts
@@ -88,11 +88,11 @@ export class TSTLErrors {
     public static UnsupportedOverloadAssignment = (node: ts.Node, name?: string) => {
         if (name) {
             return new TranspileError(`Unsupported assignment of mixed function/method overload to "${name}". `
-                                      + `All overloads should either be functions or methods.`,
+                                      + `Overloads should either be all functions or all methods, but not both.`,
                                       node);
         } else {
             return new TranspileError(`Unsupported assignment of mixed function/method overload. `
-                                      + `All overloads should either be functions or methods.`,
+                                      + `Overloads should either be all functions or all methods, but not both.`,
                                       node);
         }
     }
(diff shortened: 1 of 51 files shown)


src/Errors.ts
```diff
@@ -84,4 +84,16 @@ export class TSTLErrors {
...
     }
+
+    public static UnsupportedOverloadAssignment = (node: ts.Node, name?: string) => {
+        if (name) {
+            return new TranspileError(`Unsupported assignment of mixed function/method overload to "${name}". `
+                                      + `All overloads should either be functions or methods.`,
+                                      node);
+        } else {
+            return new TranspileError(`Unsupported assignment of mixed function/method overload. `
+                                      + `All overloads should either be functions or methods.`,
```
Perryvw: This error is a little missleading, how about rephrasing to 'Overloads should either all be methods, or all be functions.'

src/TSHelper.ts
```diff
@@ -249,37 +256,43 @@ export class TSHelper {
...
             && !(ts.getCombinedModifierFlags(sigDecl) & ts.ModifierFlags.Static)) {
             // Non-static method
-            return true;
+            return ContextType.NonVoid;
         }
         if ((ts.isPropertySignature(sigDecl.parent) || ts.isPropertyDeclaration(sigDecl.parent))
             && !(ts.getCombinedModifierFlags(sigDecl.parent) & ts.ModifierFlags.Static)) {
             // Non-static lambda property
-            return true;
+            return ContextType.NonVoid;
```
Perryvw: I thought lambda properties are currently transpiled without context, would that not mean void type here, or do we want to change this?

src/Transpiler.ts
```diff
@@ -1269,8 +1269,9 @@ export abstract class LuaTranspiler {
                 params = this.transpileArguments(node.arguments, sig);
                 return `(rawget(${expr}, ${params} )~=nil)`;
             } else {
-                const type = this.checker.getTypeAtLocation(node.expression);
-                const op = tsHelper.isFunctionWithContext(type, this.checker) ? ":" : ".";
+                const sigDecl = sig.getDeclaration();
+                const op = !sigDecl || tsHelper.getDeclarationContextType(sigDecl, this.checker) !== ContextType.Void
```
Perryvw: Try not to abbreviate variables too much, I don't think there is a lot of abbreviations on other variable names, we should probably try to be somewhat consistent.

src/TSHelper.ts
tomblind: lambda properties get a dummy context (versus stand-alone lambdas, which do not). This ensures they can be called the same as methods when obfuscated behind an interface.

tomblind:
Variable names and the error message are updated. Sorry I was getting lazy there 😄

PR #294, opened 2018-12-01
Transformer v1

lolleko:
This PR adds complete transformer support.
All transform functions are implemented,
but we still need to move a lot of functionality from the Transpiler to the Transformer.

WIP since tests are failing and some polishing is required.


test/translation/builder.spec.ts
```diff
@@ -1,47 +1,48 @@
-import { Expect, Test, TestCases } from "alsatian";
+import { Expect, FocusTest, Test, TestCases } from "alsatian";
```
Perryvw: This should probably not be comitted

test/runner.ts
```diff
@@ -23,7 +23,7 @@ fs.copyFileSync(
 testRunner.outputStream
           // this will use alsatian's default output if you remove this
           // you'll get TAP or you can add your favourite TAP reporter in it's place
-          .pipe(TapBark.create().getPipeable())
+          // .pipe(TapBark.create().getPipeable())
```
Perryvw: Should probably revert this before merging.

test/translation/builder.spec.ts
```diff
@@ -1,47 +1,48 @@
...
-    return b.toString().trim().split("\r\n").join("\n")
+    return b.toString().trim().split("\r\n").join("\n");
 }
 
 export class FileTests {
 
     @TestCases(files)
     @Test("Transformation Tests")
-    public transformationTests(tsFile: string, luaFile:string) {
+    @FocusTest
```
Perryvw: Are you sure this should be comitted?

src/TSHelper.ts
```diff
@@ -56,6 +56,16 @@ export class TSHelper {
         return false;
     }
 
+    public static isIdentifierExported(
+        identifier: ts.Identifier, scope: ts.ModuleDeclaration | ts.SourceFile, checker: ts.TypeChecker): boolean {
```
Perryvw: formatting of this function header looks pretty strange.

src/Transformer.ts
```diff
@@ -63,6 +68,78 @@ export class LuaTransformer {
...
+            case ts.SyntaxKind.ModuleBlock:
+                return this.visitModuleBlock(node as ts.ModuleBlock);
+            // EOF TOKEN
+            case ts.SyntaxKind.EndOfFileToken:
+                return this.visitEndOfFileToken(node as ts.EndOfFileToken);
+            default:
+                throw TSTLErrors.UnsupportedKind("Node", node.kind, node);
+        }
+    }
+    public visitStatement(node: ts.Statement): ts.Statement {
```
Perryvw: We should probably consistently add a white-line between methods

src/Transformer.ts
```diff
@@ -63,6 +68,78 @@ export class LuaTransformer {
...
+            case ts.SyntaxKind.DoStatement:
+            case ts.SyntaxKind.ForStatement:
+            case ts.SyntaxKind.ForOfStatement:
+            case ts.SyntaxKind.ForInStatement:
+            case ts.SyntaxKind.SwitchStatement:
+            case ts.SyntaxKind.BreakStatement:
+            case ts.SyntaxKind.TryStatement:
+            case ts.SyntaxKind.ThrowStatement:
+            case ts.SyntaxKind.ContinueStatement:
+            case ts.SyntaxKind.EmptyStatement:
```
Perryvw: If you want to split like this, is there not some kind of `ts.IsStatement(node)`?

lolleko: There is not, maybe we can reorganise it to look a bit nicer later

src/Transformer.ts
lolleko: Added in 3e791a4

PR #297, opened 2018-12-02
Check inherited accessors

TheLartians:
Iterate the base classes to find inherited accessor methods. The following code will now correctly transpile:

```
class MyBaseClass {
     private _field: number;
     public get field(): number { return this._field; }
     public set field(v: number) { this._field = v; }
}

class MyClass extends MyBaseClass {
}        

let inst = new MyClass();
inst.field = 5;
```

### DIFF
--- a/src/TSHelper.ts
+++ b/src/TSHelper.ts
@@ -99,13 +99,7 @@ export class TSHelper {
     }
 
     public static isArrayType(type: ts.Type, checker: ts.TypeChecker): boolean {
-        const baseTypes = type.getBaseTypes();
-        if (baseTypes) {
-          for (const baseType of baseTypes) {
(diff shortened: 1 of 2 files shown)


src/TSHelper.ts
```diff
@@ -80,6 +80,22 @@ export class TSHelper {
             || (ts.isBinaryExpression(node.parent) && ts.isArrayLiteralExpression(node.parent.left)));
     }
 
+    // iterate over a type and its bases until the callback returns true.
+    public static forAllTypes(type: ts.Type, callback: (type: ts.Type) => boolean): boolean {
```
Perryvw: slightly missleading name, I think 'forTypeOrAnySupertype' would be clearer

TheLartians: agreed.

PR #299, opened 2018-12-06
support in transformer for assignments

tomblind:
Everything works (tests are passing), but this could probably use some cleanup/refactoring.

Note, I had to do some split code paths to deal with statements vs expressions.

I noticed in the TS source, operators like +=, -=, etc... are referred to as 'compound assignment operators" so I'm sticking with that terminology for those.

### DIFF
--- a/src/TSHelper.ts
+++ b/src/TSHelper.ts
@@ -1,5 +1,7 @@
 import * as ts from "typescript";
 import { Decorator, DecoratorKind } from "./Decorator";
+import { TSTLErrors } from "./Errors";
+import { LuaCallExpression, LuaConcatExpression, LuaNode, LuaSyntaxKind } from "./LuaNode";
 
 export class TSHelper {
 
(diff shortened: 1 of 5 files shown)


tomblind:
Looks like I have tests to fix...

src/TSHelper.ts
```diff
@@ -220,12 +222,50 @@ export class TSHelper {
         return [false, null];
     }
 
+    public static isNonCompoundPrefixUnaryOperator(node: ts.PrefixUnaryExpression): boolean {
```
Perryvw: Would it not be nicer to have the opposite of this function and just use a negation where you need this? Depends on what the most common case is I guess.

src/TSHelper.ts
```diff
@@ -220,12 +222,50 @@ export class TSHelper {
...
+        }
+    }
+
     public static isExpressionStatement(node: ts.Expression): boolean {
         return node.parent === undefined
             || ts.isExpressionStatement(node.parent)
             || ts.isForStatement(node.parent);
     }
 
+    public static createLuaConcatExpression(left: ts.Expression, right: ts.Expression): LuaConcatExpression {
```
Perryvw: I think @lolleko made a TransformHelper.ts for functions like these.

tomblind: Totally missed it! I'll move these over.

PR #300, opened 2018-12-07
Lua AST draft

lolleko:


src/LuaAST.ts
```diff
@@ -0,0 +1,292 @@
+// Simplified Lua AST based roughly on http://lua-users.org/wiki/MetaLuaAbstractSyntaxTree,
+// https://www.lua.org/manual/5.3/manual.html (9 – The Complete Syntax of Lua) and the TS AST implementation
+
+// We can ellide a lot of nodes especially tokens and keyowords
+// becasue we dont create the AST from text
+
+namespace tstl {
```
Perryvw: We should probably just spell this out, then `import * as tstl from "typescripttolua"`?

PR #303, opened 2018-12-11
New functions - tests

tomblind:
This PR adds a full suite of tests for function assignments (both valid and invalid).

It also addresses edge cases found, specifically a case where lambda-class-properties marked with 'this: void' would still be considered method-like.

### DIFF
--- a/src/Errors.ts
+++ b/src/Errors.ts
@@ -68,4 +68,10 @@ export class TSTLErrors {
 
     public static UnsupportedObjectLiteralElement = (elementKind: ts.SyntaxKind, node: ts.Node) =>
         new TranspileError(`Unsupported object literal element: ${elementKind}.`, node)
+
+    public static UnsupportedFunctionConversion = (node: ts.Node) =>
+        new TranspileError(`Unsupported conversion from method to function.`, node)
+
+    public static UnsupportedMethodConversion = (node: ts.Node) =>
+        new TranspileError(`Unsupported conversion from function to method.`, node)
 }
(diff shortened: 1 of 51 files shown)


src/Errors.ts
```diff
@@ -71,17 +71,25 @@ export class TSTLErrors {
...
+                                      + `To fix, wrap the method in an arrow function.`,
+                                      node);
         }
     }
 
     public static UnsupportedMethodConversion = (node: ts.Node, name?: string) => {
         if (name) {
-            return new TranspileError(`Unsupported conversion from function to method "${name}".`, node);
+            return new TranspileError(`Unsupported conversion from function to method "${name}". `
+                                      + `To fix, wrap the function in an arrow function.`,
```
Perryvw: Should be a very specific arrow function right?

test/unit/assignments.spec.ts
```diff
@@ -6,6 +6,23 @@ const fs = require("fs");
 
 export class AssignmentTests {
```
Perryvw: Does this mean we can get rid of the assignments translation test file? Or is it already gone on this branch?

src/Errors.ts
tomblind: The idea is that function expressions can have their context type deduced based on what they are being assigned to, so a simple arrow function should suffice. There's a few edge cases where this doesn't work and 'this' has to be explicitly stated or you will still get this error, though.

test/unit/assignments.spec.ts
tomblind: Not in this branch, but it's removed in master, so when things get merged it should be gone.

PR #304, opened 2018-12-13
New functions - return values and context inference

tomblind:
Return values are now run through validateAssignment so that methods can't be returned as functions and vice-versa.
```ts
declare function func(): void;
function returnsMethod(): (this: Foo) => void {
    return func; // Error
}
```

Function expressions passed as arguments can have their context type inferred now.
```ts
declare function takesMethod(method: (this: Foo, s: string) => string);
takesMethod(s => s); // Works - lambda will transpile with a dummy parameter
```

Same for return values:
```ts
function returnsMethod(): (this: Foo, s: string) => string {
    return s => s; // Works
}
```

Note that more complex cases will still fail to to properly infer the context type:
```ts
declare class Foo {}
interface Method {
    (this: Foo, s: string): string;
}
const a: Method[] = [s => s]; // Error - cannot convert function to method
const b: Method[] = [function(this: Foo, s) { return s; }]; // Workaround
```

### DIFF
--- a/src/TSHelper.ts
+++ b/src/TSHelper.ts
@@ -79,11 +79,6 @@ export class TSHelper {
         return typeNode && this.isArrayTypeNode(typeNode);
     }
 
-    public static isCallableType(type: ts.Type, checker: ts.TypeChecker): boolean {
-        const sigs = checker.getSignaturesOfType(type, ts.SignatureKind.Call);
-        return sigs.length > 0;
-    }
-
     public static isFunctionType(type: ts.Type, checker: ts.TypeChecker): boolean {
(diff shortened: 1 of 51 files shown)


src/TSHelper.ts
```diff
@@ -116,6 +116,15 @@ export class TSHelper {
         }
     }
 
+    public static getFunctionReturnType(node: ts.Node, checker: ts.TypeChecker): ts.Type {
```
Perryvw: Maybe something like `getContainingFunctionReturnType` would be clearer

tomblind: Works for me.

Issue #281, opened 2018-11-11
'this' parameter not stripped from function declarations


tomblind:
TS allows you to explicitly type 'this' in function declarations, but it's just a typing hint that should be stripped on compilation. Currently, it is still added to the output declaration:
```ts
class Foo {
    public method(this: Foo, s: string): string { return "foo" + s; }
}
```
=>
```lua
Foo = Foo or {}
Foo.__index = Foo
function Foo.new(construct, ...)
    local self = setmetatable({}, Foo)
    if construct and Foo.constructor then Foo.constructor(self, ...) end
    return self
end
function Foo.constructor(self)
end
function Foo.method(self,this,s) --this is going to be a problem
    return "foo" .. s
end
```

Perryvw:
Should we take this to mean we should bind self explicitly by defining the method inside `new` so it bakes in the value of `self`? (replace the self parameter with a dummy)

Perryvw:
Actually the most straightforward way to implement is to just ignore `this` for methods, and interpret it as adding a `self` argument to any function using it. This would require some extra checks when transpiling function calls.

Issue #192, opened 2018-09-05
Class functions overwritten at runtime are not called correctly


lolleko:
```
declare class Panel {
  Paint(width: number, height: number): void;
}

let pnl = new Panel();

pnl.Paint = (width, height) => {
  // custom paint function
}
```

=>

```
local pnl = Panel.new();

pnl.Paint = function(width, height)
   // causes error when using width because width is userdata (self)
end
```

Possible fix automatically add self as first parameter, if functions are overwritten this way. But this might cause issues when using this inside the function.

Perryvw:
Relevant to #250

PR #307, opened 2018-12-15
New functions

### LINKED ISSUE #250: 'this' binding and dot vs colon
I've found a couple of issues with transpiling functions and dealing with 'this'.

Example 1:
```ts
interface Foo {
    bar(s: string): void;
}
const o: Foo = {
    bar: (s: string) => print(s)
}
o.bar("foo");
```
```lua
local o = {bar = function(s) return print(s) end};
o:bar("foo"); --prints "table: 00709330"
```
Here, the colon is mistakenly used because the interface declares the function as a method, but the implementation creates it as an arrow function. This is a tough issue to solve since there's no way to know for sure which operator to use at the call site.

Example 2:
```ts
class C {
    private prop = "foo";
    public outer() {
        const o = {
            prop: "bar",
            innerFunc: function() { print(this.prop); },
            innerArrow: () => print(this.prop)
        };
        o.innerFunc(); //Should print "bar"
        o.innerArrow(); //Should print "foo"
    }
}
let c = new C();
c.outer();
```
```lua
C = C or {}
C.__index = C
function C.new(construct, ...)
    local self = setmetatable({}, C)
    self.prop = "foo"
    if construct and C.constructor then C.constructor(self, ...) end
    return self
end
function C.constructor(self)
end
function C.outer(self)
    local o = {prop = "bar",innerFunc = function()
        print(self.prop);
    end
,innerArrow = function() return print(self.prop) end};
    o.innerFunc(); --Incorrectly prints "foo"!
    o.innerArrow();
end
local c = C.new(true);
c:outer();
```
Here the inner function incorrectly captures self from the outer method.

I suspect there's probably a lot of other similar edge cases around. They all boil down to inconsistencies with how JS handles 'this' binding.

As I understand it, in JS, all functions have a 'this', which is whatever was on the left side of the dot when called. The 2 exceptions are arrow functions and stand-alone functions (ones with no left-side to bind). Arrow functions capture the 'this' from their outer scope and stand-alone functions have the global scope (window) as their 'this'.

I believe the solution is to re-work how functions are transpiled overall to mirror that behavior:
- All functions should be given an initial 'self' parameter
- Arrow functions should name this first parameter with a dummy name ('_') so they capture the self of their outer scope
- All functions (except stand-alone) should be called with colon syntax
- Stand alone functions should (maybe?) pass _G as their first parameter

With this setup, all lua functions would have a self that mirrors 'this' from the equivalent JS. Plus, there would be no need to decide whether to use dot or colon as all calls would use colon.

The one case this doesn't handle is functions from external sources that need to be called with dot syntax. A decorator would probably be the way to handle this. I know that was shot down before (see #108), but I think it's worth reconsidering.

Thoughts?

Perryvw:
Rework of how functions are transpiled, adds bind/apply/call methods.

Closes #250, #192, #281


test/translation/lua/dotColonFunctionCalls.lua
```diff
@@ -1,6 +1,6 @@
 classInstance:colonMethod();
-classInstance.dotMethod();
+classInstance:dotMethod();
```
tomblind: We should probably get rid of this translation test. It's fully covered by functional tests and doesn't even make sense with how functions are transpiled now.

TheLartians:
When casting types, the transpiler wrongly accepts methods as functions.  
```
interface Callbacks { callback:(this:void,i:number)=>number }
let callbacks = <Callbacks>{ callback:(i)=>{ return 2*i; } }
callbacks.callback(2) // arithmetic on nil
```
I believe the second line should throw an unsupported conversion error. 
Transpiled lua code for reference:
```
local callbacks = {callback = function(____,i)
    return 2*i
end};
callbacks.callback(2);
```

PR #308, opened 2018-12-17
allow non-default properties with derived array types

TheLartians:
Generalises #289 for properties. This will allow working with interfaces that extend `Array<T>` and have member properties. Example:

```
interface NamedArray<T> extends Array<T>{
    name:string,
}
```


test/unit/array.spec.ts
```diff
@@ -39,16 +39,21 @@ export class ArrayTests {
         Expect(result).toBe(5);
     }
 
+    @TestCase("firstElement()", 3)
+    @TestCase("name", "array")
```
Perryvw: Add a testcase for `length` here (since that's what you added)

TheLartians: good point

PR #310, opened 2018-12-21
simplify ternary transpilation

TheLartians:
Transpile the ternary operator using and/or and boxing/unboxing. This reduces the cost of the ternary operator from a library call and the creation of two closures to a table allocation and lookup.

**Example**
```
declare function f(): number | undefined;
const a = true ? 0 : f();
const b = true ? f() : 0;
```

**Before**
```
-- Lua Library Imports
function __TS__Ternary(condition,cb1,cb2)
    if condition then
        return cb1()
    else
        return cb2()
    end
end

local a = __TS__Ternary(true, function() return 0 end, function() return f() end);
local b = __TS__Ternary(true, function() return f() end, function() return 0 end);
```

**After** 
LuaJIT:
```
local a = ((true) and (0) or (f()));
local b = ((true) and {f()} or {0})[1];
```
Other:
```
local a = ((true) and (0) or (f()));
local b = ((true) and function() return f(); end or function() return 0; end)();
```


tomblind:
It would probably be good to remove the lib function altogether.

TheLartians:
ok, I think I got it all now.

lolleko:
I did some benchmarks and I'm getting mixed results:

In **Lua 5.3** IIFEs are always faster than box/unboxing around 400%.

In **LuaJIT** IIFEs are faster around 20-50% for small arrays.  
12% faster for an array size around 5000000. 
After that boxing/unboxing is significantly faster (500%).  


Tested Code:
```
for i=1, arrSize do
  y = ((arr[i]) and {true} or {false})[1]
end
```

```
for i=1, arrSize do
  z = (function (condtion, val1, val2)
    if condtion then
      return val1
    else
      return val2
    end
  end)(arr[i], true, false)
end
```

I would argue that 99.9% of the time arrays have less than 50 Million entries, in which case IIFEs are defintly faster.

So we should still remove the lib call but replace it with IIFEs instead of boxing/unboxing.

TheLartians:
Wow thats really interesting. I had no idea IIFEs are that well optimised in lua. We should definitely use IIFEs then.

lolleko:
Did some more testing seems like:

```
local libTernary  = function (condtion, val1, val2)
  if condtion then
    return val1
  else
    return val2
  end
end

for i=1, arrSize do
  z = libTernary(arr[i], true, false)
end
```

performs exactly the same like:

```
for i=1, arrSize do
  z = (function (condtion, val1, val2)
    if condtion then
      return val1
    else
      return val2
    end
  end)(arr[i], true, false)
end
```

in both JIT and 5.3, so we might aswell keep the lib function since its a bit more readable and adjust to take the 2 raw expressions as parameter without wrapping them in extra functions.

TheLartians:
However, the wrapping is necessary to perform lazy evaluation which otherwise might introduce unwanted side effects. Can you also benchmark the performance with function wrapping?

TheLartians:
Completely agree. Here I just assumed that function calls must be slower than table allocation and lookup. I will try some own benchmarking later.

TheLartians:
Thanks for reminding me to always benchmark before optimising. So as boxing/unboxing is expensive we should definitely replace it with an IIFE. In the implementation below only one call is necessary as opposed to two when using the library function. So it seems performance using IIFE is better on all implementations. 

Code: https://gist.github.com/TheLartians/0b8bbdc551b67b7205e347df7525c205
Benchmark:
```
lua5.3 benchmark.lua 
boxUnboxTime time:	1.517611
IIFE time:	0.322223
lib time:	0.420415
```
```
lua5.1 benchmark.lua 
boxUnboxTime time:	1.444085
IIFE time:	1.084926
lib time:	1.961076
```
```
luajit benchmark.lua 
boxUnboxTime time:	0.041535
IIFE time:	0.254946
lib time:	0.371651
```

DoctorGester:
IMO we should have no boxing/unboxing on things guaranteed not to produce side effects

TheLartians:
@DoctorGester the boxing/closures are necessary even without side effects when dealing with boolean or possibly undefined values, when using the `and`/`or` syntax. The reason being that statements like `c and false or true` always returns `true`, while the boxed statement `(c and {false} or {true})[1]` works as expected. Alternatively, when there are no side effects we could use a lib call such as `libTernary(c, true, false)`. However, in my benchmarks this is still slower than the more general IIFE approach or boxing in luajit. Feel free to try it yourself: https://gist.github.com/TheLartians/0b8bbdc551b67b7205e347df7525c205 .
In the long run, we can probably introduce temporary variables and avoid the ternary statement altogether, however this requires AST transformations.

DoctorGester:
I understand the problem, however most cases could be handled with the usual lua ternary, I don't think settling for a general solution which is in many cases worse than a specific one is a good idea. Since it's typescript we should assume the type is known and understand what's a boolean/nullable type and what isn't. And yeah a temporary variable would be preferable in other cases

DoctorGester:
You can always go for the most defensive version when type information is unreliable.

TheLartians:
I've added a `isNonFalsible` helper function that currently returns true when the type is a non-boolean literal. In that case the ternary is transpiled without protecting boxes or closures.
Are there more cases we can easily check that are guaranteed to be non-falsible?

src/TSHelper.ts
```diff
@@ -316,4 +316,18 @@ export class TSHelper {
         return defaultArrayPropertyNames.has(methodName);
     }
 
+    public static isNonFalsible(type: ts.Type): boolean {
+        const falsibleFlags = ts.TypeFlags.Boolean
+            | ts.TypeFlags.BooleanLiteral
+            | ts.TypeFlags.Undefined
+            | ts.TypeFlags.Any;
```
tomblind: Will this catch null-able types?
```ts
declare function foo(): string | null;
const x = condition ? foo() : "bar";
```

src/Transpiler.ts
```diff
@@ -1019,13 +1018,21 @@ export abstract class LuaTranspiler {
...
         const condition = this.transpileExpression(node.condition);
         const val1 = this.transpileExpression(node.whenTrue);
         const val2 = this.transpileExpression(node.whenFalse);
+        return `((${condition}) and function() return ${val1}; end or function() return ${val2}; end)()`;
+    }
 
-        return this.transpileLuaLibFunction(LuaLibFeature.Ternary, condition,
-                                            `function() return ${val1} end`, `function() return ${val2} end`);
+    public transpileConditionalExpression(node: ts.ConditionalExpression): string {
+        if (tsHelper.isNonFalsible(this.checker.getTypeAtLocation(node.whenTrue))) {
```
tomblind: What about if strictNullChecks is off? Shouldn't the protected version always be used in this case?

TheLartians: Currently we are playing it safe and always protecting non-literal expressions. With strictNullChecks enabled we could also allow non-literal types to be transpiled without protection.
It should be trivial to extend `isNonFalsible` to support strictNullChecks in the future.

src/TSHelper.ts
TheLartians: Update: now the above only applies when strictNullChecks are not enabled. Otherwise I recursively iterate through all union types and check if they are all non-falsible.

src/Transpiler.ts
TheLartians: Update: done.

src/TSHelper.ts
tomblind: Re-reading this I realize why I got confused. Will a null literal get caught by those flags?

TheLartians: Good point, I wasn't aware that there is a difference between `null` and `undefined`. I added an additional check for it.

src/TSHelper.ts
```diff
@@ -453,4 +453,28 @@ export class TSHelper {
     public static isDefaultArrayPropertyName(methodName: string): boolean {
         return defaultArrayPropertyNames.has(methodName);
     }
+
+    public static isNonFalsible(type: ts.Type, strictNullChecks: boolean): boolean {
```
tomblind: IsNonFalsible is a pretty confusing name - perhaps it should be reversed to be IsFalsible and checked with a not "!" operator later.

TheLartians: Fine by me.

src/Transpiler.ts
```diff
@@ -1030,13 +1030,21 @@ export abstract class LuaTranspiler {
...
         const condition = this.transpileExpression(node.condition);
         const val1 = this.transpileExpression(node.whenTrue);
         const val2 = this.transpileExpression(node.whenFalse);
+        return `((${condition}) and function() return ${val1}; end or function() return ${val2}; end)()`;
+    }
 
-        return this.transpileLuaLibFunction(LuaLibFeature.Ternary, condition,
-                                            `function() return ${val1} end`, `function() return ${val2} end`);
+    public transpileConditionalExpression(node: ts.ConditionalExpression): string {
+        if (tsHelper.isFalsible(this.checker.getTypeAtLocation(node.whenTrue), this.options.strictNullChecks)) {
```
tomblind: One more thing (sorry): "strictNullChecks" is implicitly enabled if "strict" is set in options, so you'll want to check for that too.

TheLartians: Alright.

PR #313, opened 2018-12-24
correctly passing context to functions accessed as elements

### LINKED ISSUE #266: Losing context when calling method with a computed key
[Playground](https://typescripttolua.github.io/play.html#src=declare%20const%20foo%3A%20%7B%20bar(a%3F%3A%20number)%3A%20void%20%7D%3B%0Afoo.bar(1)%3B%0Afoo%5B'bar'%5D()%3B%0Afoo%5B'bar'%5D(1)%3B%0A)

Suggested solution:

```lua
-- Simple cases
foo["bar"](foo);
foo["bar"](foo, 1);
```

When object isn't a pure value:

```typescript
c.chain()['foo']('bar');
```

```lua
__TS__ComputedCall(c:chain(), "foo", "bar");
```

Chained api:

```typescript
Promise.resolve(1)
  ['then'](() => Promise.resolve(2))
  ['then'](() => Promise.reject(3))
  ['catch'](() => Promise.resolve(4))
  ['then'](console.log);
```

```lua
__TS__ComputedCall(
  __TS__ComputedCall(
    __TS__ComputedCall(
      Promise:resolve(1), "then", function() return Promise:resolve(2) end
    ), "then", function() return Promise:reject(3) end
  ), "catch", function() return Promise:resolve(4) end
), "then", console.log);
```

This implementation may be found there: https://github.com/ark120202/babel-lua/tree/master/packages/babel-plugin-lua-function-context

tomblind:
fixes #266 

Also has a bit of refactoring in transpileCallExpression()

### DIFF
--- a/src/Transpiler.ts
+++ b/src/Transpiler.ts
@@ -1171,11 +1171,16 @@ export abstract class LuaTranspiler {
             && tsHelper.isInTupleReturnFunction(node, this.checker);
         const isInDestructingAssignment = tsHelper.isInDestructingAssignment(node);
         const returnValueIsUsed = node.parent && !ts.isExpressionStatement(node.parent);
+        const wrapResult = isTupleReturn && !isTupleReturnForward && !isInDestructingAssignment && returnValueIsUsed;
 
         if (ts.isPropertyAccessExpression(node.expression)) {
             const result = this.transpilePropertyCall(node);
-            return isTupleReturn && !isTupleReturnForward && !isInDestructingAssignment && returnValueIsUsed
-                ? `({ ${result} })` : result;
+            return wrapResult ? `({ ${result} })` : result;
+        }
+
+        if (ts.isElementAccessExpression(node.expression)) {
+            const result = this.transpileElementCall(node);
+            return wrapResult ? `({ ${result} })` : result;
         }
 
         const signature = this.checker.getResolvedSignature(node);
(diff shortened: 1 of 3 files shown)


src/Transpiler.ts
```diff
@@ -1279,6 +1277,39 @@ export abstract class LuaTranspiler {
         }
     }
 
+    public transpileElementCall(node: ts.CallExpression): string {
+        if (!ts.isElementAccessExpression(node.expression)) {
+            throw TSTLErrors.InvalidPropertyCall(node);
```
Perryvw: Shouldn't this be InvalidElementCall?

src/Transpiler.ts
```diff
@@ -1279,6 +1277,39 @@ export abstract class LuaTranspiler {
...
+        }
+
+        const signature = this.checker.getResolvedSignature(node);
+        let parameters = this.transpileArguments(node.arguments, signature);
+
+        const signatureDeclaration = signature.getDeclaration();
+        if (!signatureDeclaration
+            || tsHelper.getDeclarationContextType(signatureDeclaration, this.checker) !== ContextType.Void) {
+            // Pass left-side as context
+            if (node.arguments.length > 0) {
```
Perryvw: This looks a little weird but I do understand it prevents some duplication. Maybe if you just improve the comment a little bit on the logic for this `if` statement, that would already help.

PR #315, opened 2018-12-28
Fixed transpiling break in switches with block cases and nested breaks

DoctorGester:


src/Transpiler.ts
```diff
@@ -128,6 +137,18 @@ export abstract class LuaTranspiler {
         this.exportStack[this.exportStack.length - 1].push({name: nameIn, node: nodeIn, dummy: dummyIn});
     }
 
+    public peekSpecialScope(): SpecialScope {
+        return this.scopeStack[this.scopeStack.length - 1];
+    }
+
+    public pushSpecialScope(scopeType: ScopeType): void {
+        this.scopeStack.push({ type: scopeType, id: this.genVarCounter });
```
Perryvw: You should probably increment genVarCounter after you use it, to prevent duplicates here

src/Transpiler.ts
```diff
@@ -128,6 +137,18 @@ export abstract class LuaTranspiler {
...
+    }
+
+    public popSpecialScope(): number {
+        return this.scopeStack.pop().id;
```
Perryvw: Probably better to just return the pop directly, instead of just the id

PR #317, opened 2019-01-03
Ts to lua ast assignment expressions

tomblind:
Assignments and compound assignments all seem to be working but many of the tests still fail due to unimplemented features (template expressions, etc..).


src/LuaTransformer.ts
```diff
@@ -911,8 +913,53 @@ export class LuaTransformer {
         );
     }
 
-    public transformExpressionStatement(statement: ts.ExpressionStatement): StatementVisitResult {
-        return tstl.createExpressionStatement(this.transformExpression(statement.expression));
+    public transformExpressionStatement(expression: ts.Expression): StatementVisitResult {
```
lolleko: Maybe overload this so we can pass both `ts.Expression` and `ts.ExpressionStatement`

src/LuaTransformer.ts
```diff
@@ -1267,11 +1520,55 @@ export class LuaTransformer {
     }
 
     public transformPostfixUnaryExpression(expression: ts.PostfixUnaryExpression): tstl.Expression {
-        throw new Error("Method not implemented.");
+        return this.transformCompoundAssignmentExpression(
+            expression.operand,
+            ts.createLiteral(1),
+            expression.operator === ts.SyntaxKind.PlusPlusToken
```
lolleko: Ternaries inside arguments are to very readable imo. Consider changing it to:

```
        let operator = expression.operator === ts.SyntaxKind.PlusPlusToken ? tstl.SyntaxKind.AdditionOperator : tstl.SyntaxKind.SubractionOperator;
        return this.transformCompoundAssignmentExpression(
            expression.operand,
            ts.createLiteral(1),
            operator,
            true
````

PR #325, opened 2019-01-08
Ts to lua ast fixes

tomblind:
- Fixed issue where colon was not used for method calls
- Fixed bind, call and apply
- Fixed issues with getters/setters and their tests
- Correctly using unpack vs. table.unpack based on lua target

### DIFF
--- a/src/targets/LuaTransformer.JIT.ts
+++ b/src/targets/LuaTransformer.JIT.ts
@@ -1,10 +1,16 @@
 import * as ts from "typescript";
 
+import * as tstl from "../LuaAST";
 import {LuaTransformer} from "../LuaTransformer";
 import {TSHelper as tsHelper} from "../TSHelper";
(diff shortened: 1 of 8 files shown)


src/targets/LuaTransformer.JIT.ts
```diff
@@ -1,10 +1,16 @@
 import * as ts from "typescript";
 
+import * as tstl from "../LuaAST";
 import {LuaTransformer} from "../LuaTransformer";
 import {TSHelper as tsHelper} from "../TSHelper";
 import { LuaTransformer52 } from "./LuaTransformer.52";
 
 export class LuaTransformerJIT extends LuaTransformer52 {
+	/** @override */
```
Perryvw: woops indenting

tomblind: nice catch 😄

PR #330, opened 2019-01-09
Implemented DeleteExpression and NonNullExpression, fixed some broken tests

Perryvw:


src/LuaTransformer.ts
```diff
@@ -1677,6 +1683,18 @@ export class LuaTransformer {
...
+    public transformDeleteExpression(expression: ts.DeleteExpression): tstl.CallExpression {
+        const lhs = this.transformExpression(expression.expression) as tstl.IdentifierOrTableIndexExpression;
+        const assignment = tstl.createAssignmentStatement(
+            lhs,
+            tstl.createNilLiteral(),
+            undefined,
+            expression
+        );
+
+        return this.createImmediatelyInvokedFunctionExpression([assignment], []);
```
tomblind: Technically, this should return a 'true' value: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete#Return_value

Also, it might be good to do a statement version (called from transformExpressionStatement) so the IIFE isn't created when not needed.

PR #334, opened 2019-01-12
Implemented try/catch and throw

Perryvw:


src/LuaTransformer.ts
```diff
@@ -1350,12 +1350,50 @@ export class LuaTransformer {
...
+    public transformTryStatement(statement: ts.TryStatement): StatementVisitResult {
+        const pCall = tstl.createIdentifier("pcall");
+        const tryBlock = this.transformBlock(statement.tryBlock);
+        const tryResult = tstl.createIdentifier("__TS_try");
+
+        const returnVariables = statement.catchClause && statement.catchClause.variableDeclaration
+            ? [tryResult, this.transformIdentifier(statement.catchClause.variableDeclaration.name as ts.Identifier)]
+            : [tryResult];
+
+        const catchAssignment = this.createLocalOrGlobalDeclaration(
```
tomblind: This should use tstl.createVariableDeclaration as this should never be global.

PR #343, opened 2019-01-18
Fix renamed class extension

TheLartians:
This fixes errors when extending from a renamed class.
The following example will now transpile and run correctly.

```
// base.ts
export class Type{ 
  constructor(){}
};
```

```
// main.ts
import { Type as TypeBase } from "./base";
class Type extends TypeBase{
  constructor(){ super(); }
};
const x = new Type();
```

### DIFF
--- a/src/TSHelper.ts
+++ b/src/TSHelper.ts
@@ -54,19 +54,25 @@ export class TSHelper {
         return result;
     }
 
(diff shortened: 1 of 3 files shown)


Perryvw:
Could you add a functional test for this situation?

PR #344, opened 2019-01-18
Add support for union accessors

TheLartians:
Handles the case where an accessor is called on a union type. When mixing types with and without accessors an error is thrown.

### DIFF
--- a/src/Errors.ts
+++ b/src/Errors.ts
@@ -117,6 +117,8 @@ export class TSTLErrors {
                                   node);
     }
 
-    public static UnsupportedUnionAccessor = (node: ts.Node) =>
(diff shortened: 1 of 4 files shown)


src/Errors.ts
```diff
@@ -116,4 +116,7 @@ export class TSTLErrors {
                                   + "the TupleReturn decorator.",
                                   node);
     }
+
+    public static UnsupportedUnionAccessor = (node: ts.Node) =>
+    new TranspileError(`Unsupported union of accessor with non-accessor types.`, node)
```
Perryvw: Please indent this line like the other ones

PR #347, opened 2019-01-19
Switch transformer

Perryvw:


src/LuaTransformer.ts
```diff
@@ -1351,12 +1351,67 @@ export class LuaTransformer {
         );
     }
 
-    public transformSwitchStatement(arg0: ts.SwitchStatement): StatementVisitResult {
-        throw new Error("Method not implemented.");
+    public transformSwitchStatement(statement: ts.SwitchStatement): StatementVisitResult {
+        if (this.options.luaTarget <= LuaTarget.Lua51) {
```
tomblind: We should not use math comparison for string values 😃

PR #349, opened 2019-01-20
Transformer exports

### LINKED ISSUE #286: Rework exports
Currently variables/functions are declared locally and exported at the end of their block.
TS:
```ts
export let x = 12;

x = x + 13;
```
```lua
local exports = exports or {}
local x = 12;
x = (x+13);
exports.x = x
return exports
```

We should change the translation to:
```lua
local exports = exports or {}
exports.x = 12;
exports.x = exports.x + 13;
```

The problem is that their is no easy way to fidn out if an avriable/identifier was exported after declaration.

Possible Solutions:
* Check if export flag is set in the variablestatement `parentOf(checker.getType(identifier).getSymbol().Declarations[0]` and check the parent node for export modifier

* Check the sourcefile or namespace symbol's export table and check if the identifiers symbol is in that table `this.checker.getTypeAtLocation(this.sourceFile).getSymbol().exports.includes(this.checker.getTypeAtLocation(identifier))` or `this.checker.getTypeAtLocation(this.currenNamespace).getSymbol().exports.includes(this.checker.getTypeAtLocation(identifier))`

This should be probably be handled in the transformer (#278);

lolleko:
Closes #286


src/LuaTransformer.ts
```diff
@@ -3033,7 +3113,16 @@ export class LuaTransformer {
             || this.currentNamespace
             || (tsOriginal && tsHelper.findFirstNodeAbove(tsOriginal, ts.isFunctionLike))
         ) {
-            statements.push(tstl.createVariableDeclarationStatement(lhs, undefined, parent));
+            if (!isArray(lhs)) {
```
Perryvw: Is this `isArray` correct? I thought it was `Array.isArray`, or where is this defined?

src/LuaTransformer.ts
```diff
@@ -3033,7 +3113,16 @@ export class LuaTransformer {
...
+                lhs = [lhs];
+            }
+            const shouldExport = lhs.some(i => this.isIdentifierExported(i.text));
+            if (shouldExport) {
+                statements.push(
+                    tstl.createAssignmentStatement(lhs.map(i => this.createExportedIdentifier(i)), undefined, parent));
```
Perryvw: This looks strange, if there is one thing that should be exported, it calls createdExportedIdentifier for all of them? Does that mean createExportedIdentifier handles non-exported stuff too? Why not just always call this then? What is the point of shouldExport?

src/LuaTransformer.ts
lolleko: its defined in node's util lib. But since we use Array.isArray in other places I will change it to Array.isArray swell.

src/LuaTransformer.ts
lolleko: `export let x, y, z;` will result in all 3 identifiers being exported.

PR #354, opened 2019-01-23
Async await

tjahn:
Hello,

I really like this library and decided to extend it with async-await support using lua coroutines.
Using coroutines I wrote the functions __TS__Async and __TS_Await to behave exactly like the async and await keywords in Javascript.

I really appreciate some feedback.

To use it a promise A+ compatible library has to be made available globally in lua (as it is in javascript) included https://promisesaplus.com/. Since "then" is not available in Lua as a name for the method I changed it and use "next" instead. There are several lua promise-a+ compatible libraries there to test it. Like this one https://github.com/Billiam/promise.lua, or here https://github.com/zserge/lua-promises
The interface for this promise library may look like this: 

`
interface Promise<T> {

  state: 0|1|2|3|4; 

  next<TResult1 = T, TResult2 = never>(
    onfulfilled?: ((value: T)    => TResult1 | Promise<TResult1>) | undefined | null, 
    onrejected?:  ((reason: any) => TResult2 | Promise<TResult2>) | undefined | null
  ): Promise<TResult1 | TResult2>; 
}

interface PromiseConstructor {
  resolve<T>(value: T | Promise<T>):Promise<T>; // create a resolved Promise
  resolve   ():Promise<never>;  // create a resolved Promise
}

declare var Promise: PromiseConstructor; 
`


Perryvw:
Thanks for the PR! We are currently working on a rewrite of the core transpiler structure (almost done!). Until that's done we don't want to accept any large changes on master anymore. Unfortunately that means that you will have to rewrite the Transpiler.ts portion of your pull request once it is done.

In the mean time: could you add some functional tests for this? (And figure out if our testing framework supports coroutines I guess 😄 )

Perryvw:
We also do not accept PRs changing package.json or package-lock.json unless explicitly discussed in the team

tjahn:
Hey, thank you for the response. I am pretty busy currently, but will look into async-await again when you are ready with your rewrite. 
I am pretty sure some more questions will arise. Like what to do with the .lua files and similar.

PR #360, opened 2019-01-26
Fix accessor assignment

TheLartians:
Enables using accessors as expressions.

### DIFF
--- a/src/LuaTransformer.ts
+++ b/src/LuaTransformer.ts
@@ -1797,7 +1797,10 @@ export class LuaTransformer {
             );
         }
 
-        if (ts.isPropertyAccessExpression(expression.left) || ts.isElementAccessExpression(expression.left)) {
+        if (
+            ts.isPropertyAccessExpression(expression.left) && !tsHelper.hasSetAccessor(expression.left, this.checker)
+            || ts.isElementAccessExpression(expression.left)
+        ) {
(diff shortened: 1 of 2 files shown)


src/LuaTransformer.ts
```diff
@@ -1797,7 +1797,10 @@ export class LuaTransformer {
             );
         }
 
-        if (ts.isPropertyAccessExpression(expression.left) || ts.isElementAccessExpression(expression.left)) {
+        if (
+            ts.isPropertyAccessExpression(expression.left) && !tsHelper.hasSetAccessor(expression.left, this.checker)
```
TheLartians: probably should also check for get accessor

src/LuaTransformer.ts
```diff
@@ -1797,7 +1797,10 @@ export class LuaTransformer {
             );
         }
 
-        if (ts.isPropertyAccessExpression(expression.left) || ts.isElementAccessExpression(expression.left)) {
+        if (
```
Perryvw: This is a strange line break

Perryvw: oh nvm I see how it's laid out now, maybe add parentheses around the && statement to indicate order though

TheLartians: ok I added parentheses

PR #362, opened 2019-01-26
named import prefix

TheLartians:
Prevent global shadowing by adding an tstl prefix to imported modules when using named imports.
Example below.

```
import {printer} from "./print";
declare let print;
print("test"); // errors as print has been overridden by named import 
```

### DIFF
--- a/src/LuaTransformer.ts
+++ b/src/LuaTransformer.ts
@@ -177,10 +177,6 @@ export class LuaTransformer {
         return tstl.createDoStatement(this.transformStatements(block.statements), undefined, block);
     }
 
-    public tstlIdentifier(name: string): string {
-        return "__TSTL_" + name;
-    }
-
     public transformImportDeclaration(statement: ts.ImportDeclaration): StatementVisitResult {
         if (!statement.importClause || !statement.importClause.namedBindings) {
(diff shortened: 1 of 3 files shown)


src/LuaTransformer.ts
```diff
@@ -177,6 +177,10 @@ export class LuaTransformer {
         return tstl.createDoStatement(this.transformStatements(block.statements), undefined, block);
     }
 
+    public tstlIdentifier(name: string): string {
```
Perryvw: Having doubts if this should be a function. It would probably be better to just format the string inside the import declaration function, since that is the only place it's used anyway

TheLartians: Ok, I moved the method into a local arrow function. I still like the function as it is more readable and maintainable imo.

Perryvw: Was thinking more along the lines of `const importName = "__TSTL_" + path.basename(importPath)` but this is okay too

TheLartians: Yeah that does make more sense when used only here. On the other hand wouldn't it be nice to have some common prefix for all tstl auxiliary variables?

PR #369, opened 2019-01-28
Hoisting

tomblind:


src/LuaTransformer.ts
```diff
@@ -60,7 +65,6 @@ export class LuaTransformer {
     }
 
     public setupState(): void {
-        this.scopeStack = [];
```
Perryvw: Why was this removed? To better support multiple files?

src/LuaTransformer.ts
```diff
@@ -3267,17 +3304,66 @@ export class LuaTransformer {
         return tstl.createBinaryExpression(expression, tstl.createNumericLiteral(1), tstl.SyntaxKind.AdditionOperator);
     }
 
+    protected findScope(...scopeTypes: ScopeType[]): Scope | undefined {
```
Perryvw: Might be interesting to use a bitmask for this instead of an array.

src/LuaTransformer.ts
Perryvw: Ah, the diff was hiding it!

src/LuaTransformer.ts
tomblind: that's a good idea!

tomblind:
Ok, this PR is probably worth reviewing again.

There are now 3 options for hoisting:
-None : No hoisting is performed at all. Errors are thrown when attempting to reference something before it's declared
- Full : Everything is hoisted. No additional checks required, but produces ugly/slower lua
- Required : Only variables/functions referenced before their declaration or from another hoisted function will be hoisted. Requires compiler overhead, but produces best results.

src/CommandLineParser.ts
```diff
@@ -32,6 +32,13 @@ export const optionDeclarations: YargsOptions = {
         describe: "Specify if a header will be added to compiled files.",
         type: "boolean",
     },
+    hoisting: {
+        alias: "h",
```
lolleko: h is usually help, maybe choice something different or don't set an alias at all.

src/LuaAST.ts
```diff
@@ -139,13 +141,19 @@ export function setParent(node: Node | Node[] | undefined, parent: Node): void
...
         });
     } else {
         node.parent = parent;
         if (node.pos === -1 || node.end === -1) {
             node.pos = parent.pos;
             node.end = parent.end;
         }
+        if (!node.tsOriginal) {
+            node.tsOriginal = parent.tsOriginal;
+        }
```
lolleko: It is intentionally avoided to link the Lua AST node to the tsOriginal. Is this really required?

src/CommandLineParser.ts
tomblind: good point

src/LuaAST.ts
tomblind: I think I can rework things to avoid it, but I'm curious what the reason is for avoiding it? Seems like it would be quite useful.

src/LuaAST.ts
```diff
@@ -791,15 +791,23 @@ export function createMethodCallExpression(
...
+    tsOriginal?: ts.Node,
+    symbol?: ts.Symbol,
+    parent?: Node
+): Identifier
+{
     const expression = createNode(SyntaxKind.Identifier, tsOriginal, parent) as Identifier;
     expression.text = text as string;
+    expression.symbol = symbol;
     return expression;
 }
```
lolleko: Again is it possible to avoid having ts types in the Lua AST?
Sorry I haven't elaborated on why I don't think TS stuff should be in an Lua AST.

Main reason, it's a design choice: Having no TS nodes in the AST clearly separates the transform and print step. Allowing them to run independent of each other. Not having any TS nodes in the AST allows us to easily create  Lua ASTs from the ground up (with no TS used to generate the AST).

I think a good compromise would be do just save the symbolID on the AST node, since that would just be a number and not a ts type. And as far as I can tell you only save the symbol to keep track on what should be hoisted? I think the ID could be enough for that purpose.
Or maybe map the tstl.Identifier to the Symbol somewhere else? For example in a Map<> in the Transformer, that way we wouldn't have to store the information on the node itself. Since you don't need that info in there print step there is no reason to save it on the node tbh.

tomblind: Ok, thanks for the explanation - that makes sense to me. All I need here is a way to differentiate unique symbols, so an ID system like you suggest should work.

src/LuaTransformer.ts
```diff
@@ -3095,6 +3153,34 @@ export class LuaTransformer {
...
+                } else {
+                    symbolId = this.symbolIds.get(symbol);
+                }
+                this.scopeStack.forEach(s => {
+                    if (!s.referencedSymbols) { s.referencedSymbols = new Set(); }
+                    s.referencedSymbols.add(symbolId);
+                });
+            }
+        }
+
```
Perryvw: Probably nice to refactor this block to its own function with a nice descriptive name.

src/LuaTransformer.ts
```diff
@@ -3370,17 +3502,146 @@ export class LuaTransformer {
...
+                        } else {
+                            result.splice(i, 1);
+                        }
+                    } else {
+                        // Special case for 'var's declared in child scopes
+                        this.replaceStatementInParent(declaration, assignment);
+                    }
+                    hoistedLocals.push(...declaration.left);
+                }
+            }
```
Perryvw: This for loop + 3592 would probably also be nice to refactor to its own function

Issue #372, opened 2019-01-29
rootDirs support


dustinlacewell:
Currently, when a transpiled file depends on another typescript file, `tstl` needs to determine:

* Where to emit the corresponding Lua file
* How to `require` the corresponding Lua file

As of now, `tstl` only considers the `rootDir` compiler option when attempting resolve a relative path from the file doing the importing, to the imported file. If the imported file is not actually relative to `rootDir`, the whole absolute path sequence is retained in both emission of the Lua file, and when requiring it.

`tstl` should instead check every path in `rootDirs`. If it finds that the dependency is relative to one of the `rootDirs` it should perform the same "relativizing" it does for `rootDir`.

This will have the advantage of allowing the use of (particularly developement) typescript libraries to be shared across transpiled projects without deeply nested, host-specific emissions arising in `tstl` output.

ark120202:
I've experimented with that (for #432), and it looks like `rootDirs` doesn't affect output structure by design (see https://github.com/microsoft/TypeScript/issues/8245). If there's a need for that, we can:
- Investigate if that use case is covered by modules (yarn workspaces?) or project references
- Document how API can be used to manipulate output structure
- Add a new option

PR #373, opened 2019-01-29
[WIP] Rootdirs support

dustinlacewell:
**Don't Merge!** 

This changeset addresses #372 `rootDirs support`.

Changes `emitSourceFile` and `getImportPath` to attempt relativization against `rootDirs`. This happens only if relativization fails against `rootDir` which is what is currently supported.


Perryvw:
This has gone stale.

PR #380, opened 2019-02-03
Async await

tjahn:
Hello,

Since you finished your rewriting lately (looking good by the way) I am here for the second try to get async-await support into TypescriptToLua.

For that I wrote two functions __TS__Await and __TS__Async to mimic the behaviour of the async and the await-Keyword of Javascript using Lua coroutines.
__TS__Async takes a function and creates another function with same arguments that returns a promise. 
__TS__Await can be called in the context of an async function with a promise as the argument, returning the promised value.

Supported are async functions, async arrow functions and async member functions.

I added a small unit-test to check if async and await are compiled as expected. 
However, I don't really know how to include functional tests into your repository. 
Also I am not sure where to put the files /dist/lualib/Async.lua and /dist/lualib/Await.lua. 
Writing them in typescript is a pain, since many declarations are missing at that point. 

But perhaps you might want to try the approach in the PR anyhow. I added a folder "/async_await_test" with a short demonstration. To test it just compile it with 
"./dist/index.js async_await_test/test.ts"
"lua async_await_test/test.lua"

Looking forward to your feedback


tjahn:
* Yes, providing an own promise library is still necessary. In the test_async_await example I just took one lua implementation, renamed some functions and wrote a type-declaration for that. It is unlikely that I find the time for writing an own one anytime soon. I guess there are probably pure typescript implementations for A+ promises out there, that only require minor modifications (like renaming the then-method to next). https://github.com/poelstra/ts-promise/blob/master/src/lib/Promise.ts this one with MIT license for example.

* I rewrote the lualib files in typescript now

* Using the standard declaration directly is not easily possible, since the standard promise in js uses "then" as a method. However, javascript does not rely on that and it is easily possible to change that. I actually nearly took the standard library and renamed the then-method to "next". Just renaming then to next within tstl is also not an option. In promise+ it is often necessary to check an any-object for the next-method (if it exists its an Thenable).

src/LuaAST.ts
```diff
@@ -438,6 +441,22 @@ export function createLabelStatement(name: string, tsOriginal?: ts.Node, parent?
...
+export function isAwaitExpression(node: Node): node is AwaitExpression {
+  return node.kind === SyntaxKind.AwaitExpression;
+}
+
+export function createAwaitExpression(expression: Expression): AwaitExpression {
+  const expr = createNode(SyntaxKind.AwaitExpression ) as AwaitExpression;
+  setParent(expression, expr);
+  return expr;
+}
+
```
lolleko: Please dont modify the AST.
This is a Lua AST implementation, the idea is to have no TS related stuff in here.
As fas as I can tell you dont even use the added node.

tjahn: You are absolutely right of cause. I removed all modifications to the lua ast

test/unit/async.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+import { TranspileError } from "../../src/TranspileError";
+
+export class AsyncTests {
+
+    @Test("async function declaration")
+    public asyncFunctionDeclaration(): void {
+        const testCode = `
+            async function a() { }
+        `;
+        Expect(util.transpileString(testCode)).toBe(`require("lualib_bundle\");\na = __TS__Async(function()\nend);`);
```
ark120202: This is basically the same as translation tests, which are discouraged ([guidelines](https://github.com/Perryvw/TypescriptToLua/blob/master/CONTRIBUTING.md#testing-guidelines)).

ark120202:
I think it would be better to translate `promise.then` to something like `promise._then` where `promise` has `Promise` type, instead of using .next property. Promises already have established standard, plus that would require users to use custom promise type instead of standard one.

test/unit/async.spec.ts
tjahn: Hey, yes you are right, I just check if async stays in the code. Since it gets ignored in the current master, same as the await-keyword.

Functional tests for async-await are hard to include currently, since the promise-library is not yet shipped within the code.
However, I will probably include some self-written promise.ts into the lualib soon.

tjahn:
> I think it would be better to translate `promise.then` to something like `promise._then` where `promise` has `Promise` type, instead of using .next property. Promises already have established standard, plus that would require users to use custom promise type instead of standard one.

Sounds fine to me, I would probably go with "__TS__then".
The downside is that the promise A+ standard internally checks passed objects for a then-method, what may not work with any-type objects and also interacting with the promises from pure lua might be awkward.

Currently tstl just throws an "Cannot use Lua keyword then as identifier" when trying to use it.

ark120202:
> may not work with any-type objects

That's fine, most of transforms now depend on type and fail for any.

> interacting with the promises from pure lua

Well, it can be named `.next` internally and transpile `.then` to `.next`. Though since there is no real standard in this it won't be possible to be compatible with all possible implementations. For these cases it might be better to make these apis return a different type (`EnvPromise`) and then write a small wrapper that makes them transforms `EnvPromise` to native `Promise`.

src/LuaTransformer.ts
```diff
@@ -1084,13 +1088,21 @@ export class LuaTransformer {
...
+        let functionExpression = tstl.createFunctionExpression(
             tstl.createBlock(body),
             paramNames,
             dots,
             restParamName
         );
 
+        const isAsync = node.modifiers && node.modifiers.some(m => m.kind === ts.SyntaxKind.AsyncKeyword);
+        if( isAsync ) {
+          this.importLuaLibFeature( LuaLibFeature.ArrayForEach );
```
ark120202: Can it be handled in `luaLibDependencies`?

src/LuaTransformer.ts
```diff
@@ -3186,7 +3228,17 @@ export class LuaTransformer {
             return tstl.createCallExpression(this.transformExpression(node.expression), parameters);
         } else {
             // Replace last . with : here
-            const name = node.expression.name.escapedText;
+            let name = node.expression.name.escapedText as string;
+
+            // catch usage of Lua keywords
+            if (this.luaKeywords.has(name)) {
+              if ( this.translateKeywords.has(name) ) {
```
ark120202: It should be done only when `ownerType` is a promise, because in future (#429) it might be valid to use keywords as property names

src/lualib/Promise.ts
```diff
@@ -0,0 +1,335 @@
...
+
+  public catch<TResult = never>(
+    onRejected?: ((reason: any) => TResult | PromiseImpl<TResult>) | undefined
+  ): PromiseImpl<T | TResult> {
+    return this.then( undefined, onRejected );
+  }
+
+  protected resolve( value: T ): void {
+    PromiseImpl.PromiseResolutionProcedure<T>( this, value );
+  }
```
ark120202: ES2018 also has `promise.finally`, which might be implemented as well

src/LuaTransformer.ts
tjahn: The promises in the es6 standard explicitly allow arbitrary objects with "then" method to be handled as a PromiseLike. 

#429 suggests  `const then = 1` -> `local ____then = 1`. I chose __TS__then instead.
As soon as #429 is there my current translation to __TS__then is obsolete.

ark120202: In #429 `____then` would be used only for identifiers. In this case it would be `promise["then"]`. Of course that works, but I see a point in using some renamed property because of performance. `promise:__TS__then(...)` would be better than `promise["then"](promise, ...)`, because `promise` may be non-pure value, and thus be compiled to iife.

src/lualib/Promise.ts
```diff
@@ -0,0 +1,335 @@
+declare function error( value:string, level?:number ): never;
+const logError: any = print;
```
ark120202: In `inline` lib mode all variables in lib's scope would be included as they are, so they should be prepended with `____` to avoid potential conflicts.

src/lualib/Promise.ts
```diff
@@ -0,0 +1,335 @@
...
+
+let runAtNextTick: ( func: () => void ) => void = func => func();
+function setRunAtNextTickFunction(  func: (f: () => void ) => void ): void {
```
ark120202: IMO it would be better to:
a) `Promise.runAtNextTick` (overridable)
b) `Promise.setRunAtNextTickFunction`
c) `Promise._setRunAtNextTickFunction`
d) Introduce some namespace for non-standard stuff, like `tstl.setRunAtNextTickFunction`

src/lualib/Promise.ts
```diff
@@ -0,0 +1,335 @@
...
+    return p;
+  }
+
+  public static reject( reason: any ): PromiseImpl<never> {
+    const p = new PromiseImpl();
+    p.reject( reason );
+    return p;
+  }
+
+  protected static create<T = undefined>(): PromiseImpl<T> {
```
ark120202: It's not a part of promise standard and doesn't look like it's used by something internally. Should it be removed?

src/lualib/Promise.ts
```diff
@@ -0,0 +1,335 @@
...
+  }
+
+  protected __gc(): void {
+    if ( this.state === PromiseState.REJECTED && !this.handledError ) {
+      for ( const d of this.queue ) {
+        if ( d.onRejected ) {
+          return;
+        }
+      }
+      logError("Unhandled rejected promise ", this.reason );
```
ark120202: Might be useful to provide something like `unhandledrejection` event in js

src/lualib/Promise.ts
tjahn: I chose option c). To set the tick-function you can no use something like this in your *.ts files:
```typescript
interface Promise {
   _setRunAtNextTickFunction( func: ( func: () => void ) => void ): void;
}
Promise._setRunAtNextTickFunction( ( func ) => func() );
```

src/lualib/Promise.ts
tjahn: Since there is not window Object with event handling etc I chose an approach similar to the _setAtNextTick( ) function:

```typescript
interface Promise {
  _setUnhandledRejectionCallback( func: ( reason: string ) => void ): void;
}
Promise._setUnhandledRejectionCallback( ( reason ) => print(reason) );
```

Perryvw:
I'm closing this as this has gone stale. We do still want to add support for async/await, but we want to first look very carefully at how we actually want to support this feature.

PR #383, opened 2019-02-08
Classes update

### LINKED ISSUE #256: Breaking classes with method names
```ts
class Foo {
    public method() { return "foo"; }
    public static method() { return "bar"; }
}
const bar = new Foo();
print(bar.method());
print(Foo.method());
```
```lua
Foo = Foo or {}
Foo.__index = Foo
function Foo.new(construct, ...)
    local self = setmetatable({}, Foo)
    if construct and Foo.constructor then Foo.constructor(self, ...) end
    return self
end
function Foo.constructor(self)
end
function Foo.method(self)
    return "foo"
end
function Foo.method(self) --Overrides previous definition!!
    return "bar"
end
local bar = Foo.new(true);
print(bar:method()); --prints "bar"!
print(Foo:method());
```

A common solution to this in lua bindings to other languages is to split classes into 2 tables: a class table with static methods and properties, and a separate metatable that provides all of the instance methods. Doing this would also get things closer to what's needed for #206.

Bonus problem:
```ts
class Foo {
    public new() {}
}
const foo = new Foo();
```
```lua
Foo = Foo or {}
Foo.__index = Foo
function Foo.new(construct, ...)
    local self = setmetatable({}, Foo)
    if construct and Foo.constructor then Foo.constructor(self, ...) end
    return self
end
function Foo.constructor(self)
end
function Foo.new(self) --Uh-oh...
end
local foo = Foo.new(true);
```
Perhaps adding new to the list of dis-allowed identifier names would be a good idea (also, maybe 'self' too).

### LINKED ISSUE #351: Prevent extenions/metaextension classes from being extended or used with instanceof
Classes decorated with @extension and @metaExtension should not be allowed to be extended or have instanceof called on them, since they are not TSTL-compiled classes.

```ts
declare class Foo {}

/** @extension */
declare class Bar extends Foo {}

class Baz extends Bar {} // No bueno

declare const obj: any;
if (obj instanceof Bar) {} // Also not a good idea
```

### LINKED ISSUE #353: Calling "super" from a class multiple layers of inheritance deep causes infinite recursion
`my_class.ts`
```typescript
export class MyClass {
    do_thing() {
        let x = 1 + 2
    }
}
```

`sub_my_class.ts`
```typescript
import { MyClass } from "./my_class";

export class SubMyClass extends MyClass {
    do_thing() {
        super.do_thing()
    }
}
```

`sub_sub_my_class.ts`
```typescript
import { SubMyClass } from "./sub_my_class";

export class SubSubMyClass extends SubMyClass {
    do_thing() {
        super.do_thing()
    }
}
```

`main.ts`
```typescript
import { SubSubMyClass } from "./sub_sub_my_class";

const my_class = new SubSubMyClass();

my_class.do_thing();
```

This transpiles to:
`my_class.lua`
```lua
local exports = exports or {}
local MyClass = MyClass or {}
MyClass.__index = MyClass
function MyClass.new(construct, ...)
    local self = setmetatable({}, MyClass)
    if construct and MyClass.constructor then MyClass.constructor(self, ...) end
    return self
end
function MyClass.constructor(self)
end
function MyClass.do_thing(self)
    local x = 1+2;
end
exports.MyClass = MyClass
return exports
```

`sub_my_class.lua`
```lua
local exports = exports or {}
local my_class0 = require("src.my_class")
local MyClass = my_class0.MyClass
local SubMyClass = SubMyClass or MyClass.new()
SubMyClass.__index = SubMyClass
SubMyClass.__base = MyClass
function SubMyClass.new(construct, ...)
    local self = setmetatable({}, SubMyClass)
    if construct and SubMyClass.constructor then SubMyClass.constructor(self, ...) end
    return self
end
function SubMyClass.do_thing(self)
    self.__base.do_thing(self);
end
exports.SubMyClass = SubMyClass
return exports
```

`sub_sub_my_class.lua`
```lua
local exports = exports or {}
local sub_my_class0 = require("src.sub_my_class")
local SubMyClass = sub_my_class0.SubMyClass
local SubSubMyClass = SubSubMyClass or SubMyClass.new()
SubSubMyClass.__index = SubSubMyClass
SubSubMyClass.__base = SubMyClass
function SubSubMyClass.new(construct, ...)
    local self = setmetatable({}, SubSubMyClass)
    if construct and SubSubMyClass.constructor then SubSubMyClass.constructor(self, ...) end
    return self
end
function SubSubMyClass.do_thing(self)
    self.__base.do_thing(self);
end
exports.SubSubMyClass = SubSubMyClass
return exports
```

`main.lua`
```lua
local sub_sub_my_class0 = require("src.sub_sub_my_class")
local SubSubMyClass = sub_sub_my_class0.SubSubMyClass
local my_class = SubSubMyClass.new(true);
my_class:do_thing();
```

`lua build/main.lua` throws the following:

```
lua: ./src/sub_my_class.lua:13: stack overflow
stack traceback:
        ./src/sub_my_class.lua:13: in field 'do_thing'
        ./src/sub_my_class.lua:13: in field 'do_thing'
        ./src/sub_my_class.lua:13: in field 'do_thing'
        ./src/sub_my_class.lua:13: in field 'do_thing'
        ./src/sub_my_class.lua:13: in field 'do_thing'
        ./src/sub_my_class.lua:13: in field 'do_thing'
```

---

What we actually want is for the transpilation to produce:
`sub_my_class.lua`
```lua
function SubMyClass.do_thing(self)
    MyClass.do_thing(self);
end
```

because `self.__base` when called from the child refers to the child's base, which is the current class, which causes infinite recursion.

tomblind:
fixes #256 
fixes #351 
fixes #353 

This PR reworks classes to use separate static and instance tables. It also adds errors for mis-uses of extension and metaExtension decorated classes.

Some notes:
- The instance table is assigned to "ClassName.prototype", and that table has a reference back through a "constructor" property. This is compatible with how JS classes are structured.
- The ~~"new" and~~ "constructor" functions were renamed with leading underscores to prevent potential name collisions
- "__base" was renamed to "____super" for clarity
- "super" method and constructor calls now have two ways they can be transpiled. A short version: "BaseClass.prototype", which is the default, and a long version: "DerivedClass.____super.prototype" which is used when a class extends from something other than a simple identifier ("class Derived extends makeBase() {}")

Basic Example:
```ts
class Foo {
    prop = "prop";
    constructor() {}
    method() {}
    static staticProp = "staticProp";
    static staticMethod() {}
}
const foo = new Foo();
```
```lua
Foo = Foo or {};
Foo.__index = Foo;
Foo.prototype = Foo.prototype or {};
Foo.prototype.__index = Foo.prototype;
Foo.prototype.constructor = Foo;
Foo.new = function(...)
    local self = setmetatable({}, Foo.prototype);
    self:____constructor(...);
    return self;
end;
Foo.staticProp = "staticProp";
Foo.prototype.____constructor = function(self)
    self.prop = "prop";
end;
Foo.prototype.method = function(self)
end;
Foo.staticMethod = function(self)
end;
local foo = Foo.new();
```

Sub-class example:
```ts
class Bar extends Foo {
    constructor() { super(); }
    method() { super.method(); }
}
const bar = new Bar();
```
```lua
Bar = Bar or {};
Bar.__index = Bar;
Bar.prototype = Bar.prototype or {};
Bar.prototype.__index = Bar.prototype;
Bar.prototype.constructor = Bar;
Bar.____super = Foo;
setmetatable(Bar, Bar.____super);
setmetatable(Bar.prototype, Bar.____super.prototype);
Bar.new = function(...)
    local self = setmetatable({}, Bar.prototype);
    self:____constructor(...);
    return self;
end;
Bar.prototype.____constructor = function(self)
    Foo.prototype.____constructor(self);
end;
Bar.prototype.method = function(self)
    Foo.prototype.method(self);
end;
local bar = Bar.new();
```


src/LuaTransformer.ts
```diff
@@ -40,8 +40,9 @@ interface Scope {
 
 export class LuaTransformer {
     public luaKeywords: Set<string> = new Set([
-        "and", "break", "do",  "else", "elseif", "end",    "false",  "for",  "function", "if",
-        "in",  "local", "nil", "not",  "or",     "repeat", "return", "then", "until",    "while",
+        "and", "break", "do",  "else", "elseif", "end", "false",  "for",    "function",     "if",
+        "in",  "local", "new", "nil", "not",  "or",     "repeat", "return", "self", "then", "until",
```
Perryvw: some strange spacing in this array actually

tomblind: agreed (I was just emulating what was there)

tomblind: oh, but I messed it up... 😢

PR #384, opened 2019-02-08
Generator functions

andreiradu:
~~One particular issue is that calling `iterator.next(42)` transpiles to `iterator:next(42)`, which in turn means that the corresponding `coroutine.yield()` will return `iterator, 42`. I've made it so `let a = yield()` transpiles to `local _, a = coroutine.yield()` to account for that.~~
```ts
function* seq() {
    let a = yield 1;
    return a;
}
const gen = seq();
```
will transpile to:
```lua
seq = function()
    local __co = coroutine.create(function()
        local a = coroutine.yield(1);
        return a;
    end);
    __it = {next = function(___, ...)
        local __err, __value = coroutine.resume(__co, ...);
        if not __err then
            error(__value);
        end
        return {done = coroutine.status(__co) == "dead", value = __value};
    end};
    __it[Symbol.iterator] = function()
        return __it;
    end;
    return __it;
end;
local gen = seq();
```


src/LuaAST.ts
```diff
@@ -813,6 +813,12 @@ export function createIdentifier(
     return expression;
 }
 
+export function createAnnonymousIdentifier(tsOriginal?: ts.Node, parent?: Node): Identifier {
+    const expression = createNode(SyntaxKind.Identifier, tsOriginal, parent) as Identifier;
+    expression.text = "___";
```
tomblind: this should be 4 underscores for consistency

src/LuaTransformer.ts
```diff
@@ -1031,6 +1030,143 @@ export class LuaTransformer {
...
+        nextBody.push(tstl.createReturnStatement([iteratorResult]));
+
+        //function(__, ...)
+        const nextFunctionDeclaration = tstl.createFunctionExpression(
+            tstl.createBlock(nextBody),
+            [tstl.createAnnonymousIdentifier()],
+            tstl.createDotsLiteral());
+
+        //__it = {next = function(__, ...)}
+        const iterator = tstl.createAssignmentStatement(
```
tomblind: Shouldn't this be a VariableDeclarationStatement so `__it` is marked local?

src/LuaTransformer.ts
```diff
@@ -1031,6 +1030,143 @@ export class LuaTransformer {
...
+                        [tstl.createReturnStatement([itIdentifier])]
+                    )
+                )
+            ),
+            //return __it
+            tstl.createReturnStatement([itIdentifier]),
+        ];
+        return [block, functionScope];
+    }
     public transformFunctionDeclaration(functionDeclaration: ts.FunctionDeclaration): StatementVisitResult {
```
tomblind: Can we get some line breaks between functions? 😄

src/TSHelper.ts
```diff
@@ -178,6 +178,25 @@ export class TSHelper {
             return false;
         }
     }
+    public static isInGeneratorFunction(node: ts.Node, checker: ts.TypeChecker): boolean {
```
tomblind: Is this used anywhere (maybe I'm not seeing it)?

src/TSHelper.ts
```diff
@@ -178,6 +178,25 @@ export class TSHelper {
...
+            (n): n is ts.Node => ts.isFunctionDeclaration(n) || ts.isMethodDeclaration(n)
+        );
+        if (declaration && ts.isFunctionDeclaration(declaration) && declaration.asteriskToken) {
+            return true;
+        } else {
+            return false;
+        }
+    }
+
+    public static isGeneratorYieldCall(node: ts.Node, checker: ts.TypeChecker): boolean {
```
tomblind: Also not seeing where this is used...

test/unit/functions.spec.ts
```diff
@@ -388,4 +388,35 @@ export class FunctionTests {
         const result = util.transpileAndExecute(code);
         Expect(result).toBe("foobar");
     }
+    @Test("Generator functions")
+    public generatorFunction(): void {
```
tomblind: It seems like this should be broken into separate test functions.

src/LuaTransformer.ts
andreiradu: Done :)

src/LuaTransformer.ts
```diff
@@ -1218,7 +1361,11 @@ export class LuaTransformer {
 
         return tstl.createExpressionStatement(this.transformExpression(expression));
     }
-
```
tomblind: I know it seems minor, but spacing between functions helps a lot with readability

src/LuaTransformer.ts
```diff
@@ -1218,7 +1361,11 @@ export class LuaTransformer {
...
+    public transformYield(expression: ts.YieldExpression): tstl.Expression {
+        return tstl.createCallExpression(
+            tstl.createTableIndexExpression(tstl.createIdentifier("coroutine"), tstl.createStringLiteral("yield")),
+                expression.expression?[this.transformExpression(expression.expression)]:[], expression);
+    }
     public transformReturn(statement: ts.ReturnStatement): tstl.Statement {
```
tomblind: Same as above

src/LuaTransformer.ts
andreiradu: Done

PR #392, opened 2019-02-10
Fixed enums with identifier values not being supported.

### LINKED ISSUE #318: Support reverse mappings for enums
We should consider suporting reverse mappings for enums:

https://www.typescriptlang.org/docs/handbook/enums.html#Enums%20at%20runtime

### LINKED ISSUE #378: Issue with initializing enums with identifiers
```
enum A {
    Test1,
    Test2,
    Test3 = Test1
}
```
fails with
``` 
Error: Only numeric or string initializers allowed for enums.                                    sformer.js:120:20)
    at Function.TSTLErrors.InvalidEnumMember (D:\Work\typescript_to_lua\TypescriptToLua\dist\TSTLsformer.js:49:33)Errors.js:13:42)
    at node.members.map.member (D:\Work\typescript_to_lua\TypescriptToLua\dist\LuaTransformer.js:549:51)
    at Array.map (<anonymous>)
    at LuaTransformer.computeEnumMembers (D:\Work\typescript_to_lua\TypescriptToLua\dist\LuaTransformer.js:537:29)
    at LuaTransformer.transformEnumDeclaration (D:\Work\typescript_to_lua\TypescriptToLua\dist\LuaTransformer.js:516:39)
 ```

### PR DESCRIPTION
**Pull request by Perryvw**
Added reverse mapping for enums, allowing:
```ts
const myEnum {
    abc,
    def,
}
const x = myEnum.def; // x = 1
const y = myEnum[x]; // y = "def"
```

Also added support for `constEnum["enumMemberName"]` and refactored the const enum handling stuff a little.

Fixes #378 and closes #318

### DIFF
--- a/src/LuaTransformer.ts
+++ b/src/LuaTransformer.ts
@@ -1011,13 +1011,13 @@ export class LuaTransformer {
                 if (ts.isNumericLiteral(member.initializer))
                 {
                     numericValue = Number(member.initializer.text);
-                    valueExpression = tstl.createNumericLiteral(numericValue);
+                    valueExpression = this.transformNumericLiteral(member.initializer);
                     numericValue++;
                 }
                 else if (ts.isStringLiteral(member.initializer))
                 {
                     hasStringInitializers = true;
-                    valueExpression = tstl.createStringLiteral(member.initializer.text);
+                    valueExpression = this.transformStringLiteral(member.initializer);
                 }
                 else
                 {
--- a/src/LuaTransformer.ts
+++ b/src/LuaTransformer.ts
@@ -2749,26 +2749,7 @@ export class LuaTransformer {
         }
 
         if (type.symbol && (type.symbol.flags & ts.SymbolFlags.ConstEnum)) {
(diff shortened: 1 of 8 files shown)


src/LuaTransformer.ts
```diff
@@ -986,47 +986,71 @@ export class LuaTransformer {
...
+                    valueExpression = tstl.createNumericLiteral(numericValue);
+                    numericValue++;
+                }
+                else if (ts.isStringLiteral(member.initializer))
+                {
                     hasStringInitializers = true;
-                    valueLiteral = tstl.createStringLiteral(member.initializer.text);
-                } else {
-                    throw TSTLErrors.InvalidEnumMember(member.initializer);
+                    valueExpression = tstl.createStringLiteral(member.initializer.text);
```
tomblind: would it be better to use this.transformStringLiteral() here?

Perryvw: Yep, that is nicer. I also changed the numeric literal above this one to do the same.

PR #400, opened 2019-02-13
Fixed use of length operator on array unions and intersections.

tomblind:
Also, removed restriction on assigning properties to arrays.


test/unit/array.spec.ts
```diff
@@ -36,6 +46,21 @@ export class ArrayTests {
         Expect(result).toBe(5);
     }
```
Perryvw: Could you add a test assigning some other property now we allow that

PR #401, opened 2019-02-14
Compatible getters/setters 

### LINKED ISSUE #206: Property getters/setters 'lost' when accessed from interface
Properties with custom getters/setters in typescript only work in lua if accessed from a variable with the classes type. If accessed from a variable with an interface that maps to the class, the lua tries to access the property directly instead of calling the getter/setter.

Example:
```ts
interface Foo
{
    prop: string;
}

class Bar
{
    private _prop = "BAR";
    public get prop() { return this._prop; }
    public set prop(value: string) { this._prop = value; }
}

let bar = new Bar();
print(bar.prop);

let foo: Foo = bar;
print(foo.prop);
```
Produces:
```lua
Bar = Bar or {}
Bar.__index = Bar
function Bar.new(construct, ...)
    local instance = setmetatable({}, Bar)
    instance._prop = "BAR"
    if construct and Bar.constructor then Bar.constructor(instance, ...) end
    return instance
end
function Bar.constructor(self)
end
function Bar.get__prop(self)
    return self._prop
end
function Bar.set__prop(self,value)
    self._prop = value;
end
local bar = Bar.new(true);
print(bar:get__prop()); --prints "BASE"
local foo = bar;
print(foo.prop); --prints nil
```

One possible solution would be to use a custom metatable for classes with properties like these:
```lua
Bar = Bar or {}
Bar__meta =
{
    getters =
    {
        prop = function(self)
            return self._prop
        end
    },
    setters =
    {
        prop = function(self, value)
            self._prop = value
        end
    },
    __index = function(self, key)
        local getter = Bar__meta.getters[key]
        if getter then
            return getter(self)
        end
        return rawget(self, key) or Bar[key]
    end,
    __newindex = function(self, key, value)
        local setter = Bar__meta.setters[key]
        if setter then
            setter(self, value)
        else
            rawset(self, key, value)
        end
    end
}
function Bar.new(construct, ...)
    local instance = setmetatable({}, Bar__meta)
    instance._prop = "BAR"
    if construct and Bar.constructor then Bar.constructor(instance, ...) end
    return instance
end
function Bar.constructor(self)
end
local bar = Bar.new(true);
print(bar.prop);
local foo = bar;
print(foo.prop);
```

BTW, sorry for all the bugs. I have a large lua project which would really benefit from static typing so I'm putting tstl through it's paces to see if it will work for what I need. :)

tomblind:
fixes #206 

This only affects classes that have getters and/or setters. I went with the solution here: https://github.com/Perryvw/TypescriptToLua/issues/206#issuecomment-462748760

One edge-case to note is this:
```ts
class Foo {
    prop = "foo";
}
class Bar extends Foo {
    get prop() { return "bar"; }
}
print((new Bar()).prop);
```
Because prop is assigned directly to the instance in the base class, the getter defined in the derived class would never be accessed. My solution to this is to detect this specific case and set the instance's property to back nil in derived class' constructor.
```lua
Bar.prototype.____constructor = function(self, ...)
    Foo.prototype.____constructor(self, ...);
    rawset(self, "prop", nil);
end;
```
Since the base property is effectively inaccessible in this situation, I don't think this should cause any undesirable side-effects.


src/TSHelper.ts
```diff
@@ -399,6 +345,80 @@ export class TSHelper {
...
+            c => c.members.some(ts.isSetAccessor),
+            checker
+        ) !== undefined;
+    }
+
+    public static getPropertyName(propertyName: ts.PropertyName): string | number | undefined {
+        if (ts.isIdentifier(propertyName) || ts.isStringLiteral(propertyName)) {
+            return propertyName.text;
+        } else if (ts.isNumericLiteral(propertyName)) {
+            return Number(propertyName.text);
```
Perryvw: Is this line intentional? I would expect a function ***Name function to return strings

test/unit/accessors.spec.ts
```diff
@@ -0,0 +1,146 @@
...
+        Expect(util.transpileAndExecute(code)).toBe("bar");
+    }
+
+    @Test("get accessor overridden")
+    public getAccessorOverridden(): void {
+        const code =
+            `class Foo {
+                get foo() { return "foo"; }
+            }
+            class Bar extends Foo {
```
Perryvw: You have a test base with accessor, extension without, base without, extension with accessor. Could you also add one where they both have a get accessor?

test/unit/expressions.spec.ts
```diff
@@ -249,7 +249,7 @@ export class ExpressionTests {
     @TestCase("inst.field | 3", 8 | 3)
     @TestCase("inst.field << 3", 8 << 3)
     @TestCase("inst.field >> 1", 8 >> 1)
-    @TestCase("inst.field = 3", 7)
```
Perryvw: Is this test right? We should verify if typescript returns the literal number or the field getter.

src/TSHelper.ts
tomblind: It was intentional. I wanted to distinguish between string "0" and numerical 0 (the name was just being consistant with the ts.PropertyName type). But I just checked and it turns out TS treats those keys as being identical, so this is not necessary.

test/unit/expressions.spec.ts
tomblind: This was previously returning the wrong value. When an assignment is evaluated, the value on the right is returned. The left should not be evaluated. [Playground](https://www.typescriptlang.org/play/#src=class%20Foo%20%7B%0D%0A%20%20%20%20get%20prop()%20%7B%20return%201%3B%20%7D%0D%0A%20%20%20%20set%20prop(val%3A%20number)%20%7B%7D%0D%0A%7D%0D%0Aconst%20foo%20%3D%20new%20Foo()%3B%0D%0Aalert(foo.prop%20%3D%207)%3B%0D%0A)

Issue #342, opened 2019-01-18
Data structure clear and garbage collection


ehemsley:
I noticed that the clear functions transpile to the following:

```lua
function Map.clear(self)
    self.items = {};
    self.size = 0;
    return
end
```

This is fine from a computation performance perspective, however throwing away the table creates a huge amount of pressure when dealing with garbage collection. 

In contrast,
```lua
function Map.clear(self)
  for k in pairs(self.items)
    self.items[k] = nil
  end
  self.size = 0
  return
end
```
might be moderately slower, but creates no garbage from throwing away the table. 

Garbage collection is a huge issue when it comes to programming games, so perhaps there should be a way to specify which clear method is used?

tomblind:
This was implemented as an external lib, so I'm closing it.

PR #404, opened 2019-02-15
Rewrite Map and Set clear to not generate garbage

ehemsley:
Addresses #342

### DIFF
--- a/src/lualib/Map.ts
+++ b/src/lualib/Map.ts
@@ -34,7 +34,9 @@ class Map<TKey, TValue> {
     }
 
     public clear(): void {
-        this.items = {};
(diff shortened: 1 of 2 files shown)


tomblind:
I definitely understand the desire for this (garbage collection can be murder on fps). But, this could dramatically slow down situations with very large collections. Is it a good idea to make it the default?

ehemsley:
That is definitely a valid concern, I mostly wanted to get the ball rolling and ask how we might go about providing this as an option for users.

ehemsley:
Closing this, going to provide GC optimized collection functionality as a separate library.

PR #406, opened 2019-02-15
Weak collections

ark120202:
Tests don't check weak part, because Fengari not supports it.


ark120202:
@tomblind not unless it would be runned on real Lua VM, Fengari not implements gc.

PR #402, opened 2019-02-14
Added Object.keys and Object.assign

Perryvw:
Closes #151

PR #407, opened 2019-02-15
Custom symbols

ark120202:
`typeof Symbol()` still returns `object`, since I don't see much need in this right now.

Uses the same code as #402 to use exnext `Symbol.description` type in tests.

### DIFF
--- a/src/lualib/Symbol.ts
+++ b/src/lualib/Symbol.ts
@@ -1,3 +1,19 @@
+declare function setmetatable<T extends object>(obj: T, metatable: any): T;
+
+const symbolMetatable = {
+    __tostring(): string {
+        if (this.description === undefined) {
+            return 'Symbol()';
+        } else {
+            return 'Symbol(' + this.description + ')';
+        }
+    },
+};
+
+function __TS__Symbol(description?: string | number): symbol {
+    return setmetatable({ description }, symbolMetatable) as any;
+}
+
 Symbol = {
-    iterator: {},
+    iterator: __TS__Symbol('Symbol.iterator'),
 } as any;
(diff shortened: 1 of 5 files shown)


test/unit/lualib/symbol.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+    @TestCase("name")
+    public symbolDescription(description?: string | number): void
+    {
+        const result = util.transpileAndExecute(`
+            return Symbol(${JSON.stringify(description)}).description;
+        `, this.compilerOptions);
+
+        Expect(result).toBe(description);
+    }
+
```
Perryvw: Could you add a test that shows symbols are unique, for example `Symbol("foo") === Symbol("foo")` being false?

src/lualib/Symbol.ts
```diff
@@ -1,3 +1,19 @@
+declare function setmetatable<T extends object>(obj: T, metatable: any): T;
+
+const symbolMetatable = {
```
tomblind: This should probably have leading underscores (we usually use 4) to prevent name collisions.

src/lualib/SymbolRegistry.ts
```diff
@@ -0,0 +1,15 @@
+const symbolRegistry: Record<string, symbol> = {};
```
tomblind: Same as symbolMetatable above.

src/lualib/Symbol.ts
ark120202: Oh, right. They compile to locals, so it's not a problem in bundle, but that might cause issues with inlining.

PR #410, opened 2019-02-16
numeric for loops

### LINKED ISSUE #170: For loops
Thought it'd be nice to generate plain `for` loops as a special case. 😄 What do y'all think?

TypeScript:

```ts
for (let i = 1; i <= 10; i++) {
  // ...
}
```

Lua:

```lua
for i=1,10 do
  -- ...
end
```

tomblind:
closes #170 

Numeric for loops that are compatible with lua's `for i = 1, 10 do` syntax are detected and transpiled as such.

Also fixes scoping issue with for loops that are transpiled to while loops
```ts
let i = 37;
for (let i = 1; i < 8; i *=2) {} // using *= to force using while in transpilation
print(i); // should be 37, but was 8
```

"Your scientists were so preoccupied with whether or not they could, they didn’t stop to think if they should."


tomblind:
I just realized there's a big fundamental flaw in this, so I'm killing it for now.

PR #426, opened 2019-02-20
Dot slash path fix

### LINKED ISSUE #425: ExportDeclaration require string cut off on Windows
```ts
export { GCOptimizedMap } from "./src/gc_optimized_map";
export { GCOptimizedSet } from "./src/gc_optimized_set";
```

produces

```lua
local exports = exports or {};
do
    local __TSTL_gc_optimized_map = require("rc.gc_optimized_map");
    local GCOptimizedMap = __TSTL_gc_optimized_map.GCOptimizedMap;
    exports.GCOptimizedMap = GCOptimizedMap;
end
do
    local __TSTL_gc_optimized_set = require("rc.gc_optimized_set");
    local GCOptimizedSet = __TSTL_gc_optimized_set.GCOptimizedSet;
    exports.GCOptimizedSet = GCOptimizedSet;
end
return exports;
```

hazzard993:
Fixes #425

I use `path` to parse and format both paths so that there is no leading `/` and the slashes are both the same kind.

Added some tests as well to make sure this stops happening.

### DIFF
--- a/src/LuaTransformer.ts
+++ b/src/LuaTransformer.ts
@@ -3811,7 +3811,7 @@ export class LuaTransformer {
     }
 
     private getImportPath(relativePath: string): string {
-        const rootDir = this.options.rootDir && path.resolve(this.options.rootDir) || path.resolve(".");
+        const rootDir = this.options.rootDir ? path.resolve(this.options.rootDir) : path.resolve(".");
         const absoluteImportPath = path.format(path.parse(this.getAbsoluteImportPath(relativePath)));
         const absoluteRootDirPath = path.format(path.parse(rootDir));
         if (absoluteImportPath.includes(absoluteRootDirPath)) {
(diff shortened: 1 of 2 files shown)


src/LuaTransformer.ts
```diff
@@ -3811,12 +3811,12 @@ export class LuaTransformer {
     }
 
     private getImportPath(relativePath: string): string {
-        const rootDir = this.options.rootDir || path.resolve(".");
-        const absoluteImportPath = this.formatPathToLuaPath(this.getAbsoluteImportPath(relativePath));
-        const absoluteRootDirPath = this.formatPathToLuaPath(rootDir);
+        const rootDir = this.options.rootDir && path.resolve(this.options.rootDir) || path.resolve(".");
```
Perryvw: I'm not entirely sure what I'm looking at but it looks like a lua ternary conditional? In TS those look like: `this.options.rootDir ? path.resolve(this.options.rootDir) : path.resolve(".")`

hazzard993: Been writing too much Lua, I'll turn that into a normal ternary

PR #428, opened 2019-02-20
Support special characters in module names

### LINKED ISSUE #403: Imports that use hyphen in filename break when transpiled to Lua
example:

```ts
import { vectorlight } from "lib-defs/hump/vector-light";
```

transpiles to

```lua
local __TSTL_vector-light = require("lib-defs.hump.vector-light");
local vectorlight = __TSTL_vector-light.vectorlight;
```

which throws the error:

```
expected statement near '-'
```

ehemsley:
Lua hates hyphens in variable names, so this fixes cases where an imported TypeScript module is named in kebab case.

Fixes #403


hazzard993:
Would it be ok to extend this PR to handle some other invalid characters in variable names?

You can get away with single quotes, hashes, spaces, $ as well as hypens in file names and Lua and TypeScript are ok with it. Lua just gets upset with the resulting variable names.

Extending your RegExp you can use this:

```ts
const tstlIdentifier = (name: string) => "__TSTL_" + name
    .replace(new RegExp("-|\\$| |#|'", "g"), "_");
```

The translation tests could also be extended to cover these cases.

```ts
import {TestClass} from "kebab-module"
import {TestClass} from "dollar$module"
import {TestClass} from "singlequote'module"
import {TestClass} from "hash#module"
import {TestClass} from "space module"
```

```ts
import {TestClass as RenamedClass} from "kebab-module"
import {TestClass as RenamedClass} from "dollar$module"
import {TestClass as RenamedClass} from "singlequote'module"
import {TestClass as RenamedClass} from "hash#module"
import {TestClass as RenamedClass} from "space module"
```

ehemsley:
I guess there's no harm in being thorough.

ehemsley:
This now works with different special characters and not just hyphen.

PR #431, opened 2019-02-20
Deprecate case insensitivity of compiler annotations

### LINKED ISSUE #422: Support only standard jsdoc annotation casing
It doesn't really add anything but inconsistency.
Would be a breaking change, so we should just show the warning now and stop supporting it with #365.

ark120202:
Closes #422.


Perryvw:
As explained in #422, we are not adding this since jsDoc itself is case insensitive.

PR #445, opened 2019-02-23
tupleReturn improvements

tomblind:
- fixed `@tupleReturn` on arrow functions
```ts
const fn = /** @tupleReturn */ s => [s, s];
```

- enabled inference of tupleReturn on arrow functions
```ts
/** @tupleReturn */ interface Fn { (s: string): [string, string] }
const fn: Fn = s => [s, "bar"]; // the arrow function will now correctly keep return values unpacked
```

- preventing wrap+unpack when using spread operator on tupleReturn function
```ts
/** @tupleReturn */ declare function foo(): [string, string];
declare function bar(a: string, b: string): void;
bar(...foo()); // will no longer be 'bar(table.unpack({foo()}))'
```

### DIFF
--- a/src/LuaTransformer.ts
+++ b/src/LuaTransformer.ts
@@ -2976,9 +2976,9 @@ export class LuaTransformer {
         if (ts.isBlock(node.body)) {
             body = node.body;
         } else {
-            const ret = ts.createReturn(node.body);
-            body = ts.createBlock([ret]);
-            ret.parent = body;
+            const returnExpression = ts.createReturn(node.body);
+            body = ts.createBlock([returnExpression]);
+            returnExpression.parent = body;
             body.parent = node.body.parent;
         }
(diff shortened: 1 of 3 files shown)


src/LuaTransformer.ts
```diff
@@ -2972,7 +2972,15 @@ export class LuaTransformer {
             hasContext ? context : undefined
         );
 
-        const body = ts.isBlock(node.body) ? node.body : ts.createBlock([ts.createReturn(node.body)]);
+        let body: ts.Block;
+        if (ts.isBlock(node.body)) {
+            body = node.body;
+        } else {
+            const ret = ts.createReturn(node.body);
```
Perryvw: This could use a better variable name 😄

tomblind: Oops - I had meant to go back and rename that before 😄

PR #446, opened 2019-02-23
Made boolean CLI option values optional

Perryvw:
If no value is supplied, the value defaults to true.

### DIFF
--- a/src/CommandLineParser.ts
+++ b/src/CommandLineParser.ts
@@ -10,6 +10,10 @@ type ParseResult<T> =
     { isValid: true; result: T }
     | { isValid: false, errorMessage: string};
 
+type ArgumentParseResult<T> =
+    { isValid: true; result: T; increment?: number }
+    | { isValid: false, errorMessage: string };
+
 interface ParsedCommandLine extends ts.ParsedCommandLine {
     options: CompilerOptions;
 }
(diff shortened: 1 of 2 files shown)


src/CommandLineParser.ts
```diff
@@ -232,13 +238,24 @@ function parseTSTLOptions(commandLine: ts.ParsedCommandLine, args: string[]): CL
...
+    argumentIndex: number,
+    args: string[]
+): ArgumentParseResult<string | boolean>
 {
+    const option = optionDeclarations[argumentName];
+    const argument = args[argumentIndex + 1];
+
+    if (option.type === "boolean" && (argument === undefined || argument.startsWith("-"))) {
+        // Set boolean arguments without supplied value to true
+        return { isValid: true, result: true, increment: 0 };
```
tomblind: Would it be better for `result` to be `option.default` here (in case a flag option is added that defaults to true if not specified)?

Perryvw: I guess it should be `!option.default`? Otherwise setting the option without value would be the same as not setting it at all. I'm a little concerned that's just confusing. Would it ever make sense to have an option that is true by default and can be set to false by just adding its name to CLI? I'm not sure that is a valid use case.

tomblind: Yeah, actually you're probably right. If you set a flag up like that, it could just be inverted.

PR #449, opened 2019-02-24
lua string instance methods

tomblind:
Currently, tstl doesn't allow lua's string methods to be declared and used:
```ts
interface String { upper(): string; }
print("foo".upper()); // error
```

This PR explicitly enables lua's builtin string methods.

### DIFF
--- a/src/LuaTransformer.ts
+++ b/src/LuaTransformer.ts
@@ -3515,6 +3515,35 @@ export class LuaTransformer {
                 const firstParamPlusOne = this.expressionPlusOne(params[0]);
                 return this.createStringCall("byte", node, caller, firstParamPlusOne);
             }
+            case "byte":
+            case "char":
+            case "dump":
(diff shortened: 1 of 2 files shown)


src/LuaTransformer.ts
```diff
@@ -3515,6 +3515,35 @@ export class LuaTransformer {
...
+            case "len":
+            case "lower":
+            case "match":
+            case "pack":
+            case "packsize":
+            case "rep":
+            case "reverse":
+            case "sub":
+            case "unpack":
+            case "upper":
```
Perryvw: Should we allow upper to be redefined? TS already has `toUpperCase()` that should translate to this? Same for several others of these.

tomblind: Personally, I'd like to be able to disable the es libs altogether and just use lua's in some circumstances.

PR #450, opened 2019-02-25
implemented more math function and value conversions

### LINKED ISSUE #438: Math.round isn't transpiled properly
dunno is there any lua's version, but currently transpiled to math.round, which doesn't exist

tomblind:
fixes #438 

Since I had to restructure things a bit for Math.round, I went ahead and implemented some of the other low-hanging-fruit math function conversions.

### DIFF
--- a/src/LuaTransformer.ts
+++ b/src/LuaTransformer.ts
@@ -3331,7 +3331,7 @@ export class LuaTransformer {
             case "LOG2E":
             case "SQRT1_2":
             case "SQRT2":
-                return tstl.createNumericLiteral(Math[name]);
+                return tstl.createNumericLiteral(Math[name], identifier);
 
             default:
                 throw TSTLErrors.UnsupportedProperty("math", name, identifier);
--- a/src/LuaTransformer.ts
+++ b/src/LuaTransformer.ts
@@ -3123,11 +3123,7 @@ export class LuaTransformer {
         const ownerType = this.checker.getTypeAtLocation(node.expression.expression);
 
         if (ownerType.symbol && ownerType.symbol.escapedName === "Math") {
-            return tstl.createCallExpression(
-                this.transformMathExpression(node.expression.name),
-                this.transformArguments(node.arguments),
-                node
-            );
+            return this.transformMathCallExpression(node);
         }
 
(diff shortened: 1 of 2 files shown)


src/LuaTransformer.ts
```diff
@@ -3320,34 +3316,99 @@ export class LuaTransformer {
...
-            sqrt: "sqrt",
-            tan: "tan",
-        };
+    public transformMathExpression(identifier: ts.Identifier): tstl.Expression {
+        const name = identifier.escapedText as string;
+        switch (name) {
+            case "PI":
+                const property = tstl.createStringLiteral("pi");
+                const math = tstl.createIdentifier("math");
+                return tstl.createTableIndexExpression(math, property, identifier);
```
Perryvw: Why are you treating pi different from E?

src/LuaTransformer.ts
```diff
@@ -3320,34 +3316,99 @@ export class LuaTransformer {
...
+
+            case "E":
+            case "LN10":
+            case "LN2":
+            case "LOG10E":
+            case "LOG2E":
+            case "SQRT1_2":
+            case "SQRT2":
+                return tstl.createNumericLiteral(Math[name]);
```
Perryvw: You should add the original TS node here

test/unit/math.spec.ts
```diff
@@ -8,7 +8,13 @@ export class MathTests {
     @TestCase("Math.cos()", "math.cos();")
```
Perryvw: Missing test for constants (probably would be nice to actually `util.transpileAndExecute` those)

src/LuaTransformer.ts
tomblind: Since lua has it's own constant for pi, it's probably best to use that.

Perryvw: Ah, if that's the reasoning I'm fine with that 👍

test/unit/math.spec.ts
tomblind: There are tests for `Math.PI` and `Math.E`. Should I add a test for every one?

Also, I'm not sure we should do execute tests since there could be inconsistencies with precision.

PR #452, opened 2019-02-25
Added initial support for sourcemaps

### LINKED ISSUE #237: Source Maps
Right now error reporting for the generated lua files is pretty bad; you have to open the lua file, check the specific line and guess what generated it. Debugging also suffers from the same problem - you have to debug the generated lua files. 
When transpiling to js, the typescript compiler can also generate a .map file which would solve all this issues. I propose we do something similar but with a simplified .map format. The one ts uses([here](https://docs.google.com/document/d/1U1RGAehQwRypUTovF1KRlpiOFze0b-_2gc6fAH0KY0k/edit?hl=en_US&pli=1&pli=1#heading=h.mofvlxcwqzej)) is aimed at minimizing file size, which I don't feel is as important here. 
I'm probably going to end up doing this anyway, but it's a very intrusive change, so I would prefer for it to be merged back.

lolleko:
Closes #237 

Sourcemap implementation using Mozillas source map lib (https://github.com/mozilla/source-map). The high-level SourceNode API is used in this draft.

This adds a second runtime dependency, but I think implementing our own version really isn't worth it.
Particularly the part where we would have to write our own Base64 VLQ encoder.

TODOs

- [x] Discuss the implementation (dependencies...)
- [x] Address TODOs in code
- [x] Cleanup API (LuaPrinter & LuaTranspiler)
- [x] Figure out a way to test this
- [x] Add some tests
- [x] Figure out if we can support inlineSourceMaps?


src/LuaPrinter.ts
```diff
@@ -82,11 +110,27 @@ export class LuaPrinter {
         return this.currentIndent + input;
     }
 
-    private printBlock(block: tstl.Block): string {
-        return this.ignoreDeadStatements(block.statements).map(s => this.printStatement(s)).join("");
+    private createSourceNode(
```
Perryvw: After fiddling around with this a little I found that defining this function as follows leads to much nicer code:
```ts
private createSourceNode(node: tstl.Node, chunks: Array<(string | SourceNode)>): SourceNode {
    const line = node.line !== undefined ? node.line + 1 : undefined;
    const column = node.column !== undefined ? node.column + 1 : undefined;

    return new SourceNode(line, column, this.sourceFile, chunks);
}
```

src/LuaPrinter.ts
```diff
@@ -117,132 +161,222 @@ export class LuaPrinter {
...
-        return result;
+        return this.createSourceNode(statement.line, statement.column, this.sourceFile, chunks);
     }
 
-    private printVariableDeclarationStatement(statement: tstl.VariableDeclarationStatement): string {
-        const left = this.indent(`local ${statement.left.map(e => this.printExpression(e)).join(", ")}`);
+    private printVariableDeclarationStatement(statement: tstl.VariableDeclarationStatement): SourceNode {
+        const chunks: Array<(string | SourceNode)> = [];
+        chunks.push("local ");
+        statement.left.forEach((e, i) => {
```
Perryvw: Instead of repeating this each time, we should just make a `function join(elements: Array<string | SourceNode>, separator: string): Array<string | SourceNode>` to do this for us. Would also greatly clean this up.

Perryvw:
I think the "Clean up APIs" point probably deserves its own PR, I think for now it might be nicest if we just always generate and return the source map from the printer (if that's possible?) and then let whatever is calling it (in the current case the LuaTransformer) choose to emit it or not.

We can then look at cleaning up the API together with the rest of the project.

Perryvw:
It might also be a nice idea to define `type SourceChunk = string | SourceNode` and use that instead of `string | SourceNode` every time.

test/unit/sourcemaps.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+
+// Helper functions
+
+function expectCorrectMapping(
+    original: string,
+    lua: string,
+    sourceMap: {[line: string]: number},
+    patterns: Array<[string, string]>
+): void {
+    for (const [tsPattern, luaPattern] of patterns) {
```
ark120202: These variables aren't used in test

src/LuaAST.ts
```diff
@@ -110,22 +109,25 @@ export interface Node extends TextRange {
...
     return Object.assign({}, node);
 }
 
 export function setNodeOriginal<T extends Node>(node: T, tsOriginal: ts.Node): T {
-    node.pos = tsOriginal.pos;
-    node.end = tsOriginal.end;
+    const sourcePosition = getSourcePosition(tsOriginal);
+    if (sourcePosition) {
+        node.line = sourcePosition.line;
+        node.line = sourcePosition.line;
```
ark120202:
```suggestion
        node.column = sourcePosition.column;
```

PR #443, opened 2019-02-22
Console functions

hazzard993:
For #439

Added some functions from `console` that translate to Lua. `log`, `assert` and `trace`.

```ts
console.log("Hello");               // print("Hello")
console.log("Hello %s", "there");   // prints "Hello there" in Lua
console.log("Hello", "There");      // print("Hello", "There")
console.assert(true);               // assert(true)
console.assert(true, "message");    // assert(true, "message")
console.assert(false, "Hello %s", "there"); // Fails with "Hello there" as a message
console.trace();                    // Prints a stack trace in Lua
console.trace("message");           // Prints a stack trace with "message" at the top in Lua
console.trace("Hello %s", "there"); // Prints a stack trace with "Hello there" at the top in Lua
```

Issue #439, opened 2019-02-21
Transpile console.log(...) as print(...)


Perryvw:
Transpiling `console.log(...)` as `print(...)` would improve the reusability of code between different environments (such as node + lua).

PR #454, opened 2019-02-26
Console function calls

hazzard993:
Continued from #443 as a partial solution for #439.

```ts
console.log("Hello");               // print("Hello")
console.log("Hello %s", "there");   // prints "Hello there" in Lua
console.log("Hello", "There");      // print("Hello", "There")
console.assert(true);               // assert(true)
console.assert(true, "message");    // assert(true, "message")
console.assert(false, "Hello %s", "there"); // Fails with "Hello there" as a message
console.trace();                    // Prints a stack trace in Lua
console.trace("message");           // Prints a stack trace with "message" at the top in Lua
console.trace("Hello %s", "there"); // Prints a stack trace with "Hello there" at the top in Lua
```

For the case below, the code doesn't work.

```ts
console.log("foo%%bar", "blah"); // Prints foo%bar instead of foo%%bar    blah
```

I looked into implementing these functions into LuaLibs since what happens is figured out at runtime, however something like:

```ts
function __TSTL__ConsoleLog(...args) {
    print(...args);
}
```

Transpiles to lua with `unpack`. Which means the solution won't work for all lua targets.

This is only a partial implementation of `console.log`, `console.assert` and `console.trace` since the full solution would likely be implemented in a LuaLib.

Submitting this PR just to show what I have and see if it is acceptable for now.

### DIFF
--- a/src/LuaTransformer.ts
+++ b/src/LuaTransformer.ts
@@ -3138,6 +3138,10 @@ export class LuaTransformer {
             return this.transformObjectCallExpression(node);
         }
 
+        if (ownerType.symbol && ownerType.symbol.escapedName === "Console") {
+            return this.transformObjectCallExpression(node);
+        }
+
         if (ownerType.symbol && ownerType.symbol.escapedName === "SymbolConstructor") {
             return this.transformSymbolCallExpression(node);
         }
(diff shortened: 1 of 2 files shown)


src/LuaTransformer.ts
```diff
@@ -3666,6 +3670,91 @@ export class LuaTransformer {
...
+                    }
+                }
+                // assert()
+                return tstl.createCallExpression(
+                    tstl.createIdentifier("assert"),
+                    args
+                );
+            case "trace":
+                if (expression.arguments.length > 0) {
+                    if (this.isStringFormatTemplate(expression.arguments[0])) {
```
Perryvw: Might as well merge these two if statements into one with a `&&` (JS short-circuits if the left hand side fails, so it functions exactly the same)

src/LuaTransformer.ts
```diff
@@ -3666,6 +3670,91 @@ export class LuaTransformer {
...
+                        // print(debug.traceback(string.format([arguments])))
+                        return tstl.createCallExpression(
+                            tstl.createIdentifier("print"),
+                            [tstl.createCallExpression(
+                                tstl.createTableIndexExpression(
+                                    tstl.createIdentifier("debug"),
+                                    tstl.createStringLiteral("traceback")),
+                                [tstl.createCallExpression(
```
Perryvw: This formatting is a little confusing, maybe adding some intermediate variables would help.

PR #455, opened 2019-02-26
No path resolution for ambient modules

hazzard993:
TypeScript allows the following.

**mod.d.ts**
```ts
declare module "mod" {
    export let x = 5;
}
```

**src/main.ts**
```ts
import * as module from "mod";
console.log(module.x);
```

`tstl` makes the require path `src.mod` because it tried to resolve the path.
`tsc` doesn't resolve these, it stays as `mod`.

This PR it to make sure that paths to ambient modules are not resolved.


src/LuaTransformer.ts
```diff
@@ -315,14 +315,23 @@ export class LuaTransformer {
...
             result.push(tstl.createExpressionStatement(requireCall));
             return result;
         }
 
         const imports = statement.importClause.namedBindings;
+        const type = this.checker.getTypeAtLocation(imports);
+        let requireCall: tstl.CallExpression;
+        if (type && this.isAmbientModuleDeclaration(type)) {
+            requireCall = tstl.createCallExpression(
+                tstl.createIdentifier("require"),
```
Perryvw: Might be nicer to delegate this logic to `createModuleRequire` by adding an optional `resolveModule: bool = true`  parameter.

src/LuaTransformer.ts
```diff
@@ -376,6 +385,10 @@ export class LuaTransformer {
         }
     }
 
+    public isAmbientModuleDeclaration(type: ts.Type): boolean {
+        return type.symbol && type.symbol.valueDeclaration.kind === ts.SyntaxKind.ModuleDeclaration;
```
tomblind: Typically, these types of functions are put in TSHelper

ark120202:
I don't think it's good to give `declare module` statement a different meaning. In ts it's valid to type multiple modules in a single .d.ts file, plus it may be used to annotate files by patterns with a `*` token. I think it would be better to add a new annotation (on declaration) for this.

src/TSHelper.ts
```diff
@@ -76,6 +76,10 @@ export class TSHelper {
         return extendedTypeNode && checker.getTypeAtLocation(extendedTypeNode);
     }
 
+    public static isAmbientModuleDeclaration(type: ts.Type): boolean {
+        return type.symbol && type.symbol.valueDeclaration.kind === ts.SyntaxKind.ModuleDeclaration;
```
Perryvw: After some discussion in the discord we came to the conclusion that it would be a better solution to add a compiler directive `@noResolution` on the module, to minimize the risk of this breaking something.

test/unit/require.spec.ts
```diff
@@ -33,4 +33,14 @@ export class RequireTests {
         }
     }
 
+    @Test("modules do not have their require paths adjusted")
+    public testRequireModule(): void {
+        const regex = /require\("(.*?)"\)/;
+        const lua = util.transpileString(`
+            import * as fake from "fake";
+            declare module "fake" {}
```
Perryvw: Does this test do anything? You didn't decorate your module with the decorator. (Also the name is misleading)

hazzard993: Nope it did not. It doesn't look like you can declare a module and import it in the same file. I've made a change to the StringCompilerProgram to allow multiple files to be considered during compilation so I can put through two files to try out this decorator.

test/unit/require.spec.ts
```diff
@@ -33,4 +33,20 @@ export class RequireTests {
         }
     }
 
+    @TestCase("", "src.fake")
+    @TestCase("/** @noResolution */", "fake")
+    @Test("noResolution on ambient modules causes no path alterations")
+    public testRequireModule(comment: string, expectedPath: string): void {
+        const regex = /require\("(.*?)"\)/;
```
Perryvw: Try to lay out your tests according to AAA (Arrange/Act/Assert), this regex is part of the last (assertion) step, so I'd move this to right before where you actually use it.

src/Compiler.ts
```diff
@@ -89,7 +89,10 @@ const defaultCompilerOptions: CompilerOptions = {
 };
 
 export function createStringCompilerProgram(
-    input: string, options: CompilerOptions = defaultCompilerOptions, filePath = "file.ts"): ts.Program {
+    input: string,
+    options: CompilerOptions = defaultCompilerOptions,
+    filePath = "file.ts",
+    extraFiles?: { [filename: string]: string }): ts.Program {
```
Perryvw: I kinda dislike this last parameter, I think it would be nicer if you change the type of `input` to be `string | {[fileName: string]: string}`, and then use a typeof check in the function what you are actually doing. (Maybe you could even remove the last parameter - depending on how much that breaks - since you can then do the same with the first parameter).

PR #457, opened 2019-02-26
implement Array.prototype.findIndex. closes #456

### LINKED ISSUE #456: Error: Unsupported property on array: findIndex
TypeScript code:

```typescript
const arr = ["x", "y", "z"];
arr.findIndex((item) => item === "y");
```

```
Error: Unsupported property on array: findIndex
```

endel:
this should fix #456, let me know if you have any feedback!

### DIFF
--- a/src/lualib/ArrayFindIndex.ts
+++ b/src/lualib/ArrayFindIndex.ts
@@ -1,0 +1,17 @@
+function __TS__ArrayFindIndex<T>(
+    arr: T[],
+    callbackFn: (this: void, element: T, index?: number, array?: T[]) => boolean
+): number {
+    const len = arr.length;
+    if (len === 0) {
+        return -1;
+    }
+
+    for (let i = 0; i < len; i++) {
+        if (callbackFn(arr[i], i, arr)) {
+            return i;
+        }
+    }
+
+    return -1;
+}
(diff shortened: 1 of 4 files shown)


src/lualib/ArrayFindIndex.ts
```diff
@@ -0,0 +1,17 @@
+function __TS__ArrayFindIndex<T>(
+    arr: T[],
+    callbackFn: (this: void, element: T, index?: number, array?: T[]) => boolean
+): number {
+    const len = arr.length;
+    if (len === 0) {
+        return -1;
+    }
```
Perryvw: These 4 lines are probably overkill, line 6-8 definitely are.

test/unit/lualib/lualib.spec.ts
```diff
@@ -21,6 +21,23 @@ export class LuaLibTests
...
 
+    @TestCase([0, 2, 4, 8], 10, -1)
+    @TestCase([0, 2, 4, 8], 4, 2)
+    @TestCase([0, 2, 4, 8], 8, 3)
+    @Test("findIndex")
+    public findIndex(inp: number[], searchEl: number, expected: number): void
+    {
+        const result = util.transpileAndExecute(
+            `let arrTest = ${JSON.stringify(inp)};
+            return JSONStringify(arrTest.findIndex((elem, index) => {
```
Perryvw: Would be nice if we could also have a test checking if the index and array parameters of the callback are set correctly by the lualib function.

src/lualib/ArrayFindIndex.ts
endel: right

test/unit/lualib/lualib.spec.ts
endel: I've updated, let me know what you think

PR #458, opened 2019-02-26
Fix array.sort compare function not working

ark120202:


test/unit/lualib/lualib.spec.ts
```diff
@@ -352,6 +352,22 @@ export class LuaLibTests
         // Assert
         Expect(result).toBe(JSON.stringify(expected));
     }
+
+    @TestCase("[4, 5, 3, 2, 1]", [1, 2, 3, 4, 5], "a - b")
+    @TestCase('["4", "5", "3", "2", "1"]', ["1", "2", "3", "4", "5"], "Number(a) - Number(b)")
+    @TestCase('["4", "5", "3", "2", "1"]', ["5", "4", "3", "2", "1"], "Number(b) - Number(a)")
+    @Test("array.sort with compare function")
+    public arraySortWithCompareFunction(array: string, expected: any, compare: string): void
```
Perryvw: I would prefer tests where you compare the lua result with `(a, b) => array.sort(a - b)` in typescript (even if that means splitting this up into multiple tests)

PR #464, opened 2019-03-02
Fix call expression transform with undefined signature

ark120202:
`typeChecker.getResolvedSignature` returns `Signature | undefined`.

### DIFF
--- a/src/LuaTransformer.ts
+++ b/src/LuaTransformer.ts
@@ -3100,10 +3100,10 @@ export class LuaTransformer {
         }
 
         const callPath = this.transformExpression(node.expression);
-        if (signature
-            && signature.getDeclaration()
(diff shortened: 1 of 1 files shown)


src/LuaTransformer.ts
```diff
@@ -3100,10 +3100,10 @@ export class LuaTransformer {
         }
 
         const callPath = this.transformExpression(node.expression);
-        const signatureDeclaration = signature.getDeclaration();
-        if (signatureDeclaration
+        if (signature
+            && signature.getDeclaration()
```
tomblind: Not a big deal, but this could better done as:
```ts
const signatureDeclaration = signature && signature.getDeclaration();
```
to avoid double-calling `getDeclaration()`

tomblind:
I swear I fixed this once. Must have been in a branch that wasn't merged.

PR #465, opened 2019-03-02
New functions v2

tomblind:
-all functions default to having a `self` parameter
-`@noSelf` can be added to namespace or class to set the default within that scope (recursive)
-function assignment tests re-written in a more 'procedural' way, making them more thorough, but more numerous
-fixed an issue in generators where parameters were being hidden by redundant definitions

### DIFF
--- a/src/TSHelper.ts
+++ b/src/TSHelper.ts
@@ -554,22 +554,24 @@ export class TSHelper {
                 ? ContextType.Void
                 : ContextType.NonVoid;
         }
-        if (ts.isMethodDeclaration(signatureDeclaration) || ts.isMethodSignature(signatureDeclaration)) {
-            // Method
-            return ContextType.NonVoid;
+
+        let scopeDeclaration: ts.Declaration = signatureDeclaration;
+        while (true) {
+            scopeDeclaration = TSHelper.findFirstNodeAbove(
+                scopeDeclaration,
+                (n): n is ts.ModuleDeclaration | ts.ClassLikeDeclaration =>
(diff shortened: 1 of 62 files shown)


src/TSHelper.ts
```diff
@@ -554,22 +554,24 @@ export class TSHelper {
...
-        }
-        if (ts.isBinaryExpression(signatureDeclaration.parent)) {
-            // Function expression: check type being assigned to
-            return TSHelper.getFunctionContextType(
-                checker.getTypeAtLocation(signatureDeclaration.parent.left), checker);
-        }
-        return ContextType.Void;
+
+        let scopeDeclaration: ts.Declaration = signatureDeclaration;
+        while (true) {
```
Perryvw: Probably nice to refactor this loop into its own `TSHelper.hasNoSelfAncestor`

src/lualib/ArrayFindIndex.ts
```diff
@@ -1,6 +1,7 @@
 function __TS__ArrayFindIndex<T>(
+    this: void,
     arr: T[],
-    callbackFn: (this: void, element: T, index?: number, array?: T[]) => boolean
+    callbackFn: (element: T, index?: number, array?: T[]) => boolean
```
Perryvw: So this callback inherits the void context from the function definition? Maybe it's me, but this doesn't seem very intuitive to me.

tomblind: No, it's not inherited. I set the callback to take as self, so regularly defined functions can be passed in here. Otherwise any non-expression functions used as callbacks would need the `this: void` which would get annoying.

PR #466, opened 2019-03-02
Made some adjustments to source maps

Perryvw:
Adjusted the branch with the suggestions I put on your PR, plus a bunch of extra fixes that make source maps a bit more functional :D

### DIFF
--- a/src/LuaAST.ts
+++ b/src/LuaAST.ts
@@ -99,27 +99,14 @@ export type Operator = UnaryOperator | BinaryOperator;
 
 export type SymbolId = number;
 
-export interface TextRange {
-    line?: number;
-    column?: number;
-}
-
-export interface Node extends TextRange {
+export interface Node {
     kind: SyntaxKind;
     parent?: Node;
+    tsOriginal?: ts.Node;
 }
 
 export function createNode(kind: SyntaxKind, tsOriginal?: ts.Node, parent?: Node): Node {
-    let line: number | undefined;
-    let column: number | undefined;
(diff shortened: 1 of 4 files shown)


src/LuaAST.ts
```diff
@@ -99,37 +99,22 @@ export type Operator = UnaryOperator | BinaryOperator;
...
-export interface TextRange {
-    line?: number;
-    column?: number;
-}
-
-export interface Node extends TextRange {
+export interface Node {
     kind: SyntaxKind;
     parent?: Node;
+    tsOriginal?: ts.Node;
```
lolleko: I already discussed this with @tomblind a while ago. I think it is better not to keep TS types in the AST, at least from a design point of view.

src/LuaAST.ts
```diff
@@ -140,17 +125,25 @@ export function setParent(node: Node | Node[] | undefined, parent: Node): void
...
-        }
+    }
+}
+
+export function getOriginalPos(node: Node): { line: number, column: number } {
+    while (node.tsOriginal === undefined && node.parent !== undefined) {
+        node = node.parent;
+    }
+
+    if (node.tsOriginal !== undefined && node.tsOriginal.getSourceFile() !== undefined && node.tsOriginal.pos >= 0) {
```
lolleko: I had a TODO in the original PR. it should be investigated why node.tsOriginal.getSourceFile() can return undefined instead of catching that case.

Perryvw: From what I could see working on this, it is probably because we create some ts nodes ourself in the transformer.

PR #467, opened 2019-03-04
Fixed crash related to function expression type inference

tomblind:
This one is oddly specific. The transpiler would fail to infer the type of a function expression that was being assigned to a property in an object literal that was itself, being assigned to a type that didn't contain that property. In some circumstances, this would cause a crash.

Example:
```ts
let foo: {} = {bar: s => s};
```

### DIFF
--- a/src/TSHelper.ts
+++ b/src/TSHelper.ts
@@ -484,7 +484,10 @@ export class TSHelper {
             const objType = this.inferAssignedType(expression.parent.parent, checker);
             const property = objType.getProperty(expression.parent.name.getText());
             if (!property) {
-                return objType.getStringIndexType();
+                const stringPropertyType = objType.getStringIndexType();
+                if (stringPropertyType) {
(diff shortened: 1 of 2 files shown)


src/TSHelper.ts
```diff
@@ -484,7 +484,10 @@ export class TSHelper {
             const objType = this.inferAssignedType(expression.parent.parent, checker);
             const property = objType.getProperty(expression.parent.name.getText());
             if (!property) {
-                return objType.getStringIndexType();
+                const stringPropertyType = objType.getStringIndexType();
+                if (stringPropertyType) {
+                    return stringPropertyType;
+                }
```
Perryvw: Isn't this equivalent to what it was before? Or does it return false

tomblind: This ensures that if there is no string index type, it falls back to the default, which is to return the type of the original expression itself (see end of function)

PR #468, opened 2019-03-04
check origin for standard library types

TheLartians:
Avoids name collisions with user types and standard library, such as `Math`, `Console`, etc. Example:

```typescript
export class Console { 
  test() { return 42; }
}

function test() {
  const console = new Console();
  return console.test();
}

console.log(test()); // prints 42
```

### DIFF
--- a/src/LuaTransformer.ts
+++ b/src/LuaTransformer.ts
@@ -47,6 +47,8 @@ export class LuaTransformer {
 
     private checker: ts.TypeChecker;
     protected options: CompilerOptions;
+    protected program: ts.Program;
+
     private isModule: boolean;
 
     private currentSourceFile?: ts.SourceFile;
@@ -62,8 +64,6 @@ export class LuaTransformer {
     private symbolInfo: Map<tstl.SymbolId, SymbolInfo>;
     private symbolIds: Map<ts.Symbol, tstl.SymbolId>;
     private genSymbolIdCounter: number;
-
-    private program: ts.Program;
 
     private readonly typeValidationCache: Map<ts.Type, Set<ts.Type>> = new Map<ts.Type, Set<ts.Type>>();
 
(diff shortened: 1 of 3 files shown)


lolleko:
Could you add your example as test?

TheLartians:
Just realised that it won't work to simply check if the `Console` in question was globally scoped (the solution I proposed actually just deactivated console when not in global scope). The solution I came up with now is to check if the type is actually declared in `lib.dom.d.ts`.

TheLartians:
As it turns out we can generalise this for all standard types using `ts.Program.isSourceFileDefaultLibrary`. Doesn't seem to apply to `*Constructor` types though.🤔

src/LuaTransformer.ts
```diff
@@ -63,11 +63,14 @@ export class LuaTransformer {
     private symbolIds: Map<ts.Symbol, tstl.SymbolId>;
     private genSymbolIdCounter: number;
 
+    private program: ts.Program;
```
Perryvw: Can you make this protected and put it next to the other constructor variables (below `options`)

src/TSHelper.ts
```diff
@@ -684,6 +684,20 @@ export class TSHelper {
...
+        const source = declaration.getSourceFile();
+        if (!source) { return false; }
+        return program.isSourceFileDefaultLibrary(source);
+    }
+
+    public static isStandardLibraryType(type: ts.Type, name: string, program: ts.Program): boolean {
+        const symbol = type.symbol;
+        if (!symbol || symbol.escapedName !== name) { return false; }
+        const declaration = symbol.valueDeclaration;
+        if(!declaration) { return false; }
```
tomblind: I think the best way to deal with *Constructor types would be to default to `true` here. If I override one of those types, there will be a `valueDeclaration` set on the symbol here. So if there's no `valueDeclaration` and the names match, then it's probably safe to assume it IS the lib version.
```ts
class StringConstructor {
	test() {}
}
function test() {
	const String = new StringConstructor();
	print(String.test()); // symbol.valueDeclaration will point to my override
}
```

TheLartians: Good idea, that way at least collisions become very unlikely.

PR #469, opened 2019-03-08
Function inference fixes

tomblind:
This PR fixes inferring the 'self' type on function expressions when assigned to unions, `as`-casts and type assertions:
```ts
type Fn = (this: any, s: string) => string;
const foo: string | Fn = (s => s);
const bar = (s => s) as Fn;
const baz = <Fn>(s => s);

### CODE REVIEW

**Comment by tomblind** [comment]
Also added a fix for constructor arguments:
```ts
type Fn = (this: any, s: string) => string;
class C {
    constructor(fn: Fn) {}
}
const foo = new C(s => s);
```

PR #475, opened 2019-03-10
Feature/small bugfixes

### LINKED ISSUE #471: Shift operators mixed up
In JS/TS:

`<<` is logical

`>>` arithmetic 

`>>>` logical

and `<<<` doesnt exist

we currently translate `>>` wrong

Perryvw:
Fixed an issue where `const foo = class Bar {}` would translate as:
```lua
local foo = (function()
    local Bar = Bar or {};
    -- ...
    return ____;
end)()
```
The function will now just return `Bar`.

Also swapped around the meaning of right shift operators. Fixes #471

### DIFF
--- a/src/LuaPrinter.ts
+++ b/src/LuaPrinter.ts
@@ -30,8 +30,8 @@ export class LuaPrinter {
         [tstl.SyntaxKind.BitwiseAndOperator]: "&",
         [tstl.SyntaxKind.BitwiseOrOperator]: "|",
         [tstl.SyntaxKind.BitwiseExclusiveOrOperator]: "~",
-        [tstl.SyntaxKind.BitwiseRightShiftOperator]: ">>",
-        [tstl.SyntaxKind.BitwiseArithmeticRightShift]: ">>>",
+        [tstl.SyntaxKind.BitwiseArithmeticRightShift]: ">>",
+        [tstl.SyntaxKind.BitwiseRightShiftOperator]: ">>>",
         [tstl.SyntaxKind.BitwiseLeftShiftOperator]: "<<",
         [tstl.SyntaxKind.BitwiseNotOperator]: "~",
     };
--- a/src/LuaPrinter.ts
+++ b/src/LuaPrinter.ts
@@ -30,7 +30,6 @@ export class LuaPrinter {
         [tstl.SyntaxKind.BitwiseAndOperator]: "&",
         [tstl.SyntaxKind.BitwiseOrOperator]: "|",
         [tstl.SyntaxKind.BitwiseExclusiveOrOperator]: "~",
-        [tstl.SyntaxKind.BitwiseArithmeticRightShift]: ">>>",
         [tstl.SyntaxKind.BitwiseRightShiftOperator]: ">>",
         [tstl.SyntaxKind.BitwiseLeftShiftOperator]: "<<",
         [tstl.SyntaxKind.BitwiseNotOperator]: "~",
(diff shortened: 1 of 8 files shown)


src/LuaTransformer.ts
```diff
@@ -2812,7 +2821,7 @@ export class LuaTransformer {
                 return this.transformBinaryBitLibOperation(node, left, right, operator, "bit");
 
             default:
-                if (operator === tstl.SyntaxKind.BitwiseArithmeticRightShift) {
+                if (operator === tstl.SyntaxKind.BitwiseRightShiftOperator) {
                     throw TSTLErrors.UnsupportedForTarget("Bitwise >>> operator", this.options.luaTarget, node);
```
```suggestion
                     throw TSTLErrors.UnsupportedForTarget("Bitwise >> operator", this.options.luaTarget, node);
```

Perryvw: Incorrect, >> is the arithmetic right shift (which is supported), >>> is the logical right shift (not supported).

src/LuaTransformer.ts
```diff
@@ -2664,6 +2664,15 @@ export class LuaTransformer {
         }
     }
 
+    public transformClassExpression(expression: ts.ClassExpression): ExpressionVisitResult {
+        const className = expression.name !== undefined
+            ? this.transformIdentifier(expression.name)
+            : tstl.createIdentifier("____");
```
tomblind: Same as above

src/LuaTransformer.ts
ark120202: `>>>` in JS is `>>` in Lua
`>>` in JS has no direct replacement in Lua

lolleko: no  `>>` in js is arithemtic which is `>>>` in lua

ark120202: There is no `>>>` in lua 5.3

Perryvw: >> is arithmetic in JS, 5.3 ONLY has arithmetic bit shift operators, so we accidentally got it right the first time for 5.3

lolleko: No 5.3 only has logical shifts at least according to the docs
> Both right and left shifts fill the vacant bits with zeros.
https://www.lua.org/manual/5.3/manual.html

* so `>>>` should be transpiled to `>>`
* `<<` to `<<`
* `>>` is not supported

not sure what we currently do tbh
what a confusing issue

src/LuaPrinter.ts
```diff
@@ -30,8 +30,8 @@ export class LuaPrinter {
         [tstl.SyntaxKind.BitwiseAndOperator]: "&",
         [tstl.SyntaxKind.BitwiseOrOperator]: "|",
         [tstl.SyntaxKind.BitwiseExclusiveOrOperator]: "~",
-        [tstl.SyntaxKind.BitwiseRightShiftOperator]: ">>",
         [tstl.SyntaxKind.BitwiseArithmeticRightShift]: ">>>",
```
ark120202: Might be good to remove `BitwiseArithmeticRightShift` from AST, since it's not really a valid token

tomblind: Seconded - it doesn't make sense to have AST kinds for things that don't exist in lua.

src/LuaTransformer.ts
```diff
@@ -2316,23 +2317,32 @@ export class LuaTransformer {
...
-
+            case ts.SyntaxKind.AmpersandToken:
+            case ts.SyntaxKind.BarToken:
+            case ts.SyntaxKind.CaretToken:
+            case ts.SyntaxKind.LessThanLessThanToken:
+            case ts.SyntaxKind.GreaterThanGreaterThanToken:
+            case ts.SyntaxKind.GreaterThanGreaterThanGreaterThanToken:
+                return this.transformBinaryBitOperation(tsOriginal, left, right, operator);
+            case ts.SyntaxKind.PlusToken:
+                if (ts.isBinaryExpression(tsOriginal)) {
```
tomblind: Is this necessary here? Won't transformBinaryOperator already do this?

PR #484, opened 2019-03-16
validating arguments passed to lua lib functions

### LINKED ISSUE #481: Function conversion check doesn't work on lualib functions
```ts
function foo(this: object, value: string) {
  print(value);
}
function bar(this: void, value: string) {
  print(value);
}

const array = ['a'];
array.forEach(foo);
array.forEach(bar);
```

This code doesn't throw compilation error, but has invalid runtime behavior. The same code, but with `const array: ReadonlyArray<string> = ['a'];` (#480) correctly errors, so it's caused by a missing check during handling of library functions.

tomblind:
fixes #481


src/LuaTransformer.ts
```diff
@@ -3537,9 +3538,9 @@ export class LuaTransformer {
         throw TSTLErrors.CouldNotFindEnumMember(enumDeclaration, memberName, tsOriginal);
     }
 
-    public transformStringCallExpression(node: ts.CallExpression): tstl.Expression {
+    public transformStringCallExpression(node: ts.CallExpression, signature: ts.Signature): tstl.Expression {
```
Perryvw: Thinking about the API we want to make available for people to override the transformers, I have my doubts if adding this signature parameter is a good idea. We have a couple options here: Can we just get the signature from the node inside this function? Is this a function that should not be in the public interface? Or do we just keep it like this.

My main point being it is not really clear what this signature is or where you should get it from? Is it not just part of the call expression?

tomblind: That's a good point. I'm pretty sure we can grab it inside the function - I was only passing it in since we already had it.

PR #488, opened 2019-03-18
Incremental compilation in watch mode

TheLartians:
For large projects, this reduces compilation time dramatically in watch mode, as only files actually affected by recent changes are transpiled and emitted.

_Update: Now also processes transpiler errors in watch mode._

### DIFF
--- a/src/Compiler.ts
+++ b/src/Compiler.ts
@@ -56,7 +56,8 @@ export function watchWithOptions(fileNames: string[], options: CompilerOptions):
                 while (true) {
                     const currentFile = program.getSemanticDiagnosticsOfNextAffectedFile();
                     if (!currentFile) { break; }
-                    status = status || transpiler.emitSourceFile(currentFile.affected as ts.SourceFile);
+                    const fileStatus = transpiler.emitSourceFile(currentFile.affected as ts.SourceFile);
+                    status |= fileStatus;
                 }
             }
             // do a full recompile after transpiler error.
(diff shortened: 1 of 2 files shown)


src/LuaTranspiler.ts
```diff
@@ -119,11 +121,13 @@ export class LuaTranspiler {
                     // Graciously handle transpilation errors
                     console.error("Encountered error parsing file: " + exception.message);
                     console.error(`${sourceFile.fileName} (${1 + pos.line},${pos.character})\n${exception.stack}`);
+                    return 1;
```
Perryvw: Might be nice to make these constants or an enum.

TheLartians: I agree, though we should probably discuss error codes and general error handling of tstl in a separate PR.

src/Compiler.ts
```diff
@@ -43,10 +43,27 @@ export function watchWithOptions(fileNames: string[], options: CompilerOptions):
...
+        let status = transpiler.reportErrors();
+
+        if (status === 0) {
+            if (fullRecompile) {
+                status = transpiler.emitFilesAndReportErrors();
+            } else {
+                while (true) {
+                    const currentFile = program.getSemanticDiagnosticsOfNextAffectedFile();
+                    if (!currentFile) { break; }
+                    const fileStatus = transpiler.emitSourceFile(currentFile.affected as ts.SourceFile);
```
Perryvw: currentFile has type `ts.SourceFile | ts.Program`, you should explicitly handle both cases using `ts.isSourceFile(currentFile)`, and `program.getSourceFiles()` to transpile all sourceFiles in a program, in case one is returned.

PR #490, opened 2019-03-23
Added override for traceback

Perryvw:
Added override for `debug.traceback` to apply the source map to stack traces inside Lua.


src/LuaPrinter.ts
```diff
@@ -136,11 +147,30 @@ export class LuaPrinter {
...
 
-        this.sourceFile = path.basename(sourceFile);
+        this.sourceFile = path.basename(sourceFile, ".ts");
 
         const blockNode =  this.createSourceNode(block, this.printBlock(block));
 
-        return this.concatNodes(header, blockNode);
+        let sourceNode = this.concatNodes(header, blockNode);
+
+        if (this.options.sourceMapTraceBack) {
```
tomblind: Shouldn't this be before the main body? What if there's an error executing a global statement?

src/lualib/SourceMapTraceBack.ts
```diff
@@ -2,19 +2,31 @@ declare const debug: {
...
 
-function __TS__SourceMapTraceBack(fileName: string, sourceMap: {[line: number]: number}): void {
-    getfenv(1)["traceback"] = getfenv(1)["traceback"] || {};
-    getfenv(1)["traceback"][fileName] = getfenv(1)["traceback"][fileName] || debug.traceback;
-    debug.traceback = (...args: any[]) => {
-        let trace = getfenv(1)["traceback"][fileName](...args);
+// TODO: In the future, change this to __TS__RegisterFileInfo and provide tstl interface to
+// get some metadata about transpilation.
+function __TS__SourceMapTraceBack(this: void, fileName: string, sourceMap: {[line: number]: number}): void {
+    _G["__sourcemap"] = _G["__sourcemap"] || {};
```
tomblind: I believe we usually avoid using double-underscore as that's reserved for lua.

src/lualib/string.d.ts
```diff
@@ -1,9 +1,11 @@
 /** @luaIterator */
-interface GMatchResult extends Iterable<string> { }
+interface GMatchResult extends Array<string> { }
 
+/** @noSelf */
 declare namespace string {
     /** @tupleReturn */
     function gsub(source: string, searchValue: string, replaceValue: string): [string, number];
+    function gsub(source: string, searchValue: string, replaceValue: (...groups: string[]) => string): [string, number];
```
tomblind: this will break when my tupleReturn PR is merged - you'll need to specify `@tupleReturn` for each overload

src/LuaPrinter.ts
Perryvw: That is a good point, the issue with that is that it modifies the source map input this function uses, but since it's only one line we can probably deal with this with a + or -1 somewhere, I'll investigate.

tomblind: That's a good point - we'll need to account for the header comment and inlined lua libs as well, won't we?

src/lualib/SourceMapTraceBack.ts
```diff
@@ -2,19 +2,33 @@ declare const debug: {
...
-        return trace;
-    };
+    if (_G.__originalTraceback === undefined) {
+        _G.__originalTraceback = debug.traceback;
+        debug.traceback = (thread, message, level) => {
+            const trace = _G["__TS__originalTraceback"](thread, message, level);
+
+            const [result, occurrences] = string.gsub(
+                trace,
+                "([^\\]+).lua:(%d+)",
```
tomblind: I think the pattern needs to include `/` for systems that use that in path separation.

src/LuaPrinter.ts
```diff
@@ -134,11 +136,15 @@ export class LuaPrinter {
             }
         }
 
-        this.sourceFile = path.basename(sourceFile);
+        this.sourceFile = path.basename(sourceFile, ".ts");
```
ark120202: Won't it fail if there are files with the same name in different directories?
It might be possible to use `debug.getinfo(2).short_src` inside `__TS__SourceMapTraceBack` to make it work without using some fixed path

src/lualib/string.d.ts
Perryvw: Fixed this, diff doesn't seem to pick it up

PR #491, opened 2019-03-24
Migrate tests to Jest

ark120202:
Notable changes:

* Switched testing framework to Jest
* Heavily improved performance by improving caching and excluding unnecessary types
* Enabled tslint on tests
* Enabled prettier on tests (since they were reformatted anyway)
* Fixed [options.spec.ts](https://github.com/TypeScriptToLua/TypeScriptToLua/blob/512e7d0aeb4c8969f5cb81241aafe1666ac4cd9c/test/unit/compiler/configuration/options.spec.ts) not testing anything


__tests__/unit/lualib/set.test.ts
```diff
@@ -0,0 +1,130 @@
...
+    expect(result).toBe(18);
+});
+
+test("set size", () => {
+    expect(util.transpileAndExecute(`let m = new Set(); return m.size;`)).toBe(0);
+    expect(util.transpileAndExecute(`let m = new Set(); m.add(1); return m.size;`)).toBe(1);
+    expect(util.transpileAndExecute(`let m = new Set([1, 2]); return m.size;`)).toBe(2);
+    expect(util.transpileAndExecute(`let m = new Set([1, 2]); m.clear(); return m.size;`)).toBe(0);
+    expect(util.transpileAndExecute(`let m = new Set([1, 2]); m.delete(2); return m.size;`)).toBe(
+        1,
```
Perryvw: This looks strange

src/TranspileError.ts
```diff
@@ -1,9 +1,8 @@
 import * as ts from "typescript";
 
 export class TranspileError extends Error {
-    public node: ts.Node;
-    constructor(message: string, node: ts.Node) {
+    public name = 'TranspileError';
+    constructor(message: string, public node?: ts.Node) {
```
Perryvw: Don't know if we should make `node` optional.

__tests__/unit/loops.test.ts
```diff
@@ -0,0 +1,789 @@
...
+                j--
+                continue;
+            }
+            arrTest[i] = j;
+            j--;
+        } while (j > 0)
+
+        i++;
+    } while (i < arrTest.length)
+    return JSONStringify(arrTest);`,
```
Perryvw: Indenting is messed up here

__tests__/compiler/errorreport.test.ts
```diff
@@ -0,0 +1,24 @@
...
+import { compileFilesWithOptions } from "../../src/Compiler";
+
+test.each([
+    {
+        errorMsg:
+            "Encountered error parsing file: Default Imports are not supported, please use named imports instead!",
+        fileName: "default_import.ts",
+    },
+])("Compile project (%p)", ({ errorMsg, fileName }) => {
+    jest.spyOn(console, "log").mockReturnValue(undefined);
```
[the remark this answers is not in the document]
ark120202: Yes, without it `jest.spyOn` would just track calls, but not stub default implementation. It was the same with alsatian - `SpyOn(...).andStub()` (but `andStub` didn't require correct return type)

__tests__/compiler/watcher_proccess.js
```diff
@@ -0,0 +1,6 @@
+require("ts-node/register/transpile-only");
```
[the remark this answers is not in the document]
ark120202: Before all tests were compiled to .js as a separate step, so child process could run `watcher_proccess.js`. Now all tests are compiled in-memory with `ts-jest` (for better watching DX and potential for more optimization in the future), but that doesn't work with child processes. Though maybe it would be better to load `ts-node/register/transpile-only` as a part of fork call.

src/TranspileError.ts
ark120202: Makes sense, but it would be invalid once we'll enable `strictNullChecks`, because tests have to get error messages from TSTLErrors. The only type-safe way I see is making TSTLErrors just do string formatting. It also would make stack traces a bit neater.

```ts
export const formatCouldNotFindEnumMember = (enumDeclaration: ts.EnumDeclaration, enumMember: string) => `Could not find ${enumMember} in ${enumDeclaration.name.text}`
```

__tests__/unit/loops.test.ts
ark120202: I have fixed it in this file, but looks like there are more similar cases. I'll look into solving it with a codemod

ark120202:
> Why did you rename test to \_\_tests\_\_?

It's more common and consistent with other jest directories like \_\_snapshots\_\_ and \_\_mocks\_\_. Though probably I should do it in different commits to preserve history, but that would need a manual rebase  and not using squash & merge. Maybe it would be better to leave it in test/**/*.spec.ts, I don't have a preference on it.

test/translation/ts/characterEscapeSequence.ts
```diff
@@ -2,15 +2,15 @@ let quoteInDoubleQuotes = "' ' '";
 let quoteInTemplateString = `' ' '`;
 
 let doubleQuoteInQuotes = '" " "';
-let doubleQuoteInDoubleQuotes = "\" \" \"";
+let doubleQuoteInDoubleQuotes = '" " "';
```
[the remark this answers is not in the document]
Perryvw: this is indeed a duplicate of line 4

PR #496, opened 2019-03-27
Add array.flat and array.flatMap

ark120202:


src/lualib/ArrayFlat.ts
```diff
@@ -0,0 +1,12 @@
+function __TS__ArrayFlat(this: void, array: any[], depth = 1): any[] {
```
Perryvw: Does this break if you type it `function __TS__ArrayFlat<T>(this: void, array: Array<T[]>, depth = 1): T[] {`? That seems more correct to me

test/unit/lualib/lualib.spec.ts
```diff
@@ -377,6 +377,39 @@ test.each([
...
+    { array: [1, [[2], [3]], 4], expected: [1, [2], [3], 4] },
+    { array: [1, [[[2], [3]]], 4], depth: Infinity, expected: [1, 2, 3, 4] },
+])("array.flat (%p)", ({ array, depth, expected }) => {
+    // TODO: Remove once `Infinity` would be implemented
+    const luaDepth = depth === Infinity ? "1 / 0" : depth;
+    const result = util.transpileAndExecute(`
+        return JSONStringify(${JSON.stringify(array)}.flat(${luaDepth}))
+    `);
+
+    expect(JSON.parse(result)).toEqual(expected);
```
Perryvw: Might be nicer to compare this to `array.flat()` instead of `expected`

test/unit/lualib/lualib.spec.ts
```diff
@@ -377,6 +377,39 @@ test.each([
...
+    { array: [1, 2, 3], map: (v: number, i: number) => [v * 2 * i] },
+])("array.flatMap (%p)", ({ array, map }) => {
+    const result = util.transpileAndExecute(`
+        const array = ${JSON.stringify(array)};
+        const result = array.flatMap(${map.toString()});
+        return JSONStringify(result);
+    `);
+
+    // TODO(node 12): array.flatMap(map)
+    const expected = [].concat(...(array as any[]).map(map));
```
Perryvw: Why not use actual flatmap here as expected?

test/unit/lualib/lualib.spec.ts
ark120202: https://node.green/#ES2019-features-Array-prototype--flat--flatMap-

src/lualib/ArrayFlat.ts
ark120202: It's not correct because of the depth. Without some complex types it's impossible to correctly type it (TS now supports only depth up to 4, and falls back to `any[]` - https://github.com/Microsoft/TypeScript/blob/f7c4fefeb62416c311077a699cc15beb211c25c9/src/lib/es2019.array.d.ts#L112)

PR #500, opened 2019-03-29
Improve Prettier DX

ark120202:
* Added a message on Prettier check failure with instructions that show how to fix it. Tested in cmd, powershell, and (ba)sh.
* Disabled `endOfLine` locally, but left it enabled on CI. That way it doesn't cause initial warnings with `core.autocrlf=true` (default on Windows), yet prevents committing files with inconsistent line endings to the repo.

### DIFF
--- a/.prettierrc.js
+++ b/.prettierrc.js
@@ -1,0 +1,11 @@
+const isCI = require("is-ci");
+
+/** @type {import("prettier").Options} */
+module.exports = {
+    printWidth: 100,
+    tabWidth: 4,
+    trailingComma: "all",
+    proseWrap: "always",
+    endOfLine: isCI ? "lf" : "auto",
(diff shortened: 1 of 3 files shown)


.prettierrc.js
```diff
@@ -0,0 +1,11 @@
+const isCI = require("is-ci");
+
+/** @type {import("prettier").Options} */
+module.exports = {
+    printWidth: 100,
```
Perryvw: Why is this 100 actually? I think it's 120 in tslint

ark120202: In linters max length is used as a strict limit, but it's possible to make lines shorter. In prettier `printWidth` it more like average length, as prettier tries to as fit much as possible code into `printWidth`, so usually it's better to keep it less than it was with linter. See https://prettier.io/docs/en/options.html#print-width and https://prettier.io/docs/en/rationale.html#print-width for details.
Though I'm fine either way, the difference between 100 and 120 is quite small.

PR #501, opened 2019-03-30
output cleanup

tomblind:
Removed wrapping all binary & unary expressions in parenthesis and fixed issues that arose. Also removed empty string literals from template strings.

I'd like to test this against some real code bases and see how fragile it is.

### DIFF
--- a/src/LuaPrinter.ts
+++ b/src/LuaPrinter.ts
@@ -376,24 +376,17 @@ export class LuaPrinter {
         return this.createSourceNode(statement, [this.indent("::"), statement.name, "::\n"]);
     }
 
-    private printReturnStatement(statement: tstl.ReturnStatement, inline?: boolean): SourceNode {
+    private printReturnStatement(statement: tstl.ReturnStatement): SourceNode {
         if (!statement.expressions || statement.expressions.length === 0) {
-            const ret = inline ? "return;" : this.indent("return;\n");
-            return this.createSourceNode(statement, ret);
+            return this.createSourceNode(statement, this.indent("return;\n"));
         }
 
         const chunks: SourceChunk[] = [];
 
-        chunks.push("return ");
         chunks.push(...this.joinChunks(", ", statement.expressions.map(e => this.printExpression(e))));
-        chunks.push(";");
+        chunks.push(";\n");
 
-        if (!inline) {
-            chunks.unshift(this.indent());
-            chunks.push("\n");
-        }
-
-        return this.createSourceNode(statement, chunks);
+        return this.createSourceNode(statement, [this.indent(), "return ", ...chunks]);
     }
 
     private printBreakStatement(statement: tstl.BreakStatement): SourceNode {
@@ -467,7 +460,7 @@ export class LuaPrinter {
         }
     }
 
-    private printFunctionExpression(expression: tstl.FunctionExpression): SourceNode {
+    private printFunctionParameters(expression: tstl.FunctionExpression): SourceChunk[] {
         const parameterChunks: SourceNode[] = expression.params
             ? expression.params.map(i => this.printIdentifier(i))
             : [];
@@ -475,50 +468,45 @@ export class LuaPrinter {
         if (expression.dots) {
             parameterChunks.push(this.printDotsLiteral(expression.dots));
         }
+        return ["(", ...this.joinChunks(", ", parameterChunks), ")"];
+    }
 
+    private printFunctionExpression(expression: tstl.FunctionExpression): SourceNode {
         const chunks: SourceChunk[] = [];
 
-        chunks.push("function(");
-        chunks.push(...this.joinChunks(", ", parameterChunks));
-        chunks.push(")");
+        chunks.push("function");
+        chunks.push(...this.printFunctionParameters(expression));
 
-        if (expression.body.statements && expression.body.statements.length === 1) {
-            const statement = expression.body.statements[0];
-            if (tstl.isReturnStatement(statement) && (expression.flags & tstl.FunctionExpressionFlags.Inline) !== 0) {
-                // Inline return-only functions with the flag
-                chunks.push(" ");
-                chunks.push(this.printReturnStatement(statement, true));
-                chunks.push(" end");
-                return this.createSourceNode(expression, chunks);
-            }
+        if (tstl.isInlineFunctionExpression(expression)) {
+            const returnStatement = expression.body.statements[0];
+            chunks.push(" ");
+            const returnNode: SourceChunk[] = [
+                "return ",
+                ...this.joinChunks(", ", returnStatement.expressions.map(e => this.printExpression(e))),
+                ";",
+            ];
+            chunks.push(this.createSourceNode(returnStatement, returnNode));
+            chunks.push(" end");
+
+        } else {
+            chunks.push("\n");
+            this.pushIndent();
+            chunks.push(this.printBlock(expression.body));
+            this.popIndent();
+            chunks.push(this.indent("end"));
         }
-
-        chunks.push("\n");
-        this.pushIndent();
-        chunks.push(this.printBlock(expression.body));
-        this.popIndent();
-        chunks.push(this.indent("end"));
 
         return this.createSourceNode(expression, chunks);
     }
 
     private printFunctionDefinition(statement: tstl.FunctionDefinition): SourceNode {
         const expression = statement.right[0];
-        const parameterChunks: SourceNode[] = expression.params
-            ? expression.params.map(i => this.printIdentifier(i))
-            : [];
-
-        if (expression.dots) {
-            parameterChunks.push(this.printDotsLiteral(expression.dots));
-        }
-
         const chunks: SourceChunk[] = [];
 
         chunks.push("function ");
         chunks.push(this.printExpression(statement.left[0]));
-        chunks.push("(");
-        chunks.push(...this.joinChunks(", ", parameterChunks));
-        chunks.push(")\n");
+        chunks.push(...this.printFunctionParameters(expression));
+        chunks.push("\n");
 
         this.pushIndent();
         chunks.push(this.printBlock(expression.body));
--- a/src/LuaPrinter.ts
+++ b/src/LuaPrinter.ts
@@ -3,6 +3,19 @@ import * as tstl from "./LuaAST";
 import { LuaLib, LuaLibFeature } from "./LuaLib";
 import { TSHelper as tsHelper } from "./TSHelper";
 
+type FunctionDefinition = (tstl.VariableDeclarationStatement | tstl.AssignmentStatement) & {
+    right: [tstl.FunctionExpression];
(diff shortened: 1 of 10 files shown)


src/LuaTransformer.ts
```diff
@@ -3072,6 +3078,10 @@ export class LuaTransformer {
     }
 
     public transformParenthesizedExpression(expression: ts.ParenthesizedExpression): tstl.Expression {
+        if (ts.isAssertionExpression(expression.expression)) {
```
Perryvw: What about `ts.isAsExpression`

src/LuaTransformer.ts
```diff
@@ -3567,20 +3579,24 @@ export class LuaTransformer {
                     node.arguments.length === 1
                         ? this.createStringCall("find", node, caller, params[0])
                         : this.createStringCall(
-                              "find", node, caller, params[0],
-                              this.expressionPlusOne(params[1]),
-                              tstl.createBooleanLiteral(true)
+                                "find", node, caller, params[0],
```
Perryvw: Looks like this should have shifted 2 spaces left instead of right

test/unit/expressions.spec.ts
```diff
@@ -177,10 +177,10 @@ test.each(["a>>b", "a>>=b"])("Unsupported bitop 5.3 (%p)", input => {
 
 test.each([
     { input: "1+1", lua: "1 + 1;" },
-    { input: "-1+1", lua: "(-1) + 1;" },
-    { input: "1*30+4", lua: "(1 * 30) + 4;" },
```
Perryvw: Hmmm I'm not too sure about this removal, can we be sure operator prescedence is always the same in TS as it is in lua? What about adding a testcase with input `10 - (4 + 5)`.

src/LuaTransformer.ts
tomblind: `AssertionExpression` is defined as `type AssertionExpression = TypeAssertion | AsExpression;` in TS

tomblind:
Added to this PR:
-functions are now output as `function foo()` instead of `foo = function()`, except for methods which were assigned as function expressions in the original TS
-arrow functions with no body are inlined
-arrow functions with no parameters don't include the dummy parameter `____`, since it's not necessary
-empty `do...end` statements are not generated
-continue labels are only output if the loop actually has a continue statement

src/LuaTransformer.ts
```diff
@@ -3589,7 +3620,11 @@ export class LuaTransformer {
                 } else {
                     const arg1 = params[0];
                     const arg2 = params[1];
-                    const sumArg = tstl.createBinaryExpression(arg1, arg2, tstl.SyntaxKind.AdditionOperator);
+                    const sumArg = tstl.createBinaryExpression(
+                        tstl.createParenthesizedExpression(arg1),
```
Perryvw: This seems like a random parenthesization?

src/LuaTransformer.ts
```diff
@@ -3994,24 +4031,29 @@ export class LuaTransformer {
...
+        if (parts.length === 1) {
+            return parts[0];
+        }
         return parts.reduce((prev, current) => tstl.createBinaryExpression(
             prev,
             current,
             tstl.SyntaxKind.ConcatOperator)
-        ) as tstl.BinaryExpression;
+        );
     }
```
Perryvw: Some whitelines wouldn't hurt in the above function.

src/LuaTransformer.ts
```diff
@@ -3994,24 +4031,29 @@ export class LuaTransformer {
...
-                tstl.createCallExpression(tstl.createIdentifier("tostring"), [expr]),
-                text,
-                tstl.SyntaxKind.ConcatOperator)
-            );
+            const text = tsHelper.escapeString(span.literal.text);
+            if (text.length > 0) {
+                parts.push(tstl.createStringLiteral(text, span.literal));
+            }
         });
+        if (parts.length === 1) {
```
Perryvw: Is this `if` statement needed? Doesn't the following reduce behave the same?

src/LuaTransformer.ts
```diff
@@ -3003,18 +3012,28 @@ export class LuaTransformer {
...
+        }
+
         // Build parameter string
-        const [paramNames, dotsLiteral, spreadIdentifier] = this.transformParameters(
-            node.parameters,
-            hasContext ? context : undefined
-        );
+        const [paramNames, dotsLiteral, spreadIdentifier] = this.transformParameters(node.parameters, context);
+
+        let flags = !ts.isFunctionDeclaration(node)
```
Perryvw: This statement looks very strange, when is a functionExpression not a functionDeclaration/

src/LuaPrinter.ts
```diff
@@ -3,6 +3,19 @@ import * as tstl from "./LuaAST";
...
+
+function isFunctionDeclaration(statement: tstl.VariableDeclarationStatement | tstl.AssignmentStatement)
+    : statement is FunctionDefinition
+{
+    return statement.left.length === 1
+        && statement.right
+        && statement.right.length === 1
+        && tstl.isFunctionExpression(statement.right[0]);
+}
+
```
Perryvw: This stuff looks out of place 🤔

src/LuaTransformer.ts
tomblind: Perhaps this flag can be named better - it's a hint for the printer to output as an assignment expression `foo = function()` instead of a function declaration `function foo()`

Perryvw: It's not the flag that I'm confused by, but the condition here.

src/LuaTransformer.ts
tomblind: if the arg is a complex expression (such as another binary), it needs parenthesis to avoid order-of-operation issues. Now that I re-look at this, the second one should probably be parenthesized as well.

tomblind:
Last addition to this PR: tables with more than one entry are now formatted into multiple lines.

src/LuaTransformer.ts
```diff
@@ -4557,10 +4621,10 @@ export class LuaTransformer {
             if (symbols.some(s => this.shouldHoist(s, scope))) {
                 let assignment: tstl.AssignmentStatement | undefined;
                 if (declaration.right) {
-                    assignment = tstl.createAssignmentStatement(
-                        declaration.left,
-                        declaration.right
-                    );
+                    assignment = tstl.createAssignmentStatement(declaration.left, declaration.right);
+                    // Preserve position info for sourcemap
```
Perryvw: Probably cleaner if you use `tstl.setNodeOriginal` here, instead of manually setting line and column.

src/TSHelper.ts
```diff
@@ -652,6 +652,11 @@ export class TSHelper {
         return match && match[0] === str;
     }
 
+    public static isValidLuaFunctionDeclarationName(str: string): boolean {
```
Perryvw: What is the difference between this and `isValidLuaIdentifier`?

test/unit/loops.spec.ts
```diff
@@ -763,21 +763,23 @@ test("forof forwarded lua iterator with tupleReturn", () => {
...
-    "do { i++; } while (a < b)",
-    "for (let i = 0; i < 3; i++) {}",
-    "for (let a in b) {}",
-    "for (let a of b) {}",
+    "while (a < b) { i++; continue; }",
+    "do { i++; continue; } while (a < b)",
+    "for (let i = 0; i < 3; i++) { continue; }",
+    "for (let a in b) { continue; }",
+    "for (let a of b) { continue; }",
 ])("loop versions (%p)", loop => {
```
Perryvw: Looking at what is being asserted here, this should probably have a better name, something like "loop continue for each lua target". Would probably also be nicer to have each version have its own test, but we can split that up in a later PR.

src/LuaPrinter.ts
```diff
@@ -447,14 +479,34 @@ export class LuaPrinter {
...
+        chunks.push(")");
+
+        if (expression.body.statements
+            && expression.body.statements.length === 1
+            && tstl.isReturnStatement(expression.body.statements[0])
+            && (expression.flags & tstl.FunctionExpressionFlags.Inline) !== 0)
+        {
+            // Inline return-only functions with the flag
+            chunks.push(" ");
+            chunks.push(this.printReturnStatement(expression.body.statements[0] as tstl.ReturnStatement, true));
```
Perryvw: Is this `as tstl.ReturnStatement` required? Or does the typeguard in the `if` condition not work on index expressions?

src/LuaPrinter.ts
```diff
@@ -436,7 +468,7 @@ export class LuaPrinter {
         }
     }
 
-    private printFunctionExpression(expression: tstl.FunctionExpression): SourceNode {
+    private printFunctionExpression(expression: tstl.FunctionExpression, name?: SourceChunk): SourceNode {
```
Perryvw: I think for plugin/overriding printer reasons it is best not to pollute our interface with random function arguments (similar to what we discussed recently in the transformer visitor pattern), surely there is a better way to do this right?

src/LuaTransformer.ts
tomblind: The original TS node is gone at this point. I'd have to restructure things to preserve that.

src/TSHelper.ts
tomblind: This allows a period `.`

src/LuaPrinter.ts
tomblind: Yeah, it's required - the typeguard doesn't work with array elements

src/LuaPrinter.ts
tomblind: I'll move this into it's own `printFunctionDeclaration` method

tomblind:
Updated with fixes for feedback

src/TSHelper.ts
Perryvw: This might be a little nitpicky, but would it make sense/ be worth it to implement this function as `str.split(".").every(s => s.length > 0 && isValidLuaIdentifier(s))`?

tomblind: To be honest, I'm not sure that's any clearer. Perhaps just a comment explaining the difference would be suffice?

Perryvw: That works for me.

src/LuaPrinter.ts
```diff
@@ -351,18 +376,24 @@ export class LuaPrinter {
         return this.createSourceNode(statement, [this.indent("::"), statement.name, "::\n"]);
     }
 
-    private printReturnStatement(statement: tstl.ReturnStatement): SourceNode {
+    private printReturnStatement(statement: tstl.ReturnStatement, inline?: boolean): SourceNode {
```
Perryvw: What about this inline argument, would it be possible to remove this from the interface?

tomblind: Yeah - I'll refactor some stuff around to make that work

PR #505, opened 2019-04-04
removing semi-colons, except where needed for disambiguation

tomblind:


src/LuaPrinter.ts
```diff
@@ -126,7 +126,7 @@ export class LuaPrinter {
             if ((this.options.luaLibImport === LuaLibImportKind.Require && luaLibFeatures.size > 0)
                 || this.options.luaLibImport === LuaLibImportKind.Always)
             {
-                header += `require("lualib_bundle");\n`;
+                header += `require("lualib_bundle")\n`;
```
Perryvw: What if the first line after this starts with a parenthesis?

tomblind: Nice catch!

TheLartians:
Looks good and runs fine!

PR #506, opened 2019-04-04
ensuring statement expressions are transpiled to valid lua

### LINKED ISSUE #503: Expression as statements produce invalid lua
```ts
"foobar";
1;

declare function foo(): number;
foo() + foo();
```
TS allows arbitrary expressions to be used as statements, but this is invalid in Lua. Lua only allows call expressions to be used as statements.

tomblind:
fixes #503 

This will conflict with #505, so I guess I'll see which gets approved first and fix the other 😄

### DIFF
--- a/src/LuaTransformer.ts
+++ b/src/LuaTransformer.ts
@@ -1817,6 +1817,14 @@ export class LuaTransformer {
             );
         }
 
+        if (!ts.isCallLikeExpression(expression)) {
+            // Assign expression statements to dummy to make sure they're legal lua
+            return tstl.createVariableDeclarationStatement(
+                tstl.createAnnonymousIdentifier(),
+                this.transformExpression(expression)
+            );
+        }
+
         return tstl.createExpressionStatement(this.transformExpression(expression));
     }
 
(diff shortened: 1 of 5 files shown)


lolleko:
Wouldnt it be better to just ommit those? they dont have any effect right? (except calls)

not sure if we should waste resources assigning to `local ____`

ark120202:
@lolleko Expressions may have side effects (e.g. calls) deeply nested inside them, for example a quite common pattern: `a ? foo() : bar()`. Of course it's possible to traverse it and only emit when there are possible side effects, but IMO it might not worth it.

lolleko:
Okay! People probably don't add useless statements like `"hello"; !myBool;`... to their code anyway.

Perryvw:
I guess we can choose:
* fix it like this
* omit them silently (probably not a good idea due to what @ark120202 mentioned)
* throw an error when this occurs.

I think out of these the current solu

tomblind:
Simple literals are pretty much the only thing guaranteed to not have side effects. Even something like
```ts
-a;
```
could invoke metamethods. Since having simple literals as statements isn't something people are likely to do, it doesn't really seem worth the extra logic.

PR #508, opened 2019-04-04
String concat fix

### LINKED ISSUE #504: String concat needs to convert all values to strings
```ts
print("foo" + 2 + true);
```
=>
```lua
print("foo" .. 2 .. true); --attempt to concatenate a boolean value
```

The string concat operator will only convert numbers to strings. Other types will cause an error (unlike TS which calls toString on all values);

tomblind:
fixes #504 

- wraps values that aren't guaranteed to be safe in a concat op with `tostring`
  - for now, safe values are string/number literals and other concat ops
- decision to wrap is also applied to template strings, which cleans them up a bit
- tests now include concatonating boolean values, which don't automatically coerce into strings
- a bit unrelated, but I also added the inline flag to `wrapInFunctionCall`, which is used for ternaries

### DIFF
--- a/src/LuaTransformer.ts
+++ b/src/LuaTransformer.ts
@@ -6,7 +6,6 @@ import * as tstl from "./LuaAST";
 import { LuaLibFeature } from "./LuaLib";
 import { ContextType, TSHelper as tsHelper } from "./TSHelper";
 import { TSTLErrors } from "./TSTLErrors";
-import { expression } from "@babel/template";
 
 export type StatementVisitResult = tstl.Statement | tstl.Statement[] | undefined;
 export type ExpressionVisitResult = tstl.Expression | undefined;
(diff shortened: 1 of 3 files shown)


src/LuaTransformer.ts
```diff
@@ -6,6 +6,7 @@ import * as tstl from "./LuaAST";
 import { LuaLibFeature } from "./LuaLib";
 import { ContextType, TSHelper as tsHelper } from "./TSHelper";
 import { TSTLErrors } from "./TSTLErrors";
+import { expression } from "@babel/template";
```
Perryvw: Accidental import?

test/unit/string.spec.ts
```diff
@@ -34,14 +34,41 @@ test.each([
...
+
+test.each([
+    { a: 12, b: 23, c: 43 },
+    { a: "test", b: "hello", c: "bye" },
+    { a: "test", b: 42, c: "bye" },
+    { a: "test", b: 42, c: 12 },
+    { a: "test", b: 42, c: true },
+    { a: false, b: 42, c: 12 },
+])("String Concat Operator (%p)", ({ a, b, c }) => {
+    const a1 = typeof a === "string" ? "'" + a + "'" : a;
```
Perryvw: Probaably nicer if you do `` `'${a}'` ``

ark120202: It's also possible to just `JSON.stringify(a)`

PR #512, opened 2019-04-07
Removing "or" assignment from exports, class and namespace declarations

### LINKED ISSUE #509: How to make `local exports = exports or {};` into `local exports = {};` ?
How to make `local exports = exports or {};` into `local exports = {};` ?

tomblind:
fixes #509 

- exports, class, and namespaces are now initialized as `X = {}` instead of `X = X or {}`.
- `@noClassOr` has been removed as it no longer does anything
- namespaces are only initialized if they are not being merged into a previous declaration
- nested namespaces are now only exported if they have the `export` keyword
- exported nested namespaces no longer generate a local reference in the parent if it's not needed

### DIFF
--- a/src/Decorator.ts
+++ b/src/Decorator.ts
@@ -21,8 +21,6 @@ export class Decorator {
                 return DecoratorKind.Phantom;
             case "tuplereturn":
                 return DecoratorKind.TupleReturn;
-            case "noclassor":
-                return DecoratorKind.NoClassOr;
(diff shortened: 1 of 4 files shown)


tomblind:
I've added 2 things to this PR:
- `exports` is now `____exports`
- exported classes are now cached in a local so they don't need to be references as `____exports.MyClass` everywhere

PR #515, opened 2019-04-09
Enabled strict config

Perryvw:
WIP: Still need all strict issues to be solved.

Also made some preliminary interface changes to the transformer (although that should probably be finalized in a separate PR).


src/LuaTransformer.ts
```diff
@@ -758,7 +771,7 @@ export class LuaTransformer {
         const hasStaticSetters = tsHelper.hasSetAccessorInClassOrAncestor(statement, true, this.checker);
 
         if (extendsType) {
-            const extendedTypeNode = tsHelper.getExtendedTypeNode(statement, this.checker);
+            const extendedTypeNode = tsHelper.getExtendedTypeNode(statement, this.checker)!;
```
tomblind: Should we check this result and throw if it's undefined?

src/TSHelper.ts
```diff
@@ -314,7 +316,7 @@ export class TSHelper {
         return undefined;
     }
 
-    public static isBinaryAssignmentToken(token: ts.SyntaxKind): [boolean, ts.BinaryOperator] {
+    public static isBinaryAssignmentToken(token: ts.SyntaxKind): [true, ts.BinaryOperator] | [false, never] {
```
tomblind: TS simplifies `ts.BinaryOperator | never` to `ts.BinaryOperator`, which isn't exactly what's desired here.

src/TSHelper.ts
```diff
@@ -371,7 +373,7 @@ export class TSHelper {
         node: ts.Expression,
         checker: ts.TypeChecker,
         program: ts.Program
-    ): [boolean, ts.Expression, ts.Expression]
+    ): [true, ts.Expression, ts.Expression] | [false, never, never]
```
tomblind: Same as `isBinaryAssignmentToken`

src/TSHelper.ts
```diff
@@ -720,7 +738,10 @@ export class TSHelper {
         return this.isStandardLibraryDeclaration(declaration, program);
     }
 
-    public static isEnumMember(enumDeclaration: ts.EnumDeclaration, value: ts.Expression): [boolean, ts.PropertyName] {
+    public static isEnumMember(
+        enumDeclaration: ts.EnumDeclaration,
+        value: ts.Expression
+    ): [true, ts.PropertyName] | [false, never] {
```
tomblind: Same as `isBinaryAssignmentToken`

src/LuaTransformer.ts
```diff
@@ -936,10 +971,14 @@ export class LuaTransformer {
             m => tsHelper.isGetAccessorOverride(m, classDeclarataion, this.checker)
         );
         for (const getter of getOverrides) {
+            const getterName = getter.name !== undefined
+                ? this.expectExpression(this.transformPropertyName(getter.name))
+                : tstl.createAnnonymousIdentifier();
```
tomblind: This isn't right - `getter.name` can't be undefined because the `getter` is actually a `ts.GetAccessorDeclaration` (and if it could, using an anon here wouldn't be the right thing to do). I recommend casting `getOverrides` above:
```ts
const getOverrides = classDeclarataion.members.filter(
    m => tsHelper.isGetAccessorOverride(m, classDeclarataion, this.checker)
) as ts.GetAccessorDeclaration[];
```

src/LuaTransformer.ts
```diff
@@ -1631,7 +1686,7 @@ export class LuaTransformer {
             : undefined;
         const [params, dotsLiteral, restParamName] = this.transformParameters(functionDeclaration.parameters, context);
 
-        const name = this.transformIdentifier(functionDeclaration.name);
+        const name = this.transformIdentifier(functionDeclaration.name!);
```
tomblind: We should probably throw an error if `name` is undefined.

src/LuaTransformer.ts
```diff
@@ -2229,6 +2312,11 @@ export class LuaTransformer {
 
     public transformBreakStatement(breakStatement: ts.BreakStatement): StatementVisitResult {
         const breakableScope = this.findScope(ScopeType.Loop | ScopeType.Switch);
+
+        if (breakableScope === undefined) {
+            throw new Error("Tried to pop non-existing break scope!");
```
tomblind: Needs explicit error type?

src/LuaTransformer.ts
```diff
@@ -3123,12 +3255,13 @@ export class LuaTransformer {
...
             const customDecorator = classDecorators.get(DecoratorKind.CustomConstructor);
-            if (!customDecorator.args[0]) {
+            if (customDecorator === undefined || customDecorator.args[0] === undefined) {
                 throw TSTLErrors.InvalidDecoratorArgumentNumber("@customConstructor", 0, 1, node);
             }
+
             return tstl.createCallExpression(
                 tstl.createIdentifier(customDecorator.args[0]),
-                this.transformArguments(node.arguments),
+                this.transformArguments(node.arguments || ([] as ts.Expression[])),
```
tomblind: Is this `as` necessary? usually TS would infer that sort of thing.

Perryvw: Oh I guess I must have done something wrongbefore that it didn't work, you are right it can resolve this.

jest.config.js
```diff
@@ -12,7 +10,7 @@ module.exports = {
     globals: {
         "ts-jest": {
             tsConfig: "<rootDir>/test/tsconfig.json",
-            diagnostics: { warnOnly: !isCI },
+            diagnostics: { warnOnly: false },
```
ark120202: Have you tried setting `"strict": true` in `test/tsconfig.json` instead?

src/TSTLErrors.ts
```diff
@@ -74,9 +95,12 @@ export class TSTLErrors {
     public static UnsupportedProperty = (parentName: string, property: string, node: ts.Node) =>
         new TranspileError(`Unsupported property on ${parentName}: ${property}`, node);
 
-    public static UnsupportedForTarget = (functionality: string, version: string, node: ts.Node) =>
+    public static UnsupportedForTarget = (functionality: string, version: string | undefined, node: ts.Node) =>
```
ark120202: `Lua undefined` doesn't sound right, won't it be better to handle it in the caller? It won't complain about nullable value if it would be `this.luaTarget = this.options.luaTarget || LuaTarget.LuaJIT` in the LuaTransformer constructor. That also would avoid mutating passed options object.

src/LuaTransformer.ts
```diff
@@ -4547,11 +4748,12 @@ export class LuaTransformer {
                     throw TSTLErrors.ReferencedBeforeDeclaration(identifier);
                 }
 
-            } else {
+            } else if (symbolId !== undefined) {
+                const sid = symbolId; // Capture non-undefined variable
```
ark120202: It might be simpler to replace `forEach` loop with a `for..of` instead

src/TSTLErrors.ts
Perryvw: Good idea

Issue #346, opened 2019-01-18
Custom lib that transpiles to a simple table


tomblind:
It would be nice to have a special type that can be used like JS's Map (allowing keys of any type), but that transpiles down to a simple lua table.

```ts
declare class Table<K extends {} = {}, V = any> {
    readonly length: number;
    set(key: K, value: V): void;
    get(key: K): V;
}

const tbl = new Table(); // local tbl = {}

const foo = {};
tbl.set(foo, "bar"); // tbl[foo] = "bar"
print(tbl.get(foo)); // print(tbl[foo])

tbl.set(1, "baz"); // tbl[1] = "baz"
print(tbl.length); // print(#tbl)
```

PR #516, opened 2019-04-10
Lua table

hazzard993:
For #346

`@LuaTable` can be used to mark a Map-like object to transpile down to a simple Lua table.

```ts
/** @LuaTable */
declare class Table<K extends {} = {}, V = any> {
    readonly length: number;
    set(key: K, value: V): void;
    get(key: K): V;
}

const tbl = new Table(); // local tbl = {}

const foo = {};
tbl.set(foo, "bar"); // tbl[foo] = "bar"
print(tbl.get(foo)); // print(tbl[foo])

tbl.set(1, "baz"); // tbl[1] = "baz"
print(tbl.length); // print(#tbl)
```

`get`, `set` and `length` are the only properties the transpiler pays special attention to. Other properties are treated the same as any other property.

I did make a decision to disallow `tbl.set(...)` to be used in place of an expression since it never returns a value and transpiles to an ExpressionStatement, not an Expression.

This could help with reducing the amount of Lua outputted through its use in LuaLibs and other source code.


src/LuaTransformer.ts
```diff
@@ -3569,6 +3597,16 @@ export class LuaTransformer {
         }
     }
 
+    public transformLuaTableProperty(node: ts.PropertyAccessExpression): tstl.UnaryExpression {
+        switch (node.name.escapedText) {
+            case "length":
+                return tstl.createUnaryExpression(
+                    this.transformExpression(node.expression), tstl.SyntaxKind.LengthOperator, node);
+            default:
+                throw TSTLErrors.UnsupportedProperty("string", node.name.escapedText as string, node);
```
tomblind: `"string"` should probably be `"LuaTable"` here

src/LuaTransformer.ts
```diff
@@ -3928,6 +3966,57 @@ export class LuaTransformer {
...
+        const params = this.transformArguments(expression.arguments, signature);
+
+        const getIndexExpression = () => {
+            if (expression.arguments.length === 0) {
+                throw TSTLErrors.ForbiddenLuaTableUseException(
+                    expression,
+                    "A parameter is required for set() or get() on a '@LuaTable' object."
+                );
+            } else {
+                return tstl.createTableIndexExpression(luaTable, params[0], expression);
```
tomblind: What if `set` receives more than 2 parameters, or `get` receives more than 1?

hazzard993: I was thinking this might be useful if something like `t.prop = 1, 0` needed to exist but I can't think of a use. `set()` now needs two parameters and `get()` can only take one

hazzard993:
- If a class is declared with the `LuaTable` directive without `declare` modifier the transpiler will error
- Classes extending a `LuaTable` will cause the transpiler to error
- `set()` takes two parameters, `get()` takes one

Also made the `length` property not re-assignable to avoid extra code gen

tomblind:
One more issue I see is explicit constructors. I can declare a constructor for the class, but it's basically ignored. We should throw an error or do something with the arguments.

Also, it would be good to have tests for the directive when used on an interface instead of a class.

As for assigning length, we now support that for array types with a lib function `__TS_ArraySetLength`. This could be used here as well.

hazzard993:
Now if a LuaTable is constructed with arguments the transpiler will throw an error saying no parameters are allowed.

I chose not to use `__TS_ArraySetLength` since it seems like the goal of a LuaTable object is to reduce code gen.

test/unit/luaTable.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+/** @LuaTable */
+declare class Table<K extends {} = {}, V = any> {
+    length: number;
+    constructor(notAllowed?: boolean);
+    set(key?: K, value?: V): void;
+    get(key?: K): V;
+    other(): void;
+}
+declare let tbl: Table;
+`;
```
tomblind: It would be good to run these tests against a `@luaTable` interface as well. Perhaps have 2 lib strings and do a `test.each()` to try both?

test/unit/luaTable.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+    expect(() =>
+        util.transpileString(tableLib + `const exp = tbl.set("value", 5)`),
+    ).toThrowExactError(TSTLErrors.ForbiddenLuaTableSetExpression(util.nodeStub));
+});
+
+test("LuaTables can have other properties", () => {
+    expect(() => util.transpileString(tableLib + `tbl.other()`)).not.toThrow();
+});
+
+test("LuaTable new and length", () => {
```
tomblind: This should probably be split into 2 separate tests

src/LuaTransformer.ts
```diff
@@ -4167,6 +4229,56 @@ export class LuaTransformer {
         }
     }
 
+    public transformLuaTableCallExpression<T extends boolean>(
+        expression: ts.CallExpression,
+        isWithinExpressionStatement: T
```
Perryvw: I don't like this parameter. A transformation of an expression should always return an expression. If you need a statement that is something I think you should handle at the call site.

src/LuaTransformer.ts
```diff
@@ -4167,6 +4229,56 @@ export class LuaTransformer {
...
+
+        switch (methodName) {
+            case "get":
+                if (expression.arguments.length !== 1) {
+                    throw TSTLErrors.ForbiddenLuaTableUseException(
+                        expression,
+                        "One parameter is required for get() on a '@LuaTable' object."
+                    );
+                }
+                if (isWithinExpressionStatement) {
```
Perryvw: I don't think I understand the reasoning here. If you have a `table.get(abc)` statement it is transformed to `____ = table.get(abc)`? That sounds like asking for all kinds of strange bugs.

src/TSTLErrors.ts
```diff
@@ -18,6 +18,22 @@ export class TSTLErrors {
...
+    public static ForbiddenLuaTableNonDeclaration = (node: ts.Node) =>
+        new TranspileError(`Classes with the '@luaTable' decorator must be declared.`, node);
+
+    public static InvalidExtendsLuaTable = (node: ts.Node) =>
+        new TranspileError(`Cannot extend classes with the decorator '@luaTable'.`, node);
+
+    public static InvalidInstanceOfLuaTable = (node: ts.Node) =>
+        new TranspileError(`The instanceof operator cannot be used with a '@luaTable' class.`, node);
+
+    public static ForbiddenLuaTableUseException = (node: ts.Node, description: string) =>
```
Perryvw: By convention we have node as the last parameter. I am not sure about having this function like this at all since it does nothing but passing the same parameters to the transpileError.

I would prefix the description with "Invalid @luaTable usage: " to give it some purpose in life.

src/LuaTransformer.ts
hazzard993: Since `table.get(abc)` transforms to `table[abc]` it becomes an invalid expression in Lua. In this case I chose to then assign this to an anonymous variable the same way `x;` is transformed to `local ____ = x`;

Perryvw: This sounds like a more fundamental issue. Table.get(abc) is an expression but you are returning a statement. This is related to my other comment on the function signature, I think it might automatically be resolved if you separate the assignment and expression functions.

src/LuaTransformer.ts
```diff
@@ -477,6 +477,19 @@ export class LuaTransformer {
...
+        // You cannot extend LuaTable classes
+        if (extendsType) {
+            const decorators = tsHelper.getCustomDecorators(extendsType, this.checker);
+            if (decorators.has(DecoratorKind.LuaTable)) {
+                throw TSTLErrors.InvalidExtendsLuaTable(statement);
+            }
+        }
+
+        // LuaTable classes must be declared
+        if (decorators.has(DecoratorKind.LuaTable) && !tsHelper.isDeclared(statement)) {
```
ark120202: Class may considered declared because of it's context
```suggestion
        if (decorators.has(DecoratorKind.LuaTable) && (ts.getCombinedModifierFlags(statement) & ts.ModifierFlags.Ambient) === 0) {
```

src/LuaTransformer.ts
```diff
@@ -4167,6 +4229,56 @@ export class LuaTransformer {
...
+        const method = expression.expression as ts.PropertyAccessExpression;
+        const methodName = method.name.escapedText;
+        const signature = this.checker.getResolvedSignature(expression);
+        const tableName = (method.expression as ts.Identifier).escapedText;
+        const luaTable = tstl.createIdentifier(tableName);
+        const params = this.transformArguments(expression.arguments, signature);
+
+        switch (methodName) {
+            case "get":
+                if (expression.arguments.length !== 1) {
```
ark120202: It also should check that it's argument isn't a SpreadElement

src/LuaTransformer.ts
```diff
@@ -4167,6 +4229,56 @@ export class LuaTransformer {
...
+                    throw TSTLErrors.ForbiddenLuaTableUseException(
+                        expression,
+                        "One parameter is required for get() on a '@LuaTable' object."
```
```suggestion
                        "One parameter is required for get() on a '@luaTable' object."
```

src/LuaTransformer.ts
```diff
@@ -4167,6 +4229,56 @@ export class LuaTransformer {
...
+                        tstl.createTableIndexExpression(luaTable, params[0], expression)
+                    );
+                } else {
+                    return tstl.createTableIndexExpression(luaTable, params[0], expression);
+                }
+            case "set":
+                if (params.length !== 2) {
+                    throw TSTLErrors.ForbiddenLuaTableUseException(
+                        expression,
+                        "Two parameters are required for set() on a '@LuaTable' object."
```
```suggestion
                        "Two parameters are required for set() on a '@luaTable' object."
```

src/TSTLErrors.ts
```diff
@@ -18,6 +18,22 @@ export class TSTLErrors {
     public static ForbiddenForIn = (node: ts.Node) =>
         new TranspileError(`Iterating over arrays with 'for ... in' is not allowed.`, node);
 
+    public static ForbiddenLuaTableSetExpression = (node: ts.Node) => new TranspileError(
+        `A '@LuaTable' object's 'set()' method can only be used as a Statement, not an Expression.`,
```
ark120202:
```suggestion
        `A '@luaTable' object's 'set()' method can only be used as a Statement, not an Expression.`,
```

src/LuaTransformer.ts
hazzard993: Not sure how to test this but good pickup

src/LuaTransformer.ts
hazzard993: I've separated this logic into three functions atm to keep this logic more consistent. I've grouped the error checking in `validateLuaTableCall` which takes `isWithinExpressionStatement` for validation purposes, `transformLuaTableExpressionStatement` and `transformLuaTableExpression`.

Would this be something more appropriate?

Perryvw: Seems okay, these methods should probably be private instead of public though.

src/LuaTransformer.ts
```diff
@@ -3851,6 +3912,16 @@ export class LuaTransformer {
         }
     }
 
+    public transformLuaTableProperty(node: ts.PropertyAccessExpression): tstl.UnaryExpression | undefined {
```
Perryvw: This one should probably also be private.

test/util.ts
```diff
@@ -102,13 +102,21 @@ export function transpileAndExecute(
...
+    ambientTsStr = "",
 ): any {
     const wrappedTsString = `${tsHeader ? tsHeader : ""}
         declare function JSONStringify(this: void, p: any): string;
         function __runTest(this: void): any {${tsStr}}`;
 
     const lua = `${luaHeader ? luaHeader : ""}
-        ${transpileString(wrappedTsString, compilerOptions, false)}
+        ${transpileString(
+            {
```
Perryvw: Probably better to change the signature of transpileAndExecute  to `tsStr: string | { [filename: string]: string }` (same as what transpileString accepts), and then just pass in this object from your test.

PR #521, opened 2019-04-13
ArraySetLength lib function

### LINKED ISSUE #514: array.length = 0 be translated to #array = 0
array.length = 0 be translated to #array = 0
typescript:
` 
let list = [1,2,3,4,5];
list.length = 0; //clear the array
`
lua:
`
local list = {1, 2, 3, 4, 5};
#list = 0;
`

tomblind:
fixes #514

### DIFF
--- a/src/lualib/ArraySetLength.ts
+++ b/src/lualib/ArraySetLength.ts
@@ -1,5 +1,10 @@
 function __TS__ArraySetLength<T>(this: void, arr: T[], length: number): number {
-    if (length < 0) {
+    if (length < 0
+        || length !== length // NaN
+        || length === (1 / 0) // Infinity
+        || length === (-1 / 0) // -Infinity
+        || Math.floor(length) !== length) // non-integer
+    {
         // tslint:disable-next-line:no-string-throw
         throw `invalid array length: ${length}`;
     }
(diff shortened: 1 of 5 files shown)


src/lualib/ArraySetLength.ts
```diff
@@ -0,0 +1,10 @@
+function __TS__ArraySetLength<T>(this: void, arr: T[], length: number): number {
+    if (length < 0) {
```
ark120202: In JS it fails for any non-integers (32 bit), which also includes `.length = 0.1`, `.length = Infinity` and `.length = NaN`.

tomblind:
Updated to handle NaN, Infinity, and non-integers

src/lualib/ArraySetLength.ts
```diff
@@ -0,0 +1,15 @@
+function __TS__ArraySetLength<T>(this: void, arr: T[], length: number): number {
+    if (length < 0
+        || length !== length // NaN
+        || length === (1 / 0) // Infinity
+        || length === (-1 / 0) // -Infinity
```
ark120202: I think `length < 0` should be enough to catch it

tomblind: good point

PR #538, opened 2019-04-26
false default binding pattern values

### LINKED ISSUE #529: Default values in destructuring handle false incorrectly
Input:
```ts
function test1(foo = true) {}
function test2({ foo = true }: { foo?: boolean }) {}
function test3([foo = true]: [boolean?]) {}
```

Current Result:
```lua
local function test1(self, foo)
    if foo == nil then
        foo = true
    end
end
local function test2(self, ____TS_bindingPattern0)
    local foo = ____TS_bindingPattern0.foo or true
end
local function test3(self, ____TS_bindingPattern0)
    local foo = ____TS_bindingPattern0[1] or true
end
```

Expected Result:
```lua
local function test1(self, foo)
    if foo == nil then
        foo = true
    end
end
local function test2(self, ____TS_bindingPattern0)
    local foo = ____TS_bindingPattern0.foo
    if foo == nil then
        foo = true
    end
end
local function test3(self, ____TS_bindingPattern0)
    local foo = ____TS_bindingPattern0[1]
    if foo == nil then
        foo = true
    end
end
```

hazzard993:
Closes #529

Default destructuring parameters are handled the same way default parameters are, with a nil check

### DIFF
--- a/test/unit/bindingpatterns.spec.ts
+++ b/test/unit/bindingpatterns.spec.ts
@@ -98,6 +98,7 @@ test.each([
         objectString: "{x: false, y: [false]}",
         returnVariable: "z",
     },
+    { bindingString: "[x = true]", objectString: "[false]", returnVariable: "x" },
 ])(
     "Binding patterns handle false correctly (%p)",
     ({ bindingString, objectString, returnVariable }) => {
--- a/test/unit/bindingpatterns.spec.ts
+++ b/test/unit/bindingpatterns.spec.ts
@@ -1,157 +1,110 @@
 import * as util from "../util";
 
 const testCases = [
-    { bindingString: "{x}", objectString: "{x: true}", returnVariable: "x", expectedResult: true },
-    {
-        bindingString: "[x, y]",
-        objectString: "[false, true]",
-        returnVariable: "y",
-        expectedResult: true,
-    },
-    {
-        bindingString: "{x: [y, z]}",
-        objectString: "{x: [false, true]}",
-        returnVariable: "z",
-        expectedResult: true,
-    },
-    {
-        bindingString: "{x: [, z]}",
-        objectString: "{x: [false, true]}",
-        returnVariable: "z",
-        expectedResult: true,
-    },
-    {
-        bindingString: "{x: [{y}]}",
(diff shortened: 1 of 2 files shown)


test/unit/bindingpatterns.spec.ts
```diff
@@ -1,92 +1,157 @@
...
-    { bindingString: "[x, y]", objectString: "[false, true]", returnVariable: "y" },
-    { bindingString: "{x: [y, z]}", objectString: "{x: [false, true]}", returnVariable: "z" },
-    { bindingString: "{x: [, z]}", objectString: "{x: [false, true]}", returnVariable: "z" },
-    { bindingString: "{x: [{y}]}", objectString: "{x: [{y: true}]}", returnVariable: "y" },
-    { bindingString: "[[y, z]]", objectString: "[[false, true]]", returnVariable: "z" },
-    { bindingString: "{x, y}", objectString: "{x: false, y: true}", returnVariable: "y" },
-    { bindingString: "{x: foo, y}", objectString: "{x: true, y: false}", returnVariable: "foo" },
+    { bindingString: "{x}", objectString: "{x: true}", returnVariable: "x", expectedResult: true },
+    {
+        bindingString: "[x, y]",
```
Perryvw: I don't think it's a good idea to adjust all existing tests for this. It is probably tidier if you just add a new `test.each([<...>])("Binding patterns handle false correctly (%p)"` with your false test cases.

test/unit/bindingpatterns.spec.ts
```diff
@@ -90,3 +90,21 @@ test.each(testCases)(
...
     },
 );
+
+test.each([
+    { bindingString: "{x, y = true}", objectString: "{x: false, y: false}", returnVariable: "y" },
+    {
+        bindingString: "{x, y: [z = true]}",
+        objectString: "{x: false, y: [false]}",
+        returnVariable: "z",
+    },
```
Perryvw: Could you add a test with simply the pattern `[x = false]` too?

hazzard993: Array binding patterns currently can't have default parameters. I'll add this with another commit, doesn't look like it resulted in much more code to review

PR #547, opened 2019-05-01
source map improvements

tomblind:
- support for sourceRoot compiler option
- trailing punctuation (commas, close-braces, etc...) and newlines are appended to the SourceNode they follow to avoid issues when mapping to a single lua line
- indents don't create a separate SourceNode as this caused some issues in places
- original ts nodes passed to more create functions


tomblind:
Meant to open this as a draft

PR #552, opened 2019-05-02
Load custom transformers from tsconfig

### LINKED ISSUE #433: Support typescript transformers
Required for a simple resolution of #178.
Covers many use cases of #179.

ark120202:
Resolves #433.


package.json
```diff
@@ -39,13 +39,15 @@
         "node": ">=8.5.0"
     },
     "dependencies": {
+        "resolve": "^1.10.1",
```
lolleko: Avoid runtime deps if possible.

ark120202: I think it's necessary in this case. Module resolution algorithm is complex and TS implementation isn't exposed and limited to types only. It's quite small, has only one sub-dependency and pretty much standard for customized module resolution, so IMO it's okay to add it. We'll need it for module support as well.

src/CompilerOptions.ts
```diff
@@ -1,12 +1,28 @@
...
+
+export type CompilerOptions = OmitIndexSignature<ts.CompilerOptions> & {
     noHeader?: boolean;
     luaTarget?: LuaTarget;
     luaLibImport?: LuaLibImportKind;
     noHoisting?: boolean;
     sourceMapTraceback?: boolean;
-}
+    tsTransformers?: TransformerConfig[];
+    [option: string]: ts.CompilerOptions[string] | TransformerConfig[];
```
Perryvw: Not sure I like this index type being here and the complexity to the type it adds, is this really required?

ark120202: Index signature in `ts.CompilerOptions` disallows adding new options with types that aren't defined in it. As a workaround `OmitIndexSignature` removes it from `ts.CompilerOptions` and later we re-add it with our added type.
Re-adding index signature is actually not strictly required there (because `ts.CompilerOptions` still would be a subtype), but it makes sense considering that `tstl.CompilerOptions` is supposed to be just `ts.CompilerOptions` with custom options.

PR #557, opened 2019-05-03
Add NaN, Infinity and related number functions

ark120202:


src/LuaTransformer.ts
```diff
@@ -3489,15 +3497,36 @@ export class LuaTransformer {
...
+                const numberParameters = tsHelper.isNumberType(expressionType)
+                    ? parameters
+                    : [this.transformLuaLibFunction(LuaLibFeature.Number, undefined, ...parameters)];
+
+                return this.transformLuaLibFunction(
+                    name === "isNaN" ? LuaLibFeature.NumberIsNaN : LuaLibFeature.NumberIsFinite,
+                    node,
+                    ...numberParameters
+                );
+            // TODO: default: throw TSTLErrors.UnsupportedGlobalFunction(name, node);
```
Perryvw: TODO

src/lualib/declarations/global.d.ts
```diff
@@ -1,3 +1,4 @@
+declare function tonumber(this: void, value: any, base?: number): number | undefined;
```
Perryvw: What do you think about adding `/** @noSelfInFile */` to this file instead of the `this: void`s?

test/unit/numbers.spec.ts
```diff
@@ -0,0 +1,77 @@
+import * as util from "../util";
+
+test.each([
+    "NaN === NaN",
+    "NaN !== NaN",
+    "NaN + NaN",
```
Perryvw: Might be nice to add `1 + NaN` and `NaN + 1` as testcases too

test/unit/objectLiteral.spec.ts
```diff
@@ -13,13 +13,34 @@ test.each([
...
+        const result = util.transpileAndExecute(`
+            const x = 1;
+            const o = { x };
+            return o.x;
+        `);
+
+        expect(result).toBe(1);
+    });
+
+    // TODO:
```
Perryvw: What is the TODO? Either solve it or add a clear description.

ark120202: This test shows currently invalid behavior, but probably it would be better to add it after it would be fixed

Issue #537, opened 2019-04-25
Feature request: Support constructor extension class decorator


ehemsley:
```ts
export function Reads(...message_type_args: Type<Message>[]) {
    return <T extends new(...args: any[]) => {}>(constructor: T) => {
        return class extends constructor {
            protected read_message_types = new Set<Type<Message>>(message_type_args);
        };
    };
}

@Reads(AMessage)
class AMessageReaderEngine extends Engine {}
```

I am using the following decorator pattern in my framework. It would be really nice if I could support it on the Lua side instead of having to declare the properties directly in the class, I find this much more readable.

Perryvw:
We are planning on adding this fairly soon, after finishing up some other large open PRs.

PR #558, opened 2019-05-03
Class decorators

ehemsley:


ehemsley:
Resolves #537

src/LuaTransformer.ts
```diff
@@ -1,7 +1,7 @@
 import * as path from "path";
 import * as ts from "typescript";
 import { CompilerOptions, LuaTarget } from "./CompilerOptions";
-import { DecoratorKind } from "./Decorator";
+import { DecoratorKind, Decorator } from "./Decorator";
```
Perryvw: Looks like an accidental import. (We should probably rename this to Directives or something soon to prevent confusion)

src/LuaTransformer.ts
```diff
@@ -5113,4 +5117,46 @@ export class LuaTransformer {
...
+
+        const LuaLibDecorateIdentifier = this.importAndCreateIdentifierForLuaLibFunction(LuaLibFeature.Decorate);
+
+        return tstl.createAssignmentStatement(
+            className,
+            tstl.createCallExpression(LuaLibDecorateIdentifier, decoratorArguments)
+        );
+    }
+
+    private importAndCreateIdentifierForLuaLibFunction(
```
Perryvw: This refactor is not needed and should be reverted.

src/LuaTransformer.ts
```diff
@@ -5113,4 +5117,46 @@ export class LuaTransformer {
...
+        const decoratorArguments: tstl.Expression[] = [];
+
+        const decoratorTable = tstl.createTableExpression(
+            decoratorExpressions.map(expression => tstl.createTableFieldExpression(expression))
+        );
+
+        decoratorArguments.push(decoratorTable);
+        decoratorArguments.push(className);
+
+        const LuaLibDecorateIdentifier = this.importAndCreateIdentifierForLuaLibFunction(LuaLibFeature.Decorate);
```
Perryvw: This (and the lines below) can probably just be `this.transformLuaLibFunction(LuaLibFeature.Decorate, className, ...decoratorArguments);`

test/unit/classDecorator.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+    @SetNum(410)
+    class SubTestClass extends TestClass {}
+
+    const classInstance = new SubTestClass();
+    return classInstance.decoratorNum + classInstance.decoratorTen;
+    `;
+
+    const result = util.transpileAndExecute(source);
+    expect(result).toBe(420);
+});
```
Perryvw: Could you add a testcase that ensures decorators are applies in order? Assigning different characters to a string would do.

src/LuaTransformer.ts
```diff
@@ -5113,4 +5118,37 @@ export class LuaTransformer {
...
+
+        const decoratorExpressions: tstl.Expression[] = [];
+        for (const decorator of decorators) {
+            const expressionVisitResult = this.transformExpression(decorator.expression);
+            if (expressionVisitResult !== undefined) {
+                decoratorExpressions.push(expressionVisitResult);
+            }
+        }
+
+        const decoratorArguments: tstl.Expression[] = [];
```
ark120202: Since we can't have a helper accept both functions with and without context I think we should add context there and add a check for function kind.

src/LuaTransformer.ts
```diff
@@ -5113,4 +5118,37 @@ export class LuaTransformer {
             return visitResult;
         }
     }
+
+    public generateConstructorDecorationStatement(
```
ark120202: I think `generate` is kind of out of place there, usually such methods start with `transform` or `create`.

src/LuaTransformer.ts
```diff
@@ -5113,4 +5118,37 @@ export class LuaTransformer {
...
+        declaration: ts.ClassLikeDeclaration
+    ): tstl.AssignmentStatement | undefined {
+        const className = declaration.name !== undefined
+            ? this.transformIdentifier(declaration.name)
+            : tstl.createAnonymousIdentifier();
+
+        const decorators = declaration.decorators;
+        if (!decorators) { return undefined; }
+
+        const decoratorExpressions: tstl.Expression[] = [];
```
ark120202: Might be a bit simpler with functional style

```ts
const decoratorExpressions = this.filterUndefined(
    decorators.map(decorator => this.transformExpression(decorator.expression))
);
```

src/lualib/Decorate.ts
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * SEE: https://github.com/Microsoft/TypeScript/blob/master/src/compiler/transformers/ts.ts#L3598
+ */
+function __TS__Decorate(this: void, decorators: Function[], target: {}, key?: string, desc?: any): {} {
+    let r = target;
+
+    for (let i = decorators.length; i >= 0; i--) {
+        const d = decorators[i];
+        if (d) {
+            r = (key === undefined ? d(r) : desc !== undefined ? d(target, key, r) : d(target, key)) || r;
```
ark120202: I think there's no much reason to have helpers be 1:1 with TS ones - that condition generates quite unoptimized code with few IIFEs and variables may be renamed for readability, like:

```ts
function __TS__Decorate(this: void, decorators: Function[], target: {}, key?: string, desc?: any): {} {
    let result = target;

    for (let i = decorators.length; i >= 0; i--) {
        const decorator = decorators[i];
        if (decorator) {
            const oldResult = result;

            if (key === undefined) {
                result = decorator(result);
            } else if (desc !== undefined) {
                result = decorator(target, key, result);
            } else {
                result = decorator(target, key);
            }

            result = oldResult || result;
        }
    }

    return result;
}
```

Perryvw: True, this may look like optimized js, but it does not necessarily lead to optimal lua.

Perryvw: Note that in ark's code I think he meant the `result =` assignments to be `oldResult =` in the if statement.

ark120202: Oh, I meant `result = result || oldResult`, but either one would work (though naming would be a little weird if `oldResult` was reassigned)

test/unit/classDecorator.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+    @SetString("the quick brown fox")
+    @SetString("jumped")
+    @SetString("over the lazy dog")
+    class TestClass {
+        public decoratorString = "";
+    }
+
+    const classInstance = new TestClass();
+    return classInstance.decoratorString;
+    `;
```
Perryvw: I think this test can be improved since it only really shows the result of one of the decorators. How about something like this:
```ts
const order = [];

function SetString(this: void, stringArg: string) {
    order.push("eval " + stringArg);
    return <T extends new (...args: any[]) => {}>(constructor: T) => {
        order.push("execute " + stringArg);
        return class extends constructor {
            decoratorString = stringArg;
        };
    };
}

@SetString("the quick brown fox")
@SetString("jumped")
@SetString("over the lazy dog")
class TestClass {
    public static decoratorString = "";
}

const inst = new TestClass();
return JSONStringify(order);
```

src/LuaTransformer.ts
ehemsley: Can you elaborate on this for me? I'm not really sure what this means.

ark120202: Ah, don't mind this line reference, I was a bit confused because all decorator factories in tests have `this: void` (they probably shouldn't?).

Inner decorator function still should be checked for function kind. If it has `this: void` it would transpile successfully, but fail at runtime. See `validateFunctionAssignment` for a similar check used when function is passed as a parameter.

Perryvw: For clarification: The (precompiled) Lualib function always calls the decorators with context. If a callback is passed with `this: void` it should error out.

test/unit/classDecorator.spec.ts
```diff
@@ -0,0 +1,167 @@
...
+    return classInstance.decoratorBool;
+    `;
+
+    const result = util.transpileAndExecute(source);
+    expect(result).toBe(true);
+});
+
+test("Class decorator with parameters", () => {
+    const source = `
+    function SetNum(this: void, numArg: number) {
```
Perryvw: Check if decorator factories work without `this: void`, they should work with and without it.

src/LuaTransformer.ts
```diff
@@ -5356,4 +5361,39 @@ export class LuaTransformer {
             return visitResult;
         }
     }
+
+    public createConstructorDecorationStatement(
```
Perryvw: This should probably be private.

PR #561, opened 2019-05-04
Feature/api interface

Perryvw:
Cleaned up visitorPattern interface in printer and transformer.


src/LuaTransformer.ts
```diff
@@ -1148,8 +1147,7 @@ export class LuaTransformer {
 
     public transformSetAccessorDeclaration(
         setAccessor: ts.SetAccessorDeclaration,
-        className: tstl.Identifier,
-        classDeclaration: ts.ClassLikeDeclaration
```
lolleko: Are you sure you want to remove the declaration parameter? I think there could be use cases were you would want to access it.

Or is it possible to access the ClassDeclaration via the SetAccessorDeclaration ?

src/LuaPrinter.ts
```diff
@@ -675,7 +675,11 @@ export class LuaPrinter {
         return aliveStatements;
     }
 
-    private joinChunks(separator: string, chunks: SourceChunk[]): SourceChunk[] {
+    protected isEmptyStatement(statement: tstl.Statement): boolean {
```
lolleko: Should probably name that to `isStatementEmpty`?

src/LuaTransformer.ts
Perryvw: setAccessor.parent should be the class declaration. I would also prefer to remove the className actually, but I'm not sure yet how exactly.

PR #440, opened 2019-02-21
ignore interface imports

endel:
Hey there! I'm creating this pull-request as a draft, I haven't checked the whole source-code and what side effects this change may add. Let me know if you have any feedback, I'll gladly change it.

Cheers!

PR #565, opened 2019-05-05
 Add module import/export elision

### LINKED ISSUE #178: Only generate import if import is used in value position
https://github.com/Microsoft/TypeScript/wiki/FAQ#why-are-imports-being-elided-in-my-emit

### LINKED ISSUE #494: Declaration files  should not generate import statements
Declaration files will not be transpiled, but imports from them will still create a `require(declaration)`, which causes a lua error due to the file not existing.

Ideally we also find a way to not import declared types and interfaces from non-declarations files.

### LINKED ISSUE #555: exported type
`____exports.New_Data = New_Data` needs to be removed

https://gitlab.com/the-language/the-language/blob/92239364710436f0ed16bae6a0c7460648cdcaeb/core/pure/typescript/lang.ts#L420
https://gitlab.com/the-language/the-language/blob/92239364710436f0ed16bae6a0c7460648cdcaeb/core/pure/typescript/lang.ts#L433
```typescript
type New_Data<X extends LangVal, Y extends LangVal> = LangValDataG<X, Y>
export { New_Data, new_data, data_p, data_name, data_list }
```

https://gitlab.com/the-language/the-language/blob/92239364710436f0ed16bae6a0c7460648cdcaeb/core/pure/lua/lang.lua#L1105
```lua
____exports.New_Data = New_Data
```

tsc result
https://gitlab.com/the-language/the-language/blob/92239364710436f0ed16bae6a0c7460648cdcaeb/core/pure/ecmascript6/lang.js#L116
```js
export { new_data, data_p, data_name, data_list };
```

ark120202:
Resolves #178, resolves #494, fixes #555, closes #440.


src/LuaTransformer.ts
```diff
@@ -47,11 +57,13 @@ export class LuaTransformer {
     private isStrict: boolean;
     private luaTarget: LuaTarget;
 
-    private checker: ts.TypeChecker;
+    private checker: DiagnosticsProducingTypeChecker;
     protected options: CompilerOptions;
 
-    private isModule = false;
+    // Resolver is lazy-initialized in transformSourceFile to avoid type-checking all files
+    private resolver!: EmitResolver;
```
tomblind: It might be better to do this as an optional `?` so it has to be verified before used, since it's set in a public api function (`transformSourceFile`) which could be overridden.

ark120202: It's true, but I think it complicates it's usage quite a lot with all these undefined checks. If it's not defined it would fail immediately anyway, just with a bit less clear error message.
Maybe we could have some different (internal?) method that would initialize state and then call `transformSourceFile` as a public visitor?

tomblind: Looking at this more, `setupState` couldn't be called if someone overrode `transformSourceFile` without calling the super version. So, I think you're suggestion here is probably necessary. Transformation should probably be kicked off by a non-override-able function.

test/unit/modules.spec.ts
```diff
@@ -2,6 +2,86 @@ import * as tstl from "../../src";
...
+    test("should elide namespace imports with unused values", () => {
+        expectToElideImport(`
+            import * as module from "module";
+            const foo: module.Type = "bar";
+        `);
+    });
+
+    test("should elide type exports", () => {
+        const code = `
+            // TODO: Add some Lua Types to transpileStringResult
```
Perryvw: Was this TODO meant for this PR?

ark120202: No, it's just a reminder to remove it once we'll have some global types there. Although it should error anyway if this symbol would be declared, so I'll remove it.

PR #568, opened 2019-05-05
Support rest pattern in destructuring

### LINKED ISSUE #545: Ellipsis destruction is not allowed

ark120202:
Resolves #545.


tomblind:
One way ObjectRest could be improved is by passing a set-like object for usedProperties instead of an array that has to be manually searched:
```ts
const {a, b, ...rest} = {a: "A", b: "B", c: "C", d: "D"};
```
```lua
function __TS__ObjectRest(target, usedProperties)
    local result = {}
    for property in pairs(target) do
        if usedProperties[property] == nil then
            result[property] = target[property]
        end
    end
    return result
end

local ____ = {
    a = "A",
    b = "B",
    c = "C",
    d = "D",
}
a = ____.a
b = ____.b
rest = __TS__ObjectRest(____, {
    a = true,
    b = true,
    rest = true,
})
```

src/LuaTransformer.ts
```diff
@@ -3362,6 +3388,18 @@ export class LuaTransformer {
 
                     result.push(assignmentStatement);
                     break;
+                case ts.SyntaxKind.SpreadElement:
+                    if (index !== node.elements.length - 1) break;
```
Perryvw: Should we just ignore this or should this throw an error? Is this valid TS?

src/TSHelper.ts
```diff
@@ -60,6 +60,14 @@ export function isAssignmentPattern(node: ts.Node): node is ts.AssignmentPattern
     return ts.isObjectLiteralExpression(node) || ts.isArrayLiteralExpression(node);
 }
 
+export function isDestructuringAssignment(node: ts.Node): node is ts.DestructuringAssignment {
```
Perryvw: Typescript has this one too, any reason we're not using that?

src/LuaTransformer.ts
```diff
@@ -2025,17 +2055,15 @@ export class LuaTransformer {
...
-
             const statements: tstl.Statement[] = [];
 
-            // For nested bindings and object bindings, fall back to transformBindingPattern
+            // For object, nested, omitted or rest bindings fall back to transformBindingPattern
             if (
                 ts.isObjectBindingPattern(statement.name) ||
                 statement.name.elements.some(
-                    elem => (!ts.isBindingElement(elem) || !ts.isIdentifier(elem.name)) && !ts.isOmittedExpression(elem)
+                    e => ts.isBindingElement(e) && (!ts.isIdentifier(e.name) || e.dotDotDotToken)
```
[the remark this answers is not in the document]
ark120202: `BindingElement.name` has a `BindingName` type, which is a union of `Identifier` and `BindingPattern`, so `ts.isIdentifier(e.name)` is true for `const [foo] = ...` and false for `const [[foo]] = ...`.

src/LuaTransformer.ts
ark120202: It's invalid TS and that's exactly why we shouldn't throw there. TypeScript already provides a diagnostic, so there's no point to break compilation for user.

src/TSHelper.ts
ark120202: Are you sure about that? I couldn't find any functions that have a `is ts.DestructuringAssignment` type guard

Perryvw: https://github.com/microsoft/TypeScript/blob/master/src/compiler/utilities.ts#L4023

src/LuaTransformer.ts
Perryvw: I agree.

src/TSHelper.ts
ark120202: It's in the block with `@internal` annotation, so can't be used directly

PR #569, opened 2019-05-05
Support for spread operator on iterables

### LINKED ISSUE #534: Arrays initialized with a spread operator on an iterator are empty
```ts
  const items = new Map<number, string>();
  items.set(2, "hello");
  items.set(5, "howdy");
  const new_array = [...items.entries()];
``` 
new_array will be empty in Lua.

I understand if this is not planned to be supported, but it would be nice to at least have the transpiler yell at me instead of having to debug an obscure issue in the transpiled code due to the empty array.

tomblind:
fixes #534

### DIFF
--- a/test/unit/spreadElement.spec.ts
+++ b/test/unit/spreadElement.spec.ts
@@ -47,3 +47,23 @@ test("Spread Element Lua JIT", () => {
     const lua = util.transpileString(`[...[0, 1, 2]]`, options);
     expect(lua).toBe("local ____ = {unpack({\n    0,\n    1,\n    2,\n})}");
 });
+
+test("Spread Element Iterable", () => {
+    const code = `
+        const it = {
+            i: 10,
+            [Symbol.iterator]() {
+                return this;
+            },
+            next() {
+                --this.i;
+                return {
+                    value: this.i,
+                    done: this.i < 0,
+                }
+            }
+        };
+        const arr = [...it];
+        return JSONStringify(arr)`;
+    expect(util.transpileAndExecute(code)).toBe(JSON.stringify([9, 8, 7, 6, 5, 4, 3, 2, 1, 0]));
+});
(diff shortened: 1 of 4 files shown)


test/unit/spreadElement.spec.ts
```diff
@@ -47,3 +47,23 @@ test("Spread Element Lua JIT", () => {
...
+                --this.i;
+                return {
+                    value: this.i,
+                    done: this.i < 0,
+                }
+            }
+        };
+        const arr = [...it];
+        return JSONStringify(arr)`;
+    expect(util.transpileAndExecute(code)).toBe(JSON.stringify([9, 8, 7, 6, 5, 4, 3, 2, 1, 0]));
```
Perryvw: I think the other way around would be a bit nicer:
```ts
const result = util.transpileAndExecute(code);
expect(JSON.parse(result)).toEqual([9, 8, 7, 6, 5, 4, 3, 2, 1, 0]);
```

test/unit/spreadElement.spec.ts
```diff
@@ -47,3 +47,23 @@ test("Spread Element Lua JIT", () => {
...
+
+test("Spread Element Iterable", () => {
+    const code = `
+        const it = {
+            i: 10,
+            [Symbol.iterator]() {
+                return this;
+            },
+            next() {
+                --this.i;
```
Perryvw: I think making this just count up is a bit clearer, first time I looked at this test I thought it was bugged and reversed the order.

test/unit/spreadElement.spec.ts
tomblind: I usually do this on purpose to ensure we're not somehow getting array indicies instead of values.

test/unit/spreadElement.spec.ts
tomblind: I usually do this on purpose to ensure we're not somehow getting array indicies instead of values.

Perryvw: I get that, but I'd sooner opt for adding a constant to each value, or multiplying them by 3 or something like that.

tomblind: Ok, I'll do something different here

PR #570, opened 2019-05-05
Small improvements

ark120202:


src/LuaAST.ts
```diff
@@ -878,7 +878,7 @@ export function createTableIndexExpression(
     return expression;
 }
 
-export type IdentifierOrTableIndexExpression = Identifier | TableIndexExpression;
+export type LeftHandSideExpression = Identifier | TableIndexExpression;
```
Perryvw: I think LeftHandSideExpression is probably a bit _too_ generic. What about `AssignmentLeftHandSideExrpession` or `AssignmentTargetExpression`?

src/lualib/Map.ts
```diff
@@ -1,34 +1,31 @@
-/** @tupleReturn */
-declare function next<TKey, TValue>(this: void, t: { [k: string]: TValue }, index?: TKey): [TKey, TValue];
+Map = class Map<K, V> {
```
Perryvw: What is the advantage of this notation over what it was previously?

test/unit/array.spec.ts
```diff
@@ -10,7 +10,7 @@ test("Array access", () => {
 
 test("Readonly Array access", () => {
     const result = util.transpileAndExecute(
-        `const arr: ReadonlyArray<number> = [3,5,1];
+        `const arr: readonly number[] = [3,5,1];
```
Perryvw: Maybe would be nice to test both these cases actually.

PR #571, opened 2019-05-05
Add string.{repeat,padStart,padEnd}

ark120202:


test/unit/string.spec.ts
```diff
@@ -309,12 +320,35 @@ test.each<{ inp: string; args: Parameters<string["endsWith"]> }>([
...
-    { input: "abcd", index: 3 },
-    { input: "abcde", index: 3 },
-    { input: "abcde", index: 0 },
-    { input: "a", index: 0 },
-])("string index (%p)", ({ input, index }) => {
-    const result = util.transpileAndExecute(`return "${input}"[${index}];`);
+    { inp: "hello test", count: 0 },
+    { inp: "hello test", count: 1 },
+    { inp: "hello test", count: 2 },
+])("string.repeat (%p)", ({ inp, count }) => {
```
Perryvw: Missing a testcase where count is not an integer.

PR #575, opened 2019-05-09
Sourcemap fixes

tomblind:
- Node positions are no longer hierarchical (this caused a lot of undesired effects in sourcemaps)
- Source maps are now built with a custom method to ensure no 'empty' mappings are generated
- Adjustments to which original ts nodes are used for positions on imports, class boilerplate statements, and for...of temps
- A few additional tests added for source mappings

There's still plenty more that can be done (mostly handling mappings for extra generated code like IIFEs and generators). But, this covers a bunch of common cases.

### DIFF
--- a/src/LuaPrinter.ts
+++ b/src/LuaPrinter.ts
@@ -1,6 +1,6 @@
 import * as path from "path";
 
-import {SourceNode, SourceMapGenerator} from "source-map";
+import { Mapping, SourceNode, SourceMapGenerator } from "source-map";
 
 import * as tstl from "./LuaAST";
 import { CompilerOptions, LuaLibImportKind } from "./CompilerOptions";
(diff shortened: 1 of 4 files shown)


src/LuaPrinter.ts
```diff
@@ -59,22 +59,23 @@ export class LuaPrinter {
 
         const rootSourceNode = this.printImplementation(block, luaLibFeatures, sourceFile);
 
-        const codeWithSourceMap = rootSourceNode
-            // TODO is the file: part really required? and should this be handled in the printer?
-            .toStringWithSourceMap({file: path.basename(sourceFile, path.extname(sourceFile)) + ".lua"});
+        const sourceRoot = this.options.sourceRoot
+            || (this.options.outDir ? path.relative(this.options.outDir, this.options.rootDir || process.cwd()) : ".");
```
Perryvw: Tsc makes the source filename either only the filename or maybe a relative path (Definitely no absolute paths), but this does produce absolute paths, which should probably be changed.

Issue #429, opened 2019-02-20
Support invalid lua identifiers


ark120202:
Would be a more general solution for #428. Related to #188.

Of course such property names are anti-pattern, but it's linters responsibility to detect anti-patterns, plus it's easy to support.

`const then = 1` -> `local ____then = 1`
`const ɥɣɎɌͼƛಠ = 1` -> `local ____613_611_590_588_892_411_3232 = 1`
`const x = { ɥɣɎɌͼƛಠ: 1 }` -> `local x = { ["ɥɣɎɌͼƛಠ"] = 1 }`
`x.ɥɣɎɌͼƛಠ` -> `x["ɥɣɎɌͼƛಠ"]`

PR #577, opened 2019-05-14
Allow lua keywords

### LINKED ISSUE #188: Mixing typeof and parameters named type

### LINKED ISSUE #536: Declaring a variable "error" and then throwing it causes a crash
```ts
const error = new MyError()
throw error.toString()
```
transpiles to

```lua
        local error = MyError.new()
        error(tostring(error))
```
which throws the error
`attempt to call local 'error' (a table value)`

tomblind:
fixes #536 
fixes #188 

Identifiers with names that would conflict with lua keywords/built-ins are prepended with 4 underscores in the output lua. This can be expanded later to address everything in #429.

### DIFF
--- a/src/LuaTransformer.ts
+++ b/src/LuaTransformer.ts
@@ -21,6 +21,8 @@ export enum ScopeType {
 interface SymbolInfo {
     symbol: ts.Symbol;
     firstSeenAtPos: number;
+    name: string; // Transformed name (modified if un-exported lua keyword)
+    localName?: string; // Name used for local version of exported class or module
 }
 
 interface FunctionDefinitionInfo {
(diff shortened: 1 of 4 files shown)


tomblind:
Sorry, this was a bit premature. It needs some refactoring and more tests.

PR #582, opened 2019-05-15
Fixed @tupleReturn on interface function properties

### LINKED ISSUE #578: TupleReturn only works on method definitions, not on function-typed variables.
Consider the following TypeScript:
```ts
interface MyInterface {
    /** @tupleReturn */
    getRaw(name: string): [number, number];
    /** @tupleReturn */
    getRaw2: (name: string) => [number, number];
}
declare var inst: MyInterface;

const [a, b] = inst.getRaw("abc");
const [c, d] = inst.getRaw2("abc");
```

I would expect both `getRaw` and `getRaw2` to respect the tupleReturn directive and produce the same code, but instead this produces the following code:

```lua
local a, b = inst:getRaw("abc")
local c, d = table.unpack(inst:getRaw2("abc"))
```

[Playground link](https://typescripttolua.github.io/play.html#src=interface%20MyInterface%20%7B%0A%20%20%20%20%2F**%20%40tupleReturn%20*%2F%0A%20%20%20%20getRaw(name%3A%20string)%3A%20%5Bnumber%2C%20number%5D%3B%0A%20%20%20%20%2F**%20%40tupleReturn%20*%2F%0A%20%20%20%20getRaw2%3A%20(name%3A%20string)%20%3D%3E%20%5Bnumber%2C%20number%5D%3B%0A%7D%0Adeclare%20var%20inst%3A%20MyInterface%3B%0A%0Aconst%20%5Ba%2C%20b%5D%20%3D%20inst.getRaw(%22abc%22)%3B%0A%0Aconst%20%5Bc%2C%20d%5D%20%3D%20inst.getRaw2(%22abc%22)%3B)

tomblind:
fixes #578 

Also fixed detection of @tupleReturn on object literal methods/function-properties that are assigned to interaces:
```ts
interface Foo {
    /** @tupleReturn **/
    method(): [number, number];

    /** @tupleReturn **/
    funcProp: () => [number, number];
}
const foo: Foo = {
    method() { return [3, 4]; }, // Should not wrap return value
    funcProp: () => [3, 4], // Also should not wrap return value
};
```

### DIFF
--- a/src/TSHelper.ts
+++ b/src/TSHelper.ts
@@ -218,10 +218,8 @@ export class TSHelper {
                 if (propertySymbol) {
                     functionType = checker.getTypeOfSymbolAtLocation(propertySymbol, declaration);
                 }
-                if (functionType === undefined) {
-                    functionType = checker.getTypeAtLocation(declaration);
-                }
-            } else {
+            }
+            if (functionType === undefined) {
                 functionType = checker.getTypeAtLocation(declaration);
             }
 
--- a/src/TSHelper.ts
+++ b/src/TSHelper.ts
@@ -208,9 +208,19 @@ export class TSHelper {
     public static isInTupleReturnFunction(node: ts.Node, checker: ts.TypeChecker): boolean {
         const declaration = TSHelper.findFirstNodeAbove(node, ts.isFunctionLike);
         if (declaration) {
-            let functionType: ts.Type;
+            let functionType: ts.Type | undefined;
             if (ts.isFunctionExpression(declaration) || ts.isArrowFunction(declaration)) {
                 functionType = TSHelper.inferAssignedType(declaration, checker);
+            } else if (ts.isMethodDeclaration(declaration) && ts.isObjectLiteralExpression(declaration.parent)) {
+                // Manually lookup type for object literal properties declared with method syntax
+                const interfaceType = TSHelper.inferAssignedType(declaration.parent, checker);
+                const propertySymbol = interfaceType.getProperty(declaration.name.getText());
+                if (propertySymbol) {
+                    functionType = checker.getTypeOfSymbolAtLocation(propertySymbol, declaration);
+                }
+                if (functionType === undefined) {
+                    functionType = checker.getTypeAtLocation(declaration);
+                }
             } else {
                 functionType = checker.getTypeAtLocation(declaration);
             }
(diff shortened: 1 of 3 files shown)


test/unit/tuples.spec.ts
```diff
@@ -119,6 +119,19 @@ test("Tuple Static Method Return Destruct", () => {
...
+    const result = util.transpileAndExecute(
+        `class Test {
+            /** @tupleReturn */
+            static tuple: () => [number, number, number] = () => [3,5,1];
+        }
+        const [a,b,c] = Test.tuple();
+        return b;`,
+    );
+
+    expect(result).toBe(5);
```
Perryvw: I guess it works, but would any of these tests fail if `@tupleReturn` was just ignored?

src/TSHelper.ts
```diff
@@ -208,9 +208,19 @@ export class TSHelper {
     public static isInTupleReturnFunction(node: ts.Node, checker: ts.TypeChecker): boolean {
         const declaration = TSHelper.findFirstNodeAbove(node, ts.isFunctionLike);
         if (declaration) {
-            let functionType: ts.Type;
+            let functionType: ts.Type | undefined;
```
Perryvw: I don't quite understand how this does not break on line 299 if `functionType` is undefined. Maybe this should be initialized as `checker.getTypeAtLocation(declaration)` be default and then overwritten if appropriate?

tomblind: TS can infer that it can't be `undefined` at that point because every branching path sets it to a value. We could initialize it, but it adds unnecessary overhead for some code paths.

test/unit/tuples.spec.ts
tomblind: That's a problem with all of these tests (even the previously existing ones). Perhaps we could check the transpiled lua for `unpack`?

src/TSHelper.ts
Perryvw: Aha I see, well maybe just removing the `else` and moving the `if (functionType === undefined) { functionType = ` as a separate step after this if if statement would be clearner then.

test/unit/tuples.spec.ts
Perryvw: Yeah that's what I was considering, I don't think we really have a good alternative?

src/TSHelper.ts
```diff
@@ -208,10 +208,18 @@ export class TSHelper {
...
             if (ts.isFunctionExpression(declaration) || ts.isArrowFunction(declaration)) {
                 functionType = TSHelper.inferAssignedType(declaration, checker);
-            } else {
+            } else if (ts.isMethodDeclaration(declaration) && ts.isObjectLiteralExpression(declaration.parent)) {
+                // Manually lookup type for object literal properties declared with method syntax
+                const interfaceType = TSHelper.inferAssignedType(declaration.parent, checker);
+                const propertySymbol = interfaceType.getProperty(declaration.name.getText());
```
Perryvw: Are these steps required? Can this not be handled the same as functions or arrow functions?

tomblind: `getContextualType` (which is what `inferAssignedType uses`) won't take a method declaration because it's not considered an expression.

PR #583, opened 2019-05-16
Remove new from luaKeywords

hazzard993:
`new` is not a Lua keyword

```lua
local new = 0
print(new)
```

Output: `0`

[Busted](https://olivinelabs.com/busted/) (a Lua testing framework) uses this keyword in `spy.new(...)`.

It is possible to use `new` as a property name in TS.

```ts
const object = {
    new: () => {}
};
object.new();
```

At the moment TSTL thinks it is a keyword and disallows this from working.

### DIFF
--- a/src/LuaTransformer.ts
+++ b/src/LuaTransformer.ts
@@ -40,7 +40,7 @@ interface Scope {
 
 export class LuaTransformer {
     public luaKeywords: Set<string> = new Set([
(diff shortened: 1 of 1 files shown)


tomblind:
#581 will re-enable use of the keywords in that list. So, this probably isn't needed.

The reason `new` is listed is to prevent users from accidentally breaking classes:
```ts
class Foo {
    static new() {}
}
const bar = new Foo(); // uh oh
```
I think a special check for this case is the better way to go.

PR #585, opened 2019-05-18
Smart for loop

tomblind:
Allows certain for loops to be transpiled as lua numerical for loops.

```ts
for (let i = 0; i < limit; ++i) {}
```
=>
```lua
for i = 0, limit - 1 do
end
```
- `i` must be declared with `let` and not modified inside the loop body
- condition must be one of:
  - `i < limit` or `i <= limit` for forward (addition) loops
  - `i > limit` or `i >= limit` for reverse (subtraction) loops
- `limit` must be either:
  - a const or number literal, when condition uses `<=` or `>=`
  - a **whole** number literal, when condition uses `<` or `>` (to avoid issues with offsetting)
- the incrementor must be one of:
  - unary increment or decrement (`++i`, `--i`, etc...)
  - compound assignment addition or subtraction (`i += x` or `i -= x`)
  - assignment to addition or subtraction operation using same variable (`i = i + x` or `i = i - x`)

There is also support for a special case:
```ts
for (let i = 0, limit = x; i < limit; ++i) {}
```
=>
```lua
for i = 0, x - 1 do
end
```
Since this is a common paradigm in (oldschool) javscript, it seemed like a good thing to support. Note that it will only be transpiled as a numerical `for` if the `limit` variable is not referenced in the the loop body at all.

Also included in this PR:
- `for...of` loops over arrays now use `ipairs` instead of a numerical lua for loop
- dis-allowed object destructuring in `for...of` loops (this silently produced bad code before)
- fixed issues with source mappings of lua `for...in` statements


tomblind:
I'm taking this out of draft as I'm not seeing any ways to simplify/clarify the code any better.

ark120202:
Offset limit when using < or > is always 1 now, but it should be equal to increment:

```ts
for (let i = 0; i < 2; i += 0.5) {
    console.log(i); // 0 0.5 1 1.5
}
```
```lua
for i = 0, 1, 0.5 do
    print(i) -- 0 0.5 1
end
```

tomblind:
After looking at this more closely, setting the offset to the step value only works in certain situations.
```ts
for (let i = 0; i < 1; i += 0.3) {
    console.log(i); // 0, 0.3, 0.6, 0.9
}
```
```lua
for i = 0, 1 - 0.3, 0.3 do
    print(i) -- 0, 0.3, 0.6
end
```

tomblind:
For now, I'm dis-allowing optimization on loops using `<` or `>` that don't have a whole number literal as a step value.

ark120202:
True, what about using epsilon?

```lua
for i = 0, 1 - 2.220446049250313e-16, 0.3 do
    print(i) -- 0, 0.3, 0.6, 0.9
end

for i = 0, 2 - 2.220446049250313e-16, 0.5 do
    print(i) -- 0, 0.5, 1, 1.5
end
```

tomblind:
I considered that, but any value we use could have precision issues in some lua implementations. It's too bad lua doesn't have a built in const for that like math.huge

tomblind:
Closing as implicit optimization is out of scope of the main TSTL project. This is better done as a plugin.

PR #586, opened 2019-05-19
fixed usage of lua keywords as property names

tomblind:


src/LuaTransformer.ts
```diff
@@ -3775,13 +3769,18 @@ export class LuaTransformer {
...
                     || tsHelper.getDeclarationContextType(signatureDeclaration, this.checker) !== ContextType.Void)
                 {
-                    // table:name()
-                    return tstl.createMethodCallExpression(
-                        table,
-                        this.transformIdentifier(node.expression.name),
-                        parameters,
-                        node
-                    );
+                    if (this.isUnsafeName(node.expression.name.text)) {
```
ark120202: It shouldn't check for builtins there

ark120202:
Some inconsistent behavior:

```ts
declare var local: any;
declare var _G: any;

const a1 = { local };
const a2 = { local: local };
const b1 = { _G };
const b2 = { _G: _G };
```

```lua
local a1 = {["local"] = ____local}
local a2 = {["local"] = local}
local b1 = {_G = _____G}
local b2 = {_G = _G}
```

For `_G` there it should always be `{_G = _G}`, but it's more complex for `local`. Because we also could have `_G.local` somewhere it could be `{["local"] = _G["local"]}`, but because we also have env it should be `{["local"] = _ENV["local"]}` or `{["local"] = getfenv(1)["local"]}` (depending on target).

tomblind:
I don't think it makes sense to declare a variable `local`, since that wouldn't be valid in lua. I think we should throw an error for that and let users explicitly access it the same way they would need to in lua (`_G["local"]`)

src/LuaTransformer.ts
```diff
@@ -141,7 +141,7 @@ export class LuaTransformer {
     public transformStatement(node: ts.Statement): StatementVisitResult {
         // Ignore declarations
         if (node.modifiers && node.modifiers.some(modifier => modifier.kind === ts.SyntaxKind.DeclareKeyword)) {
-            return undefined;
+            return this.transformDeclareStatement(node);
```
ark120202: These types can be declared in other file that isn't going to be transformed, so it should be handled when it's used

src/LuaTransformer.ts
```diff
@@ -147,7 +141,7 @@ export class LuaTransformer {
     public transformStatement(node: ts.Statement): StatementVisitResult {
         // Ignore declarations
```
Perryvw: We no longer just 'ignore' declarations, so this comment should probably be removed.

src/LuaTransformer.ts
```diff
@@ -209,6 +203,34 @@ export class LuaTransformer {
         }
     }
 
+    public transformDeclareStatement(node: ts.Statement): StatementVisitResult {
```
ark120202: IMO we shouldn't do this anymore. First, it would be quite a surprising behavior if some code would work in .d.ts  files (since we don't transform them) and won't in regular. Second, it's pretty much valid to declare it, because it also may be used to augment _G type.

tomblind: I see your point about the inconsistency being confusing, but I don't see how that can be a valid declaration. Such variables could be declared as a property of `_G`, but this error wouldn't be triggered by that.

Also I think it would be a good idea to leave this function in (just returning undefined) as it makes sense as an api method.

ark120202: The best type for `_G` is `typeof globalThis`, which includes all globally declared symbols

PR #588, opened 2019-05-20
Enable prettier on all files

ark120202:


src/LuaPrinter.ts
```diff
@@ -163,8 +161,8 @@ export class LuaPrinter {
 
         return originalPos !== undefined && originalPos.line !== undefined && originalPos.column !== undefined
             ? new SourceNode(originalPos.line + 1, originalPos.column, this.sourceFile, chunks)
-            // tslint:disable-next-line:no-null-keyword
-            : new SourceNode(null, null, this.sourceFile, chunks);
+            : // tslint:disable-next-line:no-null-keyword
+              new SourceNode(null, null, this.sourceFile, chunks);
```
tomblind: This is kind of weird. Does the tslint:disable still work here?

src/LuaAST.ts
```diff
@@ -41,58 +41,73 @@ export enum SyntaxKind {
...
     // Relational Ops
     EqualityOperator,
     InequalityOperator,
     LessThanOperator,
     LessEqualOperator,
-    GreaterThanOperator,   // Syntax Sugar `x > y` <=> `not (y <= x)`
-                           // but we should probably use them to make the output code more readable
-    GreaterEqualOperator,  // Syntax Sugar `x >= y` <=> `not (y < x)`
+    GreaterThanOperator, // Syntax Sugar `x > y` <=> `not (y <= x)`
+    // but we should probably use them to make the output code more readable
```
tomblind: We should probably avoid aligning comment like this and put them on their own lines.

src/TSHelper.ts
```diff
@@ -651,7 +674,14 @@ export class TSHelper {
...
 
     public static fixInvalidLuaIdentifier(name: string): string {
-        return name.replace(/[^a-zA-Z0-9_]/g, c => `_${c.charCodeAt(0).toString(16).toUpperCase()}`);
+        return name.replace(
+            /[^a-zA-Z0-9_]/g,
+            c =>
+                `_${c
+                    .charCodeAt(0)
+                    .toString(16)
+                    .toUpperCase()}`
```
tomblind: Why was this broken up like this?

src/LuaPrinter.ts
tomblind: Looks like it does still work

ark120202: It indeed looks weird, but kind of makes sense, since comment is attached to `new SourceNode(...)` node. Instead we may use `tslint:disable-line`.
Issue: https://github.com/prettier/prettier/issues/2076.

src/TSHelper.ts
ark120202: Prettier breaks call chains with 3 or more elements. Relevant issue: https://github.com/prettier/prettier/issues/3107.

src/LuaPrinter.ts
tomblind: Personally, I'd prefer to stick with explicitly disabling specific rules instead of the whole line (the line could break a different rule still). The odd formatting isn't a huge deal and could be fixed in the future.

ark120202: It's possible to disable specific rule:
```ts
return originalPos !== undefined && originalPos.line !== undefined && originalPos.column !== undefined
    ? new SourceNode(originalPos.line + 1, originalPos.column, this.sourceFile, chunks)
    : new SourceNode(null, null, this.sourceFile, chunks); // tslint:disable-line:no-null-keyword
```

tomblind: Nice - didn't know that

PR #589, opened 2019-05-21
fixed default binding pattern parameters

### LINKED ISSUE #530: Default function parameter doesn't work with destructuring
```ts
function test({ foo }: { foo?: any } = {}) {}
```

```
TypeError: Cannot read property 'length' of undefined
    at LuaTransformer.getIdentifierText (typescript-to-lua/src/LuaTransformer.ts:4340:25)
    at LuaTransformer.transformIdentifier (typescript-to-lua/src/LuaTransformer.ts:4360:34)
    at LuaTransformer.transformParameterDefaultValueDeclaration (typescript-to-lua/src/LuaTransformer.ts:1290:36)
    at parameters.filter.map.declaration (typescript-to-lua/src/LuaTransformer.ts:1260:38)
    at Array.map (<anonymous>)
    at LuaTransformer.transformFunctionBody (typescript-to-lua/src/LuaTransformer.ts:1260:14)
    at LuaTransformer.transformFunctionDeclaration (typescript-to-lua/src/LuaTransformer.ts:1721:20)
    at LuaTransformer.transformStatement (typescript-to-lua/src/LuaTransformer.ts:168:29)
    at statements.forEach.statement (typescript-to-lua/src/LuaTransformer.ts:215:74)
    at Array.forEach (<anonymous>)
```

tomblind:
fixes #530

### DIFF
--- a/src/LuaTransformer.ts
+++ b/src/LuaTransformer.ts
@@ -1355,10 +1355,12 @@ export class LuaTransformer {
 
         // Add default parameters
         const defaultValueDeclarations = parameters
-            .filter(declaration => declaration.initializer !== undefined)
-            .map(declaration => this.transformParameterDefaultValueDeclaration(declaration));
-
-        headerStatements.push(...defaultValueDeclarations);
+            .filter(declaration => declaration.initializer !== undefined && ts.isIdentifier(declaration.name))
+            .map(declaration => this.transformParameterDefaultValueDeclaration(
+                this.transformIdentifier(declaration.name as ts.Identifier),
+                declaration.initializer
+            )
+        );
 
         // Push spread operator here
         if (spreadIdentifier) {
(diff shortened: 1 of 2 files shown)


src/LuaTransformer.ts
```diff
@@ -1369,15 +1371,20 @@ export class LuaTransformer {
...
-        parameters.forEach(binding => {
+        parameters.forEach((binding, i) => {
             if (ts.isObjectBindingPattern(binding.name) || ts.isArrayBindingPattern(binding.name)) {
                 const identifier = tstl.createIdentifier(`____TS_bindingPattern${identifierIndex++}`);
                 bindingPatternDeclarations.push(...this.statementVisitResultToArray(
                     this.transformBindingPattern(binding.name, identifier)
                 ));
+                if (binding.initializer !== undefined) {
+                    const declaration = this.transformParameterDefaultValueDeclaration(identifier, binding.initializer);
+                    defaultValueDeclarations.splice(i, 0, declaration);
```
Perryvw: I think this is is unnecessarily complex, it seems to me this is is a hack so you can keep the `map` at 1359. In this case it might be simpler to just replace that map with a `for ... of` or `forEach` and push the declaration plus an additional optional push for the value to some predefined list.

tomblind: I won't deny it was a bit of a hack. I've refactored this into something a bit more sensible.

PR #598, opened 2019-05-25
Added support for importequals declaration

### LINKED ISSUE #587: Unsupported Statement kind: ImportEqualsDeclaration
I get this error when I try to pull a Lua function that is found somewhere down a global table hierarchy, e.g. GlobalLuaTable.Graphics.draw, into the file scope:

```typescript
// Somewhere in .d.ts
declare namespace GlobalLuaTable {
  namespace Graphics {
    function draw(): void;
 }
}

// Then at the start of some .ts:
import draw = GlobalLuaTable.Graphics.draw;
```
This leads to the error `Unsupported Statement kind: ImportEqualsDeclaration`.

Of course, I might as well do 
```typescript
const draw = GlobalLuaTable.Graphics.draw;
```
but my IDE likes to insert these `import = ...` style imports instead. 

Any chance these will be supported? 

Thank you for a great project!

Perryvw:
Closes #587


Perryvw:
I feel like there has to be a better way to do transformers.ts:503, but I don't know how.

src/LuaTransformer.ts
```diff
@@ -471,6 +473,39 @@ export class LuaTransformer {
...
+
+    public transformEntityName(entityName: ts.EntityName): ExpressionVisitResult {
+        return ts.isQualifiedName(entityName)
+            ? this.transformQualifiedName(entityName)
+            : this.transformIdentifier(entityName);
+    }
+
+    public transformQualifiedName(qualifiedName: ts.QualifiedName): ExpressionVisitResult {
+        const right = this.transformIdentifier(qualifiedName.right);
+        const rightLiteral = tstl.createStringLiteral(right.text, qualifiedName.right);
```
ark120202: > I feel like there has to be a better way to do transformers.ts:503, but I don't know how.

`getIdentifierText`?

src/LuaTransformer.ts
```diff
@@ -471,6 +473,39 @@ export class LuaTransformer {
...
+    }
+
+    public transformEntityName(entityName: ts.EntityName): ExpressionVisitResult {
+        return ts.isQualifiedName(entityName)
+            ? this.transformQualifiedName(entityName)
+            : this.transformIdentifier(entityName);
+    }
+
+    public transformQualifiedName(qualifiedName: ts.QualifiedName): ExpressionVisitResult {
+        const right = this.transformIdentifier(qualifiedName.right);
```
```suggestion
        const right = this.transformIdentifierExpression(qualifiedName.right);
```

Right now there's no much difference, but still it should make more sense and also be consistent with regular variable declaration.

src/LuaTransformer.ts
```diff
@@ -471,6 +473,39 @@ export class LuaTransformer {
...
+    public transformExternalModuleReference(
+        externalModuleReference: ts.ExternalModuleReference
+    ): ExpressionVisitResult {
+        return this.transformExpression(externalModuleReference.expression);
+    }
+
+    public transformEntityName(entityName: ts.EntityName): ExpressionVisitResult {
+        return ts.isQualifiedName(entityName)
+            ? this.transformQualifiedName(entityName)
+            : this.transformIdentifier(entityName);
```
```suggestion
            : this.transformIdentifierExpression(entityName);
```

src/LuaTransformer.ts
```diff
@@ -471,6 +473,39 @@ export class LuaTransformer {
...
     }
 
+    public transformImportEqualsDeclaration(declaration: ts.ImportEqualsDeclaration): StatementVisitResult {
+        const name = this.transformIdentifier(declaration.name);
+        const expression = this.transformModuleReference(declaration.moduleReference);
+
+        return tstl.createVariableDeclarationStatement(name, expression, declaration);
+    }
+
+    public transformModuleReference(moduleReference: ts.ModuleReference): ExpressionVisitResult {
```
ark120202: I wonder does it really make sense to expose it as a public visitor?
Also, have you considered to use `transformExpression` and just add new nodes there?

ark120202:
Since imports are transformed into regular vars they are also supposed to be hoisted

src/LuaTransformer.ts
Perryvw: That would lose the renaming logic that's in transformIdentifier.

src/LuaTransformer.ts
Perryvw: I think it does, since these are just AST nodes being visited, even if we only know of one place where they are right now.

src/LuaTransformer.ts
```diff
@@ -471,6 +473,38 @@ export class LuaTransformer {
...
+            name,
+            expression,
+            declaration
+        );
+    }
+
+    public transformExternalModuleReference(
+        externalModuleReference: ts.ExternalModuleReference
+    ): ExpressionVisitResult {
+        return this.transformExpression(externalModuleReference.expression);
```
tomblind: Shouldn't this be wrapped in a `require`? Right now,
```ts
import foo = require("./bar");
```
transpiles to:
```lua
local foo = "./bar"
```

PR #609, opened 2019-05-28
Add test builder

### LINKED ISSUE #602: Simplify result comparison tests
Most of unit tests have a very simple pattern:

```ts
const result = util.transpileAndExecute(`
	return foo(${case});
`)

expect(result).toBe(foo(case))
```

Because `foo` has to be available both in transpiled string and test module, it may become quite complex (when  instead of `foo(case)` there are some complex constructs). Instead we may emit both Lua and JS, execute JS in a vm and compare results.

ark120202:
Resolves #602.


test/util.ts
```diff
@@ -203,8 +205,419 @@ export function expectToBeDefined<T>(subject: T | null | undefined): subject is
...
+        const testName = name === undefined ? version : `${name} [${version}]`;
+        test(testName, () => {
+            const builder = common();
+            builder.setOptions({ luaTarget: version });
+            specialBuilder(builder);
+        });
+    }
+}
+
+interface TranspiledJsFile {
```
Perryvw: Didn't we already have something like this in transpile.ts you can import?

test/util.ts
```diff
@@ -203,8 +205,419 @@ export function expectToBeDefined<T>(subject: T | null | undefined): subject is
...
+    }
+
+    @memoize
+    protected getJsCodeWithWrapper(): string {
+        return this.getMainJsCodeChunk() + `\n;module.exports = module.exports${this.accessor}`;
+    }
+}
+
+class ModuleTestBuilder extends AccessorTestBuilder {
+    public setExport(name: string): this {
```
Perryvw: How about `setReturnExport`? I think this is a bit confusing.

test/util.ts
```diff
@@ -203,8 +205,419 @@ export function expectToBeDefined<T>(subject: T | null | undefined): subject is
...
+export class ExecutionError extends Error {
+    public name = "ExecutionError";
+    constructor(message: string) {
+        super(message);
+    }
+}
+
+export type ExecutableTranspiledFile = tstl.TranspiledFile & { lua: string; sourceMap: string };
+export type TapCallback = (builder: TestBuilder) => void;
+export abstract class TestBuilder {
```
Perryvw: What do you think of splitting this off into its own file?

Perryvw:
I'm seeing `11 skipped, 2 todo` in the test log, those should ideally be resolved or reverted before merging.

PR #620, opened 2019-06-01
fix for parenthesis stripping

### LINKED ISSUE #619: Parenthesized expression brackets stripped when they should not
```ts
const abc = (4 - 6 as number).toLocaleString()
```
Is translated to:
```lua
local abc = 4 - 6:toLocaleString()
```

I'm not entirely sure why this happens.

[Playground link](https://typescripttolua.github.io/play.html#src=const%20abc%20%3D%20(4%20-%206%20as%20number).toLocaleString())

tomblind:
- logic moved to printer
- decision to strip is now explicit based on inner expression type to avoid surprises

fixes #619

### DIFF
--- a/test/unit/expressions.spec.ts
+++ b/test/unit/expressions.spec.ts
@@ -581,3 +581,12 @@ test.each([
     const lua = util.transpileString(code, undefined, false);
     expect(lua).toMatch(/\(.+\)/);
 });
+
+test("not operator precedence (%p)", () => {
+    const code = `
+        const a = true;
+        const b = false;
+        return !a && b;`;
+
+    expect(util.transpileAndExecute(code)).toBe(false);
+});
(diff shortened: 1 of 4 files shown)


test/unit/expressions.spec.ts
```diff
@@ -527,3 +527,57 @@ test.each([
...
+        declare function z(this: void): unknown;
+        ${expression}`;
+
+    const lua = util.transpileString(code, undefined, false);
+    expect(lua).not.toMatch(/\(.+\)/);
+});
+
+test.each([
+    "(x + 1 as any).foo;",
+    "(!x as any).foo;",
```
Perryvw: Could you add a test to check if this:
```ts
const a = true;
const b = false;
return !a && b;
```
Returns the same value in TS as in Lua? I'm a little worried about the `not` operator precedence.

PR #623, opened 2019-06-03
Simplify const enum transform

ark120202:


src/LuaTransformer.ts
```diff
@@ -3960,46 +3959,15 @@ export class LuaTransformer {
...
-                    if (member.initializer === undefined) {
-                        enumValue++;
-                    } else if (ts.isNumericLiteral(member.initializer)) {
-                        enumValue = Number(member.initializer.text) + 1;
-                    }
-                }
-
-                throw TSTLErrors.CouldNotFindEnumMember(enumDeclaration, memberName, tsOriginal);
-            }
+    private getConstEnumValue(
```
Perryvw: This method name might give the wrong idea when reading code, I think something along the lines of `tryGetConstEnumValue` or `getConstEnumValueIfExists` would better describe what it does.

PR #625, opened 2019-06-06
typeof improvements

### LINKED ISSUE #612: typeof nil should be undefined
In JS `typeof undefined` is `"undefined"`, but in Lua it is `"nil"`. It also probably should be extracted to a helper.

### LINKED ISSUE #613: Optimize typeof comparisions
`typeof foo === "string"` -> `type(foo) == "string"` (we can directly transform it when we know that we don't need to change type)

`typeof foo === "undefined`" -> `type(foo) == "nil"` (instead of wrapping left-hand part we can just change right)

tomblind:
closes #613 
fixes #612 

typeof is now transpiled into a lib function, except in cases where it is directly compared to a string literal, in which case lua's `type()` is used and the literal is changed (if needed).

### DIFF
--- a/src/LuaTransformer.ts
+++ b/src/LuaTransformer.ts
@@ -2655,6 +2655,22 @@ export class LuaTransformer {
         }
     }
 
+    protected transformTypeOfLiteralComparison(
+        typeOfExpression: ts.TypeOfExpression,
+        comparedExpression: tstl.StringLiteral,
+        operator: ts.BinaryOperator,
+        tsOriginal: ts.Node
+    ): ExpressionVisitResult {
+        if (comparedExpression.value === "object") {
+            comparedExpression.value = "table";
+        } else if (comparedExpression.value === "undefined") {
+            comparedExpression.value = "nil";
+        }
+        const innerExpression = this.transformExpression(typeOfExpression.expression);
+        const typeCall = tstl.createCallExpression(tstl.createIdentifier("type"), [innerExpression], typeOfExpression);
+        return this.transformBinaryOperation(typeCall, comparedExpression, operator, tsOriginal);
+    }
+
     public transformBinaryExpression(expression: ts.BinaryExpression): ExpressionVisitResult {
         // Check if this is an assignment token, then handle accordingly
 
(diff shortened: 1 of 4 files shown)


src/LuaTransformer.ts
```diff
@@ -2697,7 +2715,13 @@ export class LuaTransformer {
             case ts.SyntaxKind.EqualsEqualsEqualsToken:
             case ts.SyntaxKind.ExclamationEqualsToken:
             case ts.SyntaxKind.ExclamationEqualsEqualsToken:
-                return this.transformBinaryOperation(lhs, rhs, expression.operatorToken.kind, expression);
+                // Custom handling for 'typeof(foo) === "type"'
+                if (ts.isTypeOfExpression(expression.left) && tstl.isStringLiteral(rhs)) {
```
Perryvw: I don't like adding extra logic here, we could just add this to transformBinaryOperation right?

test/unit/typechecking.spec.ts
```diff
@@ -140,3 +140,52 @@ test("instanceof Symbol.hasInstance", () => {
...
+    { expression: "undefined", operator: "===", compareTo: "undefined", expectResult: "TRUE" },
+    { expression: "() => {}", operator: "===", compareTo: "function", expectResult: "TRUE" },
+    { expression: "1", operator: "===", compareTo: "number", expectResult: "TRUE" },
+    { expression: "true", operator: "===", compareTo: "boolean", expectResult: "TRUE" },
+    { expression: `"foo"`, operator: "===", compareTo: "string", expectResult: "TRUE" },
+])("typeof literal comparison (%p)", ({ expression, operator, compareTo, expectResult }) => {
+    const code = `
+        let val = ${expression};
+        if (typeof val ${operator} "${compareTo}") {
+            return "TRUE";
```
Perryvw: This looks a little odd, any reason you don't just `return typeof val ${operator} "${compareTo}"`?

src/LuaTransformer.ts
tomblind: I agree having the logic out here is not very nice. But, I need to check the original ts expression types, which are not available in `transformBinaryOperation` currently. I also need to limit this logic to comparison operators, so I would need to re-check the operator in that method.

tomblind: I've refactored things a bit to clean it up.

src/LuaTransformer.ts
```diff
@@ -2671,10 +2671,26 @@ export class LuaTransformer {
         return this.transformBinaryOperation(typeCall, comparedExpression, operator, tsOriginal);
     }
 
+    protected transformComparisonExpression(expression: ts.BinaryExpression): ExpressionVisitResult {
```
Perryvw: Did you just rename this? I could have sworn it was `transformTypeOfComparisonExpression` before (which I prefer)

src/LuaTransformer.ts
```diff
@@ -2685,28 +2701,27 @@ export class LuaTransformer {
...
             case ts.SyntaxKind.PlusToken:
             case ts.SyntaxKind.AmpersandAmpersandToken:
             case ts.SyntaxKind.BarBarToken:
             case ts.SyntaxKind.MinusToken:
             case ts.SyntaxKind.AsteriskToken:
             case ts.SyntaxKind.AsteriskAsteriskToken:
             case ts.SyntaxKind.SlashToken:
-            case ts.SyntaxKind.PercentToken:
+            case ts.SyntaxKind.PercentToken: {
+                const lhs = this.transformExpression(expression.left);
```
Perryvw: Any reason we calculate these in here at all, instead of inside `transformBinaryOperation`? Seems like a historic thing that's probably not relevant anymore?

src/LuaTransformer.ts
tomblind: No - `transformTypeOfLiteralComparison` is above this function (I didn't change the name from before)

src/LuaTransformer.ts
tomblind: `transformBinaryOperation` expects already transformed expressions, because it is sometimes given manually constructed ones (like from compound assignments)

PR #626, opened 2019-06-06
@forRange directive

tomblind:
This new directive is used to create lua numeric for loops by using an ambient function declaration which can be used in TS for...of loops.

```ts
/** @forRange **/
declare function luaForRange(start: number, limit: number, step?: number): Iterable<number>;

for (const i of luaForRange(1, 10, 2)) {
    console.log(i);
}
```
=>
```lua
for i = 1, 10, 2 do
    print(i)
end
```

An error will be thrown if the function:
- doesn't take 2-3 number arguments, or it returns something other than a number iterable/array
- doesn't declare a single control variable
- is non-ambient
- is referenced outside of a for...of loop

### DIFF
--- a/src/Decorator.ts
+++ b/src/Decorator.ts
@@ -29,6 +29,8 @@ export class Decorator {
                 return DecoratorKind.NoSelf;
             case "noselfinfile":
                 return DecoratorKind.NoSelfInFile;
+            case "forrange":
+                return DecoratorKind.ForRange;
         }
 
         return undefined;
@@ -61,4 +63,5 @@ export enum DecoratorKind {
     LuaTable = "LuaTable",
     NoSelf = "NoSelf",
     NoSelfInFile = "NoSelfInFile",
+    ForRange = "ForRange",
 }
--- a/src/LuaTransformer.ts
+++ b/src/LuaTransformer.ts
@@ -2384,19 +2384,6 @@ export class LuaTransformer {
             throw TSTLErrors.InvalidForRangeCall(statement.expression, "Expression must be a call expression.");
         }
 
-        const symbol = this.checker.getSymbolAtLocation(statement.expression.expression);
-        if (!symbol) {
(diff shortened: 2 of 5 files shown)


hazzard993:
This is looking pretty solid. I have got something picky that is kinda 🤷‍

Thought about it after LuaTables

When declaring things like this an identifier is revealed to TypeScript that doesn't actually exist. This identifier can be used almost everywhere.

```ts
console.log(luaForRange);
luaForRange.call(null, 0, 0, 0);
let array = [0, luaForRange, 1];
const call = cb => cb();
call(luaForRange);
```

If this is something that is considered a problem I'm thinking a functional solution could be to iterate over all identifiers in each source file.

Each identifier in a source file would have their parents' kinds and directives checked to make sure identifiers revealing _forRange_ functions are being used in call expressions on the right hand side of _for..of_ statements. I am not entirely sure how to iterate through every node, but it might look something like the below code:

```ts
// TSHelper
validateNode(checker: ts.Checker, node: Node) {
    if (node.kind === ts.SyntaxKind.Identifier) {
        const type = checker.getTypeAtLocation(node);
        if (this.getCustomDecorators(type).has(DecoratorKind.ForRange)) {
            // further checks for a valid ForRange use
        }
    }
    ts.forEachChild(node, this.validateNode);
}

// LuaTransformer
transformSourceFile(node: ts.SourceFile) {
    ...
    ts.forEachChild(node, tsHelper.validateNode.bind(tsHelper, this.checker));
}
```

tomblind:
While all directives have situations where they can be broken (they fall into the "use at your own risk" category), you're right that these cases can be caught easily enough. We actively try to avoid "scanning" the AST, but I was able to add a general check in `transformIdentifier` to catch all of the illegal references. This also had the side effect of preventing non-ambient declarations since those counted as an illegal reference. We might want to try doing something like this as well with `@luaTable`.

PR #628, opened 2019-06-09
Optimized Rest Parameter -> Spread Operator

tomblind:
When a rest parameter is referenced with a spread operator, the transformed code packs and unpacks it:
```ts
function foo(this: void, ...args: unknown[]) {
    console.log(...args);
}
```
=>
```lua
function foo(...)
    local args = ({...})
    print(unpack(args))
end
```

~~This PR optimizes this so that the elipsis operator is used instead.~~ It also prevents generation of the temp local if it is not needed in the function.

*Update*
Due to issues with the optimization being applied inside generated IIFE's, this PR has been updated to instead provide an explicit `@elipsisForward` directive to apply this optimization.

```ts
/** @elipsisForward */
declare function elipsisForward<A extends unknown[]>(args?: A): A;

function foo(this: void, ...args: unknown[]) {
    console.log(...elipsisForward(args));
}
```
=>
```lua
function foo(...)
    print(...)
end
```

The new directive can only be used on a function, and that function can only be used in a spread expression, and passed a rest argument.

You can also pass no argument, which is useful for accessing the file-scope elipsis:
```ts
const arg = [...elipsisForward()];
```
=>
```lua
local arg = {...}
```

### DIFF
--- a/src/Decorator.ts
+++ b/src/Decorator.ts
@@ -29,8 +29,8 @@ export class Decorator {
                 return DecoratorKind.NoSelf;
             case "noselfinfile":
                 return DecoratorKind.NoSelfInFile;
-            case "elipsisforward":
-                return DecoratorKind.ElipsisForward;
+            case "vararg":
+                return DecoratorKind.VarArg;
         }
 
         return undefined;
@@ -63,5 +63,5 @@ export enum DecoratorKind {
     LuaTable = "LuaTable",
     NoSelf = "NoSelf",
     NoSelfInFile = "NoSelfInFile",
-    ElipsisForward = "ElipsisForward",
+    VarArg = "VarArg",
 }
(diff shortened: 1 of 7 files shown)


ark120202:
This optimization shouldn't be applied if it's used in the inner function:

```ts
function foo(...a: any[]) {
    function bar() {
        print(...a);
    }
}
```

```lua
function foo(self, ...)
    local function bar(self)
        print(...)
    end
end
```

```
3: cannot use '...' outside a vararg function near '...'
```

ark120202:
It also shouldn't optimize it in try-catch

```ts
function foo(...a: any[]) {
    try {
        print(...a);
    } catch {}
}
```

and expressions that require IIFE

```ts
function foo(...a: any[]) {
    const _ = (0, print(...a));
}
```

tomblind:
closing to open as a fresh PR since it's changed so radically

PR #630, opened 2019-06-12
@varArg directive

tomblind:
When a rest parameter is referenced with a spread operator, the transformed code packs and unpacks it:
```ts
function foo(this: void, ...args: unknown[]) {
    console.log(...args);
}
```
=>
```lua
function foo(...)
    local args = ({...})
    print(unpack(args))
end
```

This PR adds a new directive `@varArg` to allow direct access to lua vararg operator `...`. This can be applied to an array-like type and any references to an identifier with that type will be directly transformed into `...`:
```ts
/** @varArg */
interface LuaVarArg<T> extends Array<T> {}

function foo(this: void, ...args: LuaVarArg<unknown>) {
    console.log(...args);
}
```
=>
```lua
function foo(...)
    print(...)
end
```

Aside from function varargs, this can be used to access the file-scope varargs:
```ts
declare const arg: LuaVarArg<string>;
console.log(...arg);
```
=>
```lua
print(...)
```

To support tuple rest arguments, the vararg type must be declared like this:
```ts
/** @varArg */
type LuaVarArg<A extends unknown[]> = A & { __luaVarArg?: never };

function foo(this: void, ...args: LuaVarArg<[string, number]>) {
    console.log(...args);
}
foo("A", 2);
```

Note that this PR also adds stripping of the rest parameter variable (`local args = ({...})`) if it is not needed.

### DIFF
--- a/src/Decorator.ts
+++ b/src/Decorator.ts
@@ -29,8 +29,8 @@ export class Decorator {
                 return DecoratorKind.NoSelf;
             case "noselfinfile":
                 return DecoratorKind.NoSelfInFile;
-            case "elipsisforward":
-                return DecoratorKind.ElipsisForward;
+            case "vararg":
+                return DecoratorKind.VarArg;
         }
 
         return undefined;
@@ -63,5 +63,5 @@ export enum DecoratorKind {
     LuaTable = "LuaTable",
     NoSelf = "NoSelf",
     NoSelfInFile = "NoSelfInFile",
-    ElipsisForward = "ElipsisForward",
+    VarArg = "VarArg",
 }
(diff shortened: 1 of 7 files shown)


ark120202:
I think it would be better to call it `@vararg`, without considering them separate words. That's how it's written in [Lua Reference Manual](https://www.lua.org/manual/5.3/manual.html).

tomblind:
> 
> 
> I think it would be better to call it `@vararg`, without considering them separate words. That's how it's written in [Lua Reference Manual](https://www.lua.org/manual/5.3/manual.html).

Sure. The directives aren't case-sensitive, so it's just a matter of what we put in the wiki when documenting it.

PR #631, opened 2019-06-12
globalThis to _G

### LINKED ISSUE #592: Transform 'globalThis' to '_G'
TypeScript now supports `globalThis` which is very similar to lua's `_G`. We should probably just transform it directly.

lolleko:
Closes #592


src/LuaTransformer.ts
```diff
@@ -4630,6 +4630,10 @@ export class LuaTransformer {
     }
 
     public transformIdentifier(identifier: ts.Identifier): tstl.Identifier {
+        if (identifier.escapedText === "globalThis") {
```
ark120202: It should be checked in `transformIdentifierExpression` (because for example in `foo.bar` `foo` is considered identifier expression and `bar` is just identifier). It also should check it with `isStandardLibraryType`, to allow reassigning it

src/LuaTransformer.ts
```diff
@@ -5172,7 +5176,9 @@ export class LuaTransformer {
     protected hasUnsafeSymbolName(symbol: ts.Symbol, tsOriginal?: ts.Identifier): boolean {
         const isLuaKeyword = luaKeywords.has(symbol.name);
         const isInvalidIdentifier = !tsHelper.isValidLuaIdentifier(symbol.name);
-        const isAmbient = symbol.declarations.some(d => tsHelper.isAmbient(d));
+        // TODO investigate: declarations will be undefined if "globalThis" is passed, could be a TS bug
```
ark120202: https://github.com/microsoft/TypeScript/issues/24706

src/LuaTransformer.ts
```diff
@@ -4699,6 +4691,15 @@ export class LuaTransformer {
             return this.createExportedIdentifier(identifier, exportScope);
         }
 
+        if (expression.originalKeywordKind === ts.SyntaxKind.UndefinedKeyword) {
+            return tstl.createNilLiteral();
+        }
+
+        const identifierType = this.checker.getTypeAtLocation(expression);
+        if (identifierType.symbol && identifierType.symbol.escapedName === "globalThis") {
```
ark120202: I think we shouldn't use type there and only transform actual `globalThis` identifier.
Declaring a variable is just a way to tell type system about already existing thing, so transforming it to something else would be really unexpected behavior for me.
Also not transforming it would avoid confusing things like an object property with type of globalThis working only when it's assigned to a variable.

lolleko: The idea was to allow the extension `typeof globalThis` e.g:
```
interface Test {
    foo: string;
}
declare let globalAlias: typeof globalThis & Test;
globalAlias.foo = "bar";
```
This would allow typesafe access to `_G`. Without  the ability to extend `typeof globalThis` you will only be able to access global variables that are declared in Lua by casting to any `(print(<any>globalThis.myLuaGlobal))`,

But you are right this would conflict with stuff like object properties that are assigned to globalThis.
I reverted the changes in the latest commit.

PR #633, opened 2019-06-18
More source map stuff

tomblind:
- function `end` is now mapped to the function declaration, since lua debuggers step on that line when returning
- class `new` and generated constructors now map to the class declaration
- added support for mapped names for translating `self`->`this` and variables which were lua identifiers or had invalid characters

### DIFF
--- a/src/LuaTransformer.ts
+++ b/src/LuaTransformer.ts
@@ -1229,6 +1229,8 @@ export class LuaTransformer {
 
         const block: tstl.Block = tstl.createBlock(bodyWithFieldInitializers);
 
+        const constructorWasGenerated = statement.pos === -1;
+
         const result = tstl.createAssignmentStatement(
             this.createConstructorName(className),
             tstl.createFunctionExpression(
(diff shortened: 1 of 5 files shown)


src/LuaTransformer.ts
```diff
@@ -1229,7 +1238,7 @@ export class LuaTransformer {
                 restParamName,
                 tstl.FunctionExpressionFlags.Declaration
             ),
-            statement
+            statement.pos >= 0 ? statement : classDeclaration
```
Perryvw: This seems strange, either move this to a variable with descriptive name or add a comment with explanation.

src/LuaTransformer.ts
```diff
@@ -1229,7 +1238,7 @@ export class LuaTransformer {
                 restParamName,
                 tstl.FunctionExpressionFlags.Declaration
             ),
-            statement
+            statement.pos >= 0 ? statement : classDeclaration // Map to class declaration if constructor was generated
```
Perryvw: Why would pos >= 0 indicate it is generated? Couldn't you technically put a constructor at the very end of your class definition?

tomblind: Sorry, I guess this still just wasn't clear. I've reworked it a tad to hopefully be more understandable.

PR #636, opened 2019-06-19
Fix noResolution for named imports

### LINKED ISSUE #635: noResolution directive not detected on named imports
```ts
/** @noResolution */
declare module "module" {
    export const x: number;
}
```

```ts
import { x } from "module";
x;
```

The `@noResolution` directive in this instance is not picked up so TypeScriptToLua will attempt to resolve the path to `"module"` which will change based off the file using this import.

hazzard993:
Closes #635

`@noResolution` should still only be applicable to ambient modules.

If a module is merged with another and one has a `@noResolution`, no path resolution will be performed.

```ts
/** @noResolution */
declare module "module" {}
declare module "module" {
    export const x: number;
}
// ✔ will stay "module"
import { x } from "module";
```

### DIFF
--- a/src/LuaTransformer.ts
+++ b/src/LuaTransformer.ts
@@ -357,11 +357,28 @@ export class LuaTransformer {
             scope.importStatements = [];
         }
 
+        let shouldResolve = true;
+        const moduleOwnerSymbol = this.checker.getSymbolAtLocation(statement.moduleSpecifier);
+        if (moduleOwnerSymbol) {
+            for (const declaration of moduleOwnerSymbol.declarations) {
+                if (tsHelper.isNonNamespaceModuleDeclaration(declaration)) {
+                    const type = this.checker.getTypeAtLocation(declaration);
+                    if (type) {
+                        const decorators = tsHelper.getCustomDecorators(type, this.checker);
+                        if (decorators.has(DecoratorKind.NoResolution)) {
+                            shouldResolve = false;
+                            break;
+                        }
+                    }
+                }
+            }
+        }
+
         const moduleSpecifier = statement.moduleSpecifier as ts.StringLiteral;
         const importPath = moduleSpecifier.text.replace(new RegExp('"', "g"), "");
+        const requireCall = this.createModuleRequire(statement.moduleSpecifier as ts.StringLiteral, shouldResolve);
 
         if (!statement.importClause) {
-            const requireCall = this.createModuleRequire(statement.moduleSpecifier as ts.StringLiteral);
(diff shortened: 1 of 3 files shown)


src/LuaTransformer.ts
```diff
@@ -376,8 +376,26 @@ export class LuaTransformer {
             throw TSTLErrors.UnsupportedImportType(statement.importClause);
         }
 
-        const type = this.checker.getTypeAtLocation(imports);
-        const shouldResolve = !tsHelper.getCustomDecorators(type, this.checker).has(DecoratorKind.NoResolution);
+        let shouldResolve = true;
+        if (ts.isNamedImports(imports)) {
+            for (const importSpecifier of imports.elements) {
```
ark120202: I don't think we can rely on elements there. First, what about `import {} from "fake"`? Second, I'm not sure if that would work with re-exporting.
We also have to consider a case when we have no elements to check: `import "fake"`.
I think we should check out the actual type of module, not types of imported elements. I'm not really sure how to do it, but editors that use TSLS seem to have resolved module path on hover, so type checker should have it

ark120202:
Looks like we can use `this.checker.getSymbolAtLocation(statement.moduleSpecifier)` to get it in all cases. It's a symbol, so instead of `getCustomDecorators` it should use `collectCustomDecorators` to get decorators.

src/LuaTransformer.ts
hazzard993: Thanks for the hand! I can't think of a way to test `import {} from "fake"` but I've got `import "fake"` validated by two new tests.

Also using `this.checker.getSymbolAtLocation(statement.moduleSpecifier)` looks much more reliable

hazzard993:
I've removed the `isNonNamespaceModuleDeclaration` method as the more I think about it the more useless it seems to be.

This directive can technically be used on a `namespace` and a `module`. General rule is if an `import` statement is importing something that comes from a `namespace` or `module` with the `@noResolution` directive, no resolution attempts will be made.

src/LuaTransformer.ts
```diff
@@ -357,11 +357,21 @@ export class LuaTransformer {
             scope.importStatements = [];
         }
 
+        let shouldResolve = true;
+        const moduleOwnerSymbol = this.checker.getSymbolAtLocation(statement.moduleSpecifier);
+        if (moduleOwnerSymbol) {
+            const decMap = new Map<DecoratorKind, Decorator>();
```
Perryvw: Don't abbreviate variable names please.

PR #638, opened 2019-06-20
Tagged template literals support

### LINKED ISSUE #611: Feature request: Tagged template literals
Tagged template literals: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#Tagged_templates

We would translate:
```ts
tag`myLiteral ${argument1}, ${argument2}`;
```

To something like
```ts
tag(["myLiteral ", ", "], argument1, argument2);
```

hazzard993:
Closes #611

[Raw strings](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#Raw_strings) are usable too.

Equivalent output code does look hard to read.

```ts
function func(strings: TemplateStringsArray, ...expressions: any[]) {}

func`\u00A9 ${4}`;
```

```lua
function func(self, strings, ...)
end
func(_G, {
    "© ",
    "",
    raw = {
        "\\u00A9 ",
        "",
    },
}, 4)

```

### DIFF
--- a/src/LuaTransformer.ts
+++ b/src/LuaTransformer.ts
@@ -4640,36 +4640,53 @@ export class LuaTransformer {
         const strings: ts.StringLiteral[] = [];
         const rawStrings: string[] = [];
         const expressions: ts.Expression[] = [];
+
         if (ts.isTemplateExpression(expression.template)) {
             // Expressions are in the string.
             strings.push(ts.createStringLiteral(expression.template.head.text));
-            rawStrings.push(expression.template.head.getText());
+            // If there is a head, it beings with "`" and ends with "${"
+            rawStrings.push(
+                expression.template.head
+                    .getText()
+                    .replace(/\$\{/, "")
+                    .replace(/^`/, "")
+            );
             strings.push(...expression.template.templateSpans.map(span => ts.createStringLiteral(span.literal.text)));
-            rawStrings.push(...expression.template.templateSpans.map(span => span.literal.getText()));
+            // Following spans end with "}"
+            rawStrings.push(
+                ...expression.template.templateSpans.map(span => {
+                    if (span.literal.kind === ts.SyntaxKind.TemplateMiddle) {
+                        // The middle sections end with "${"
+                        return span.literal
+                            .getText()
+                            .replace(/^\}/, "")
+                            .replace(/\$\{$/, "");
+                    } else {
+                        // The tail ends with "`"
+                        return span.literal
+                            .getText()
+                            .replace(/^\}/, "")
+                            .replace(/`$/, "");
+                    }
+                })
+            );
             expressions.push(...expression.template.templateSpans.map(span => span.expression));
-        } else if (ts.isNoSubstitutionTemplateLiteral(expression.template)) {
+        } else {
             // No expressions are in the string.
             strings.push(ts.createStringLiteral(expression.template.text));
-            rawStrings.push(expression.template.getText());
-        } else {
-            throw TSTLErrors.UnsupportedKind("TaggedTemplateExpression", expression.kind, expression);
+            rawStrings.push(expression.template.getText().replace(/(^`|`$)/g, ""));
         }
 
-        // Preserve some special characters
-        const preservedRawStrings = rawStrings.map(rawString => {
-            const preservedString = rawString
-                .replace(/`/g, "")
-                .replace(/\$/g, "")
-                .replace(/\{/g, "")
-                .replace(/\}/g, "")
-                .replace(/\\/g, "\\\\");
-            return tstl.createStringLiteral(preservedString);
+        // Preserve "\"
+        const preparedRawStrings = rawStrings.map(rawString => {
+            const preparedString = rawString.replace(/\\/g, "\\\\");
+            return tstl.createStringLiteral(preparedString);
         });
 
         // Construct call expression
         const stringArray = ts.createArrayLiteral(strings);
         const rawStringArray = tstl.createTableExpression(
-            preservedRawStrings.map(rawString => tstl.createTableFieldExpression(rawString))
+            preparedRawStrings.map(rawString => tstl.createTableFieldExpression(rawString))
         );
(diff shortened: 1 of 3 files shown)


src/LuaTransformer.ts
```diff
@@ -4634,6 +4636,76 @@ export class LuaTransformer {
...
+            throw TSTLErrors.UnsupportedKind("TaggedTemplateExpression", expression.kind, expression);
+        }
+
+        // Preserve some special characters
+        const preservedRawStrings = rawStrings.map(rawString => {
+            const preservedString = rawString
+                .replace(/`/g, "")
+                .replace(/\$/g, "")
+                .replace(/\{/g, "")
+                .replace(/\}/g, "")
```
tomblind: I'm confused about why these characters are being stripped. Shouldn't the raw version of `` `foo$` `` be `"foo$"`, and `` `foo\${1}` `` be `"foo\\${1}"`?

tomblind:
I did some more tests:

Works:
- `TemplateStringsArray`
- `readonly string[]`
- `ReadonlyArray<string>`
- `any`
- `unknown`

Doesn't work:
- `string[]`
- `any[]`

So it seems `getResolvedSignature` fails if the type is a non-readonly array type.

Perhaps we should just require the type of that parameter to be `TemplateStringsArray` to keep things simple?

ark120202:
Turns out `getResolvedSignature` resolves the correct signature there when called from a non-diagnostic-producing type checker. It is caused by a slight difference between them: https://github.com/microsoft/TypeScript/pull/28584#issuecomment-445942729. This difference was removed in https://github.com/microsoft/TypeScript/pull/28564, but it doesn't seem like it's going to get merged any time soon. For now I don't think there's need to do anything - user would get a semantic error in this case anyway.

tomblind:
I think we should do something since the result right now is bad lua being silently generated.

src/LuaTransformer.ts
```diff
@@ -4634,6 +4636,93 @@ export class LuaTransformer {
...
+        const strings: ts.StringLiteral[] = [];
+        const rawStrings: string[] = [];
+        const expressions: ts.Expression[] = [];
+
+        if (ts.isTemplateExpression(expression.template)) {
+            // Expressions are in the string.
+            strings.push(ts.createStringLiteral(expression.template.head.text));
+            // If there is a head, it beings with "`" and ends with "${"
+            rawStrings.push(
+                expression.template.head
```
ark120202: Maybe it's possible to simplify it a bit? TypeScript version doesn't do any direct replacements: https://github.com/microsoft/TypeScript/blob/37b20f6afd3eee694c26ca39c586c42ec97be98e/src/compiler/transformers/es2015.ts#L3996-L4014

src/LuaTransformer.ts
hazzard993: Changed the replacement code and added some more tests for these cases

src/LuaTransformer.ts
hazzard993: Thanks for the link, I've created `tsHelper.getRawLiteral` to neaten up the method some more

hazzard993:
Is there still anything weird to fix up with the `getResolvedSignature` issue?

src/LuaTransformer.ts
```diff
@@ -4646,6 +4648,60 @@ export class LuaTransformer {
...
+            tsHelper.getDeclarationContextType(signatureDeclaration, this.checker) !== ContextType.Void;
+        const context = useSelfParameter ? (this.isStrict ? ts.createNull() : ts.createIdentifier("_G")) : undefined;
+
+        // Argument evaluation.
+        const parameters = this.transformArguments([stringArray, ...expressions], signature, context);
+
+        // Assign raw strings to the "raw" property of the string table.
+        const tableParameterIndex = useSelfParameter ? 1 : 0;
+        const tableLiteralExpression =
+            parameters.length > tableParameterIndex &&
```
ark120202: Is there a case when it's not true?

hazzard993: Yep, when there is no self parameter the table containing all the strings will be the first (0th) argument

tomblind:
`getResolvedSignature` should be fine - turns out an error is thrown by TS if the wrong type is attempted to be used, and I just wasn't seeing it when I was testing.

src/LuaTransformer.ts
hazzard993: This unnecessary check has now been removed

test/unit/taggedTemplateLiterals.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+        joinAllResult: "hello { brackets }",
+        joinRawResult: "hello {  }",
+    },
+    {
+        callExpression: "func`hello \\``",
+        joinAllResult: "hello `",
+        joinRawResult: "hello \\`",
+    },
+    {
+        callExpression: "obj.func`hello ${'propertyAccessExpression'}`",
```
ark120202: Looks like it always calls functions with global context:

```ts
const obj = {
    func(strings: TemplateStringsArray, ...expressions: any[]) {
        console.log(this === obj); // `true` in JS, `false` in Lua
    },
};

obj.func`hello`;
```

```lua
local obj = {func = function(self, strings, ...)
    print(self == obj)
end}
obj.func(_G, { -- should be `obj`
    "hello",
    raw = {"hello"},
})
```

Also there is a case where `obj` isn't pure:

```ts
getObj()["func"]`hello`;
```

In call expressions it's handled like that:

```lua
(function()
    local ____TS_self = getObj(_G)
    return ____TS_self.func(____TS_self, {
        "hello",
        raw = {"hello"},
    })
end)()
```

hazzard993: Good find! Took a bit to understand what to do here, I've added `transformContextualCallExpression` in an attempt to preserve this left hand side behaviour to call-like expressions which includes these template literals.

PR #639, opened 2019-06-20
updated call & table expression formatting

tomblind:
Arguments/fields will now only be split into multiple lines if one or more has evaluation side-effects that would be beneficial for a debugger to step on.

Examples:
----
```ts
func("foo", "bar");
```
```lua
func("foo", "bar")
```
----
```ts
func(func("foo"), obj.method());
```
```lua
func(
    func("foo"),
    obj:method()
)
```
----
```ts
func({foo: "bar"});
```
```lua
func({foo = "bar"})
```
----
```ts
func("foo", () => "bar");
```
```lua
func(
    "foo",
    function() return "bar" end
)
```
---
```ts
const arr = [1, 2, 4 - 1];
```
```lua
local arr = {1, 2, 4 - 1}
```
----
```ts
const arr = [func(), obj.method()];
```
```lua
local arr = {
    func(),
    obj:method()
}
```
----
```ts
const tbl = {foo: "foo"};
```
```lua
local tbl = {foo = "foo"}
```
----
```ts
const tbl = {foo: func("foo"), bar: obj.method({foo: "foo", bar: () => { console.log("bar"); } })};
```
```lua
local tbl = {
    foo = func("foo"),
    bar = obj:method(
        {
            foo = "foo",
            bar = function()
                print("bar")
            end
        }
    )
}
```

### DIFF
--- a/src/LuaPrinter.ts
+++ b/src/LuaPrinter.ts
@@ -709,47 +709,10 @@ export class LuaPrinter {
         return result;
     }
 
-    protected isSimpleExpression(expression: tstl.Expression): boolean {
-        switch (expression.kind) {
-            case tstl.SyntaxKind.CallExpression:
-            case tstl.SyntaxKind.MethodCallExpression:
-            case tstl.SyntaxKind.FunctionExpression:
-                return false;
-
-            case tstl.SyntaxKind.TableExpression:
-                const tableExpression = expression as tstl.TableExpression;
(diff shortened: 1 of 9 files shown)


src/LuaPrinter.ts
```diff
@@ -713,6 +707,62 @@ export class LuaPrinter {
         return result;
     }
 
+    protected isSimpleExpression(expression: tstl.Expression): boolean {
```
Perryvw: Is this maybe a tsHelper function?

PR #642, opened 2019-06-25
Handling return from try and catch

### LINKED ISSUE #252: Returning inside try block fails
```ts
function foo() {
    try {
        return "foo";
    } finally {
        return "bar";
    }
}
print(foo());
```
```lua
function foo()
    xpcall(function()
        return "foo"
end,
function(e)
end)
    return "bar"
end
print(foo()); --Prints "bar"
```
The return statement only returns out of the function passed to xpcall. I'm not totally sure what a good way to deal with this would be.

tomblind:
fixes #252 

To support proper returning from `try`/`catch` (while still executing `finally`), a `true` literal is prepended to returned values. This is checked after `finally` is executed and the return value is forwarded on. Note that the `catch` block will always be wrapped in an IIFE in this case.
```ts
function foobar() {
    try {
        return mayThrow();
    } catch (err) {
        return "error thrown: " + err;
    } finally {
        console.log("finally");
    }
}
```
=>
```lua
function foobar(self)
    do
        local ____TS_try, err, ____TS_returnValue = pcall(function()
            return true, mayThrow()
        end)
        if not ____TS_try then
            err, ____TS_returnValue = (function()
                return true, "error thrown: " .. tostring(err)
            end)()
        end
        do
            print("finally")
        end
        if err then
            return ____TS_returnValue
        end
    end
end
```

In the case of `@tupleReturn` functions, the return value will be wrapped and later unpacked when forwarded.
```ts
/** @tupleReturn */
function foobar() {
    try {
        return [mayThrow(), mayThrow()];
    } catch (err) {
        return ["error thrown", err];
    } finally {
        console.log("finally");
    }
}
```
=>
```lua
function foobar(self)
    do
        local ____TS_try, err, ____TS_returnValue = pcall(function()
            return true, ({
                mayThrow(),
                mayThrow(),
            })
        end)
        if not ____TS_try then
            err, ____TS_returnValue = (function()
                return true, ({
                    "error thrown",
                    err,
                })
            end)()
        end
        do
            print("finally")
        end
        if err then
            return unpack(____TS_returnValue)
        end
    end
end
```

### DIFF
--- a/src/LuaTransformer.ts
+++ b/src/LuaTransformer.ts
@@ -17,6 +17,8 @@ export enum ScopeType {
     Loop = 0x8,
     Conditional = 0x10,
     Block = 0x20,
+    Try = 0x40,
+    Catch = 0x80,
 }
 
 interface SymbolInfo {
@@ -37,6 +39,7 @@ interface Scope {
     functionDefinitions?: Map<tstl.SymbolId, FunctionDefinitionInfo>;
     importStatements?: tstl.Statement[];
     loopContinued?: boolean;
+    functionReturned?: boolean;
 }
 
 export interface EmitResolver {
@@ -2111,23 +2114,33 @@ export class LuaTransformer {
     }
 
     public transformReturnStatement(statement: ts.ReturnStatement): StatementVisitResult {
+        // Bubble up explicit return flag and check if we're inside a try/catch block
+        let insideTryCatch = false;
+        for (let i = this.scopeStack.length - 1; i >= 0; --i) {
+            const scope = this.scopeStack[i];
+            scope.functionReturned = true;
+
+            if (scope.type === ScopeType.Function) {
+                break;
+            }
+
+            insideTryCatch = insideTryCatch || scope.type === ScopeType.Try || scope.type === ScopeType.Catch;
+        }
+
+        let results: tstl.Expression[];
+
         if (statement.expression) {
+            const expressionType = this.checker.getTypeAtLocation(statement.expression);
             const returnType = tsHelper.getContainingFunctionReturnType(statement, this.checker);
             if (returnType) {
-                const expressionType = this.checker.getTypeAtLocation(statement.expression);
                 this.validateFunctionAssignment(statement, expressionType, returnType);
             }
             if (tsHelper.isInTupleReturnFunction(statement, this.checker)) {
                 // Parent function is a TupleReturn function
                 if (ts.isArrayLiteralExpression(statement.expression)) {
                     // If return expression is an array literal, leave out brackets.
-                    return tstl.createReturnStatement(
-                        statement.expression.elements.map(elem => this.transformExpression(elem))
-                    );
-                }
-
-                const expressionType = this.checker.getTypeAtLocation(statement.expression);
-                if (
+                    results = statement.expression.elements.map(elem => this.transformExpression(elem));
+                } else if (
                     !tsHelper.isTupleReturnCall(statement.expression, this.checker) &&
                     tsHelper.isArrayType(expressionType, this.checker, this.program)
                 ) {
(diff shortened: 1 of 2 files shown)


src/LuaTransformer.ts
```diff
@@ -2571,39 +2597,101 @@ export class LuaTransformer {
         }
     }
 
+    protected transformScopeBlock(block: ts.Block, scopeType: ScopeType): [tstl.Block, Scope] {
```
lolleko: since this functions is only used for try/catch it might be better to name it accordingly, transformScopeBlock sounds a bit generic. Unless there are plans to use this function in more places in the future.

test/unit/error.spec.ts
```diff
@@ -62,3 +62,194 @@ test("re-throw (no catch var)", () => {
...
+                return evaluate(e);
+            } finally {
+                return "finally";
+            }
+        }
+        return foobar() + " " + x;
+    `;
+    expect(util.transpileAndExecute(code)).toBe("finally evaluated");
+});
+
```
lolleko: tuple return from finally?

src/LuaTransformer.ts
tomblind: It could be used for other purposes at well. It's really just an alternate form of transformBlock which returns the scope as well. I could just modify that function and use it everywhere.

src/LuaTransformer.ts
```diff
@@ -2111,23 +2114,33 @@ export class LuaTransformer {
     }
 
     public transformReturnStatement(statement: ts.ReturnStatement): StatementVisitResult {
+        // Bubble up explicit return flag and check if we're inside a try/catch block
+        let insideTryCatch = false;
+        for (let i = this.scopeStack.length - 1; i >= 0; --i) {
```
Perryvw: Can insideTryCatch be set to false after being true again? Otherwise you could also add `!insideTryCatch` to the loop condition here.

src/LuaTransformer.ts
```diff
@@ -2571,39 +2597,101 @@ export class LuaTransformer {
...
         }
 
-        if (statement.finallyBlock) {
-            result.push(tstl.createDoStatement(this.transformBlock(statement.finallyBlock).statements));
+        if (statement.finallyBlock && statement.finallyBlock.statements.length > 0) {
+            result.push(...this.statementVisitResultToArray(this.transformBlockAsDoStatement(statement.finallyBlock)));
+        }
+
+        if (returnCondition && returnedIdentifier) {
+            // if [____TS_try and] ____TS_returned then return ____TS_returnValue end
```
Perryvw: What's up with [____TS_try and], looks a little confusing to me. I the brackets have a specific meaning you can maybe explain it better, or just remove the brackets if not.

src/LuaTransformer.ts
```diff
@@ -2571,39 +2597,101 @@ export class LuaTransformer {
...
-            result.push(tstl.createIfStatement(notTryResult, this.transformBlock(statement.catchClause.block)));
+            result.push(tstl.createIfStatement(notTryCondition, catchBlock));
+        } else if (tryScope.functionReturned) {
+            // try with return, but no catch
+            returnedIdentifier = tstl.createIdentifier("____TS_returned");
+            const returnedVariables = [tryResultIdentfier, returnedIdentifier, returnValueIdentifier];
+            result.push(tstl.createVariableDeclarationStatement(returnedVariables, tryCall));
+
+            // change return condition to '____TS_try and ____TS_returned'
+            returnCondition = tstl.createBinaryExpression(
```
Perryvw: Maybe I'm looking at the wrong spot or misunderstanding, but how does this work when the try returns nil and that is loaded into ____TS_try?

test/unit/error.spec.ts
```diff
@@ -62,3 +62,239 @@ test("re-throw (no catch var)", () => {
     const result = util.transpileAndExecute(source);
     expect(result).toBe("z");
 });
+
+test("return from try", () => {
+    const code = `
+        function foobar() {
+            try {
+                return "foobar";
```
Perryvw: Can we get this test duplicated with a void return?

src/LuaTransformer.ts
tomblind: the brackets mean optionally: the `___TS_try and` is only required when there is no `catch`

src/LuaTransformer.ts
tomblind: If the loop exits on the first `try`/`catch` it hits, the `functionReturned` flag won't bubble up to nested try/catch statements.

src/LuaTransformer.ts
tomblind: The `nil` return value would be loaded in `____TS_returned` in that case.

test/unit/error.spec.ts
tomblind: That's what's being tested below in `return nil from try`, or am I misunderstanding what you're asking here?

Perryvw: ah yes looked over that one.

src/LuaTransformer.ts
Perryvw: But `true and nil` would mean the condition is false?

tomblind: Sorry, what I said before was incorrect.
- `____TS_try` will contain the success of the `pcall`
- `____TS_returned` will contain a boolean stating whether the clause returned or not
- `____TS_returnValue` would contain the `nil` that was actually returned.

PR #643, opened 2019-06-26
Omitted expression support

### LINKED ISSUE #603: Unsupported OmittedExpression in array
```ts
let x = [0, , 2];
```

`src/main.ts:10:12 - error TSTL0: Unsupported expression kind: OmittedExpression`

hazzard993:
Closes #603

Allows OmittedExpressions to be used in an array.

```ts
const array = [0, , 2];
```

```lua
local array = {0, nil, 2}
```

It is important to note that Lua will stop iterating through an array when it reaches `nil`.

These OmittedExpressions can also be used in array binding assignment statements:

```ts
let a, c;
[a, , c] = [0, 1, 2];
```

### DIFF
--- a/src/LuaTransformer.ts
+++ b/src/LuaTransformer.ts
@@ -2909,14 +2909,9 @@ export class LuaTransformer {
 
         if (ts.isArrayLiteralExpression(expression.left)) {
             // Destructuring assignment
-            const omittedExpressionAssignmentIdentifier = tstl.createAnonymousIdentifier();
             const left =
                 expression.left.elements.length > 0
-                    ? expression.left.elements.map(e =>
-                          ts.isOmittedExpression(e)
-                              ? omittedExpressionAssignmentIdentifier
-                              : this.transformExpression(e)
-                      )
+                    ? expression.left.elements.map(e => this.transformArrayBindingExpression(e))
                     : [tstl.createAnonymousIdentifier(expression.left)];
             let right: tstl.Expression[];
             if (ts.isArrayLiteralExpression(expression.right)) {
@@ -3474,7 +3469,8 @@ export class LuaTransformer {
     }
 
     public transformOmittedExpression(node: ts.OmittedExpression): ExpressionVisitResult {
-        return tstl.createNilLiteral(node);
+        const isWithinBindingAssignmentStatement = tsHelper.isWithinLiteralAssignmentStatement(node);
+        return isWithinBindingAssignmentStatement ? tstl.createAnonymousIdentifier() : tstl.createNilLiteral(node);
     }
 
     public transformDeleteExpression(expression: ts.DeleteExpression): ExpressionVisitResult {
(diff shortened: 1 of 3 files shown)


test/unit/array.spec.ts
```diff
@@ -191,3 +191,12 @@ test.each([
...
     expect(() => util.transpileAndExecute(code)).toThrowError(`invalid array length: ${result}`);
 });
+
+test("Array OmittedExpression", () => {
+    const result = util.transpileAndExecute(
+        `const myarray = [1, , 2];
+        return myarray[2];`
+    );
+
+    expect(result).toBe(2);
```
Perryvw: I think this test should check the result of the omittedexpression, not the expression after it (so check myarray[1] == undefined). I'm fine with also testing the elements after like you do now, either as additional expect in this test, or as a separate test.

test/unit/array.spec.ts
```diff
@@ -191,3 +191,21 @@ test.each([
     `;
     expect(() => util.transpileAndExecute(code)).toThrowError(`invalid array length: ${result}`);
 });
+
+test("Array OmittedExpression", () => {
+    const result = util.transpileAndExecute(
+        `const myarray = [,];
```
Perryvw: I'd prefer it if you kept the array the same in the two tests, so check index 1  on [1,,2] for this one.

hazzard993: I've used a `test.each` to check that array's contents

src/LuaTransformer.ts
```diff
@@ -2907,9 +2909,14 @@ export class LuaTransformer {
 
         if (ts.isArrayLiteralExpression(expression.left)) {
             // Destructuring assignment
+            const omittedExpressionAssignmentIdentifier = tstl.createAnonymousIdentifier();
             const left =
                 expression.left.elements.length > 0
-                    ? expression.left.elements.map(e => this.transformExpression(e))
+                    ? expression.left.elements.map(e =>
+                          ts.isOmittedExpression(e)
```
ark120202: Have you considered using `transformArrayBindingElement` there?

src/LuaTransformer.ts
```diff
@@ -3466,6 +3473,10 @@ export class LuaTransformer {
         return tstl.createTableExpression(properties, expression);
     }
 
+    public transformOmittedExpression(node: ts.OmittedExpression): ExpressionVisitResult {
+        return tstl.createNilLiteral(node);
```
ark120202: I think it's not really good to always return `nil` there. `nil` is valid only in one context it could be used (array literals), but it's not in other (destructuring).
I think it would be better to decide it in the visitor (using `node.parent`) and use it in `transformArrayBindingElement`.

hazzard993: This neatens things up nicely, I've added `isWithinLiteralAssignmentStatement` which can be used to check if a node is being used in a binding assignment expression. This should work with nested binding patterns too once those are supported.

I've called elements of an array on the left hand side of an assignment statement "ArrayBindingExpressions" since TS refers to them as Expressions and they seem to be a super-set of ArrayBindingElements with PropertyAccessExpressions being another thing they could be.

PR #646, opened 2019-06-29
Non-module namespace merge

### LINKED ISSUE #641: Missing parent namespace prefix when multiple "export" classes declared in the same "export" nested namespace
**Parent namespace prefix** are missed in all the rest classes except the one firstly scaned by TSTL if more than one classes are declared **in the same "export" sub namespace nested with its parent namespace**.

Class Test1 in Test1.ts
```
namespace _ns {
    export namespace _inner_ns {
        export class Test1 {
            public static func() :void {}
        }
    }
}
```

Class Test2 in Test2.ts
```
namespace _ns {
    export namespace _inner_ns {
        export class Test2 {
            public static func() :void {}
        }
    }
}
```
Test1 and Test2 included in TestIdx.ts by order:
```
import "./Test1"
import "./Test2"
```

lua translated from Test1.ts (which is OK)
```
_ns = {}
do
    _ns._inner_ns = {}
    local _inner_ns = _ns._inner_ns
    do
        _inner_ns.Test1 = {}
        local Test1 = _inner_ns.Test1
        Test1.name = "Test1"
        Test1.__index = Test1
        Test1.prototype = {}
        Test1.prototype.__index = Test1.prototype
        Test1.prototype.constructor = Test1
        function Test1.new(...)
            local self = setmetatable({}, Test1.prototype)
            self:____constructor(...)
            return self
        end
        function Test1.prototype.____constructor(self)
        end
        function Test1.func(self)
        end
    end
end
```

lua translated from Test2.ts (in which **parent namespace prefix missed**)
```
do
    do
       -- now parent namespace "_ns" missed
        _inner_ns.Test2 = {}
        local Test2 = _inner_ns.Test2
        Test2.name = "Test2"
        Test2.__index = Test2
        Test2.prototype = {}
        Test2.prototype.__index = Test2.prototype
        Test2.prototype.constructor = Test2
        function Test2.new(...)
            local self = setmetatable({}, Test2.prototype)
            self:____constructor(...)
            return self
        end
        function Test2.prototype.____constructor(self)
        end
        function Test2.func(self)
        end
    end
end
```

Class Test1 would be error if Class Test2 is included before Test1.

thanks!

tomblind:
fixes #641 

When namespaces with the same name are declared in non-module files, TS assumes they'll be merged somehow (the example in the bug wouldn't work as you can't import non-modules, but there are other ways they code could be merged).

This fixes that case by using an `or` assignment when a namespace is declared in a non-module file.
```lua
NS = NS or {}
```

### DIFF
--- a/src/LuaTransformer.ts
+++ b/src/LuaTransformer.ts
@@ -1585,8 +1585,13 @@ export class LuaTransformer {
 
         const symbol = this.checker.getSymbolAtLocation(statement.name);
         const hasExports = symbol !== undefined && this.checker.getExportsOfModule(symbol).length > 0;
-
         const nameIdentifier = this.transformIdentifier(statement.name as ts.Identifier);
+        const exportScope = this.getIdentifierExportScope(nameIdentifier);
+
+        // Non-module namespace could be merged if:
+        // - is top level
+        // - is nested and exported
+        const isNonModuleMergeable = !this.isModule && (!this.currentNamespace || exportScope);
 
         // This is NOT the first declaration if:
         // - declared as a module before this (ignore interfaces with same name)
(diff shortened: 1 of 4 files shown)


ark120202:
Is there really a need to have any logic there? TS seems to use code equal to `ns = ns or {}` always when namespace is used.

PR #647, opened 2019-06-29
Feature/emit host

Perryvw:


ark120202:
I think it would be better to make `EmitHost` a subset of system, like all other hosts in TS are, so default for that would be just `ts.sys` and we won't have a direct dependency on `fs` there.

```ts
interface EmitHost {
    readFile(path: string): string | undefined;
    writeFile: ts.WriteFileCallback;
}
```

The signature of `readFile` is a bit different and default encoding is utf8, so it might need some changes on usage.

Also, `writeFile` doesn't seem to be used there.

Perryvw:
> I think it would be better to make `EmitHost` a subset of system, like all other hosts in TS are, so default for that would be just `ts.sys` and we won't have a direct dependency on `fs` there.
> 
> ```ts
> interface EmitHost {
>     readFile(path: string): string | undefined;
>     writeFile: ts.WriteFileCallback;
> }
> ```
> 
> The signature of `readFile` is a bit different and default encoding is utf8, so it might need some changes on usage.
> 
> Also, `writeFile` doesn't seem to be used there.

@ark120202 good call, I changed that. Any idea how to further organize this? I'm not really happy with the multiple definitions of the same default emit host.

ark120202:
@Perryvw if host it a subset of `ts.System`, we can just use default system as a default host (`= ts.sys`).

src/index.ts
```diff
@@ -28,8 +28,9 @@ export interface TranspileFilesResult {
 
 export function transpileFiles(rootNames: string[], options: CompilerOptions = {}): TranspileFilesResult {
     const program = ts.createProgram(rootNames, options);
+    const emitHost = { readFile: ts.sys.readFile };
     const { transpiledFiles, diagnostics: transpileDiagnostics } = transpile({ program });
-    const emitResult = emitTranspiledFiles(program.getCompilerOptions(), transpiledFiles);
+    const emitResult = emitTranspiledFiles(program.getCompilerOptions(), transpiledFiles, emitHost);
```
ark120202: Since `emitTranspiledFiles` has a default host it shouldn't be required there

src/Transpile.ts
```diff
@@ -27,14 +27,20 @@ export interface TranspileOptions {
     customTransformers?: ts.CustomTransformers;
     transformer?: LuaTransformer;
     printer?: LuaPrinter;
+    emitHost?: EmitHost;
+}
+
+export interface EmitHost {
+    readFile: (path: string, encoding?: string | undefined) => string | undefined;
```
ark120202: If we never specify encoding we can just drop this argument to simplify custom implementations (ts does that with few hosts). Also, it could be a bit more consistent with method syntax:

```suggestion
    readFile(path: string): string | undefined;
```

PR #648, opened 2019-07-01
Out file support

### LINKED ISSUE #357: Single file target
Add an option to `tstl` that produces a single-file output. This is useful for environments where only a single Lua artifact can be provided such as Tabletop Simulator. This will avoid the need for secondary bundling solutions which may or may not work with TypescriptToLua's output. `tstl` is in the best position to compose the output for single file output.

hazzard993:
Closes #357

Tried out supporting the `--outFile` option and seemed to make some good progress.

- [x] Modules are placed in the output file in the correct order
- [x] Modules are registered in `tstlpackage.preload`
- [x] LuaLibs can be used
- [x] outFiles do not depend on `require` and `package` to exist in the target Lua environment
- [x] outFiles can deal with cyclic imports

The new option `luaEntry` can be used to specify what file will be executed when running the outFile. The exports of that TypeScript file are what the outFile as a whole will return.

Any other high or low level considerations regarding a single output file would be appreciated on this PR.


src/LuaTransformer.ts
```diff
@@ -141,11 +141,31 @@ export class LuaTransformer {
...
+                    );
+                } else {
+                    // return exports
+                    statements.push(tstl.createReturnStatement([this.createExportsIdentifier()]));
+                }
             }
         }
 
+        if (this.options.outFile) {
+            return [tstl.createBlock([tstl.createDoStatement(statements)], node), this.luaLibFeatureSet];
```
tomblind: We shouldn't wrap these in blocks when not in a module, as variables declared with let/const in one script should be able to be referenced from another.

ark120202:
`outFile` in TS has 2 modes:

First one is a file concatenation. It is enabled when module is anything except system or amd. What it's doing is simply joining all files, without any wrappers or something like that. Because of that everything declared in files is available in all other files (what @tomblind said in a comment above). However, it doesn't allow imports and exports in all files in the project.

The second mode is bundling. It's enabled when module is `system` or `amd`. In this mode all files are wrapped in a function that registers modules, but not executes.This mode should be used with some runtime (for example https://requirejs.org/). Loading output compiled with this mode won't execute any module, so user has to call something like `require(["entrypointModule"])` manually.

Currently this PR mixes both of these ways, which won't work in a lot of cases I can think of.

`package.preload` may be used there, but there could be environments without `package`.

> Might have to think about giving the --module option a default value as a value has to be picked but whatever it is it currently has no effect on the output.

I don't think we should change any options implicitly, because if it's not specified in config it won't work with language service anyway.

ark120202:
Sorry, closed it accidentally

src/Transpile.ts
```diff
@@ -140,5 +144,23 @@ export function transpile({
...
     }
 
+    if (options.outFile) {
+        const lua = transpiledFiles.reduce((luaCode, transpiledFile) => {
+            if (transpiledFile.luaAst && transpiledFile.lua) {
+                luaCode += transpiledFile.lua;
+            }
+            return luaCode;
+        }, "");
+        const declarationFile = transpiledFiles.find(transpiledFile => transpiledFile.declaration);
```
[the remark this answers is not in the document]
ark120202: TS already bundles declarations when `outFile` is specified. Though it doesn't really work well.

ledniy:
Any progress on that? I would like to use it for single output file

hazzard993:
I'll look into this some more today

hazzard993:
I'm leaving source map tracebacks for bundles out of scope for this PR as I'm not too sure how to proceed with it and this PR is pretty old but does allow --outFile support.

I've added --luaModuleSystem (`"none"` or `"require"`) as a new option which allows users to generate code that can be used in environments without `package` or `require` if they want to output a single file.

src/LuaTransformer.ts
```diff
@@ -104,21 +104,53 @@ export class LuaTransformer {
...
+                this.resolver = this.checker.getEmitResolver(originalSourceFile);
+                const transformResult = this.transformSourceFile(sourceFile);
+                return [...statements, ...transformResult.statements];
+            },
+            []
+        );
+
+        if (this.options.luaEntry) {
+            const requireCalls = this.options.luaEntry.map(entry => {
+                const formattedEntryName = tsHelper.formatPathToLuaPath(entry.replace(/.ts$/, ""));
```
ark120202: That looks very loose. I think we should decide either to do a full path resolution (i.e. `./src/file.ts` -> `file`, make it relative to tsconfig file location) or don't transform it at all. I personally lean towards the second option now.

src/LuaTransformer.ts
```diff
@@ -104,21 +104,53 @@ export class LuaTransformer {
...
+                return tstl.createExpressionStatement(
+                    this.createModuleRequire(ts.createStringLiteral(formattedEntryName), false)
+                );
+            });
+            combinedStatements.push(...requireCalls);
```
ark120202: We should also return something from the bundle. There are 3 options:
- Merge all exported values (might have non-obvious behavior with default exports)
- Return only last entry
- Support only one entry point (I think that would be the best option)

src/lualib/LuaRequire.ts
```diff
@@ -0,0 +1,14 @@
+const tstlpackage = { preload: {}, loaded: {} };
+
+function __TS__LuaRequire(this: void, moduleName: string): any {
+    if (!tstlpackage.loaded[moduleName]) {
+        const module: (this: void, module: string) => any = tstlpackage.preload[moduleName];
+        if (module) {
+            tstlpackage.loaded[moduleName] = module(moduleName);
+        } else {
+            // tslint:disable-next-line: no-string-throw
+            throw `module '${moduleName}' not found:`;
```
ark120202: I guess we won't have a resolution stack, so

```suggestion
            throw `module '${moduleName}' not found`;
```

test/unit/outFile.spec.ts
```diff
@@ -0,0 +1,152 @@
...
+            `,
+            "main.ts": `
+                import { array } from "./module";
+                if (array[2] !== 3) {
+                    throw "Array's third item is not three";
+                }            
+            `,
+        },
+    ],
+])("outFile tests (%s)", (_, files) => {
```
ark120202: I think it's better to have individual tests instead, common code can be extracted to a function

test/unit/outFile.spec.ts
```diff
@@ -0,0 +1,152 @@
...
+    const testBuilder = util.testBundle`
+        ${files["main.ts"]}
+    `.setOptions({ outFile: "main.lua", module: ts.ModuleKind.AMD, luaEntry: ["main.ts"] });
+
+    const extraFiles = Object.keys(files)
```
ark120202:
```
Object.entries(files)
	.filter(([fileName]) => fileName !== "main.ts")
	.forEach(([fileName, code]) => testBuilder.addExtraFile(fileName, code))
```

src/lualib/LuaRequire.ts
```diff
@@ -0,0 +1,14 @@
+const tstlpackage = { preload: {}, loaded: {} };
+
+function __TS__LuaRequire(this: void, moduleName: string): any {
+    if (!tstlpackage.loaded[moduleName]) {
+        const module: (this: void, module: string) => any = tstlpackage.preload[moduleName];
+        if (module) {
+            tstlpackage.loaded[moduleName] = module(moduleName);
```
ark120202: What about cyclic imports?

src/LuaTransformer.ts
hazzard993: Changed `luaEntry` to only accept a single string specifying the path to the TypeScript file that is intended to be the entrypoint to the output file

src/lualib/LuaRequire.ts
hazzard993: Good opportunity to support that here. Those should work now

src/CommandLineParser.ts
```diff
@@ -22,9 +22,18 @@ interface CommandLineOptionOfBoolean extends CommandLineOptionBase {
     type: "boolean";
 }
 
-type CommandLineOption = CommandLineOptionOfEnum | CommandLineOptionOfBoolean;
+interface CommandLineOptionOfString extends CommandLineOptionBase {
```
ark120202: Maybe it would be better to

```ts
interface CommandLineOptionOfPrimitiveType extends CommandLineOptionBase {
    type: "string" | "boolean";
}

switch (option.type) {
    case "string":
    case "boolean": {
      if (typeof value !== option.type) {
          return {
              value: undefined,
              error: diagnosticFactories.compilerOptionRequiresAValueOfType(option.name, option.type),
          };
      }

      return { value };
    }
}
```

src/LuaTransformer.ts
```diff
@@ -104,21 +104,55 @@ export class LuaTransformer {
...
+        } else {
+            return [this.transformBundle(node), this.luaLibFeatureSet];
+        }
+    }
+
+    public transformBundle(bundle: ts.Bundle): tstl.Block {
+        this.isWithinBundle = true;
+        const combinedStatements = bundle.sourceFiles.reduce(
+            (statements: tstl.Statement[], sourceFile: ts.SourceFile) => {
+                this.currentSourceFile = sourceFile;
```
ark120202: Instead of duplicating it, it can be extracted to something like `transformSourceFileWithState`

src/LuaTransformer.ts
```diff
@@ -104,21 +104,55 @@ export class LuaTransformer {
...
+                const originalSourceFile = ts.getParseTreeNode(sourceFile, ts.isSourceFile) || sourceFile;
+                this.resolver = this.checker.getEmitResolver(originalSourceFile);
+                const transformResult = this.transformSourceFile(sourceFile);
+                return [...statements, ...transformResult.statements];
+            },
+            []
+        );
+
+        if (this.options.luaEntry) {
+            const sourceFile = this.program.getSourceFile(this.options.luaEntry);
```
ark120202: Wouldn't `getSourceFile` expect absolute path?

test/unit/outFile.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+import * as ts from "typescript";
+import * as TSTLErrors from "../../src/TSTLErrors";
+import { CompilerOptions } from "../../src/CompilerOptions";
+
+const exportValueSource = "export const value = true;";
+const reexportValueSource = 'export { value } from "./export";';
+const exportResultSource = "export const result = value;";
+
+function outFileOptionsWithEntry(luaEntry: string): CompilerOptions {
+    return { outFile: "main.lua", noHoisting: true, module: ts.ModuleKind.AMD, luaEntry };
```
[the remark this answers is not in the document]
hazzard993: Needed it for the cyclic imports outFile test. This option is now only used for that test alone

src/LuaTransformer.ts
hazzard993: It works with relative paths too but they seem to resolve based on the host's working directory so this code causes an issue if the CLI is in a different directory to a specified tsconfig file.

I've changed it so if the CLI command is given the `luaEntry` option, it will use that relative name in `program.getSourceFile(...)` whose host seems to resolve relative paths to the current directory.

However if a project/tsconfig location is specified via CLI, a full path is resolved based off the project/tsconfig location and given to `program.getSourceFile(...)` instead.

src/lualib/LuaRequire.ts
```diff
@@ -0,0 +1,17 @@
+const __TS__MODULES: Record<string, (this: void, exports: object) => any> = {};
```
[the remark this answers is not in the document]
ark120202: Well, I guess `__TS__Modules` isn't camelCase anyway. Maybe just `____modules` and `____moduleCache` then?

src/LuaTransformer.ts
```diff
@@ -104,24 +102,54 @@ export class LuaTransformer {
...
+                const transformResult = this.transformSourceFileWithState(sourceFile);
+                return [...statements, ...transformResult.statements];
+            },
+            []
+        );
+
+        if (this.options.luaEntry) {
+            // If specified via CLI, the program's host should pick up a relative luaEntry path
+            let luaEntryPath = this.options.luaEntry;
+            if (this.options.configFilePath) {
```
ark120202: A case when `configFileName === undefined` is already handled inside this condition

src/LuaTransformer.ts
```diff
@@ -511,7 +565,11 @@ export class LuaTransformer {
               )
             : moduleSpecifier.text;
         const modulePath = tstl.createStringLiteral(modulePathString);
-        return tstl.createCallExpression(tstl.createIdentifier("require"), [modulePath], moduleSpecifier);
+        const requireCallString = this.options.outFile ? "__TS__LuaRequire" : "require";
```
```suggestion
        const requireFunctionName = this.options.outFile ? "__TS__LuaRequire" : "require";
```

src/LuaTransformer.ts
```diff
@@ -104,24 +102,54 @@ export class LuaTransformer {
...
+
+        if (this.options.luaEntry) {
+            // If specified via CLI, the program's host should pick up a relative luaEntry path
+            let luaEntryPath = this.options.luaEntry;
+            if (this.options.configFilePath) {
+                // If specified via project. Figure out the full path to the file based off the project's path
+                const configFileName = this.options.configFilePath as string | undefined;
+                const basedir = configFileName ? path.dirname(configFileName) : process.cwd();
+                luaEntryPath = path.resolve(basedir, this.options.luaEntry);
+            }
```
```suggestion
            }

```

src/LuaTransformer.ts
```diff
@@ -511,7 +565,11 @@ export class LuaTransformer {
               )
             : moduleSpecifier.text;
         const modulePath = tstl.createStringLiteral(modulePathString);
-        return tstl.createCallExpression(tstl.createIdentifier("require"), [modulePath], moduleSpecifier);
+        const requireCallString = this.options.outFile ? "__TS__LuaRequire" : "require";
+        if (this.options.outFile) {
+            this.importLuaLibFeature(LuaLibFeature.LuaRequire);
+        }
```
```suggestion
        }

```

src/diagnostics.ts
```diff
@@ -2,9 +2,9 @@ import * as ts from "typescript";
 import { TranspileError } from "./TranspileError";
 
 export const transpileError = (error: TranspileError): ts.Diagnostic => ({
-    file: error.node.getSourceFile(),
-    start: error.node.getStart(),
-    length: error.node.getWidth(),
+    file: error.node.kind === ts.SyntaxKind.Unknown ? undefined : error.node.getSourceFile(),
```
ark120202: I wonder what error required that change?

hazzard993: It is because of `TSTLErrors.LuaEntryNotFound`. A TranspilerError which requires a node which isn't available when encountering this error

test/unit/outFile.spec.ts
```diff
@@ -0,0 +1,113 @@
+import * as util from "../util";
+import * as TSTLErrors from "../../src/TSTLErrors";
+
+const exportValueSource = "export const value = true;";
+const reexportValueSource = 'export { value } from "./export";';
+const exportResultSource = "export const result = value;";
```
ark120202: `exportValueSource` and `exportResultSource` have similar names, but assign different values. I'm not really sure there's a need for these to be variables, it might be easier to read with it inlined.

src/lualib/LuaRequire.ts
```diff
@@ -0,0 +1,18 @@
+// tslint:disable: variable-name
+const ____modules: Record<string, (this: void, exports: object) => any> = {};
+const ____modulecache: Record<string, any> = {};
```
ark120202:
```suggestion
const ____moduleCache: Record<string, any> = {};
```

src/TSTLErrors.ts
```diff
@@ -60,6 +60,9 @@ export const InvalidThrowExpression = (node: ts.Node) =>
 export const ForbiddenStaticClassPropertyName = (node: ts.Node, name: string) =>
     new TranspileError(`Cannot use "${name}" as a static class property or method name.`, node);
 
+export const LuaEntryNotFound = (entryName: string) =>
+    new TranspileError(`Could not find luaEntry file '${entryName}'.`, ts.createNode(ts.SyntaxKind.Unknown));
```
ark120202: `ts.createNode(ts.SyntaxKind.Unknown)` is quite hacky.
I think it would be better to make `node` in `TranspileError` not required

test/unit/outFile.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+        ${exportResultSource}
+    `
+        .addExtraFile(
+            "b.ts",
+            `
+                import { a } from "./main";
+                export const value = a;
+            `
+        )
+        .setOptions({ noHoisting: true })
```
ark120202: It should be possible to test it without `noHoisting`, like

```ts
// b.ts
import { a } from "./main";
export const value = a;
export const lazyValue = () => a;
```

```ts
// main.ts
import { value, lazyValue } from "./b";
export const a = true;
export const valueResult = value;
export const lazyValueResult = lazyValue();
```

src/LuaTransformer.ts
```diff
@@ -136,25 +160,51 @@ export class LuaTransformer {
...
+                tstl.createTableIndexExpression(moduleTableIdentifier, tstl.createStringLiteral(exportPath)),
+                tstl.createFunctionExpression(tstl.createBlock(statements, sourceFile), moduleParameters)
+            );
+            return tstl.createBlock([moduleDeclaration], sourceFile);
+        }
+
         return tstl.createBlock(statements, sourceFile);
     }
 
+    private transformSourceFileWithState(node: ts.SourceFile): tstl.Block {
```
ark120202: I think it would be a bit clearer to move it above `transformSourceFile`, since that's an internal method

Perryvw:
Bundling implemented by #750 instead.

PR #655, opened 2019-07-08
Separate transformer run method

ark120202:
See https://github.com/TypeScriptToLua/TypeScriptToLua/pull/565#discussion_r281023939.

All transformer state initialization is extracted to `run` method. `transformSourceFile` is now just a simple overridable visitor. Since `setupState` is now guaranteed to be called (unless one calls visitors directly), I have also removed initializers duplication.

### DIFF
--- a/src/LuaPrinter.ts
+++ b/src/LuaPrinter.ts
@@ -38,18 +38,9 @@ export class LuaPrinter {
         [tstl.SyntaxKind.BitwiseNotOperator]: "~",
     };
 
-    private options: CompilerOptions;
-    private emitHost: EmitHost;
-
-    private currentIndent: string;
-
+    private currentIndent = "";
     private sourceFile = "";
-
-    public constructor(options: CompilerOptions, emitHost: EmitHost) {
-        this.options = options;
-        this.emitHost = emitHost;
-        this.currentIndent = "";
-    }
+    public constructor(private options: CompilerOptions, private emitHost: EmitHost) {}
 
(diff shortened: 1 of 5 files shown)


src/LuaPrinter.ts
```diff
@@ -38,18 +38,9 @@ export class LuaPrinter {
...
-
+    private currentIndent = "";
     private sourceFile = "";
-
-    public constructor(options: CompilerOptions, emitHost: EmitHost) {
-        this.options = options;
-        this.emitHost = emitHost;
-        this.currentIndent = "";
-    }
+    public constructor(private options: CompilerOptions, private emitHost: EmitHost) {}
```
Perryvw: I wouldn't mind a newline before the constructor

src/LuaTransformer.ts
```diff
@@ -54,85 +54,81 @@ export interface DiagnosticsProducingTypeChecker extends ts.TypeChecker {
...
 
-    public transformSourceFile(node: ts.SourceFile): [tstl.Block, Set<LuaLibFeature>] {
-        this.setupState();
+    protected currentSourceFile!: ts.SourceFile;
+    protected isModule!: boolean;
+    protected resolver!: EmitResolver;
 
-        this.currentSourceFile = node;
+    /** @internal */
+    public run(sourceFile: ts.SourceFile): [tstl.Block, Set<LuaLibFeature>] {
```
Perryvw: Maybe `transform` would be better? I don't really like `run` as a name.

PR #658, opened 2019-07-09
Fixed empty destructuring declarations/assignments in for...of loops

### LINKED ISSUE #607: luaIterator + tupleReturn used with empty destructuring
```ts
/** @luaIterator @tupleReturn */
interface MyIter extends Array<[string]> {}

declare function iter(this: void): MyIter;

for (const [] of iter()) {}
```
=>
```lua
for  in iter() do -- no bueno
end
```

tomblind:
fixes #607

### DIFF
--- a/test/unit/loops.spec.ts
+++ b/test/unit/loops.spec.ts
@@ -754,6 +754,142 @@ test("forof forwarded lua iterator with tupleReturn", () => {
     expect(result).toBe("0a1b2c");
 });
 
+test("forof array empty destructuring declaration", () => {
+    const code = `
+        const arr = [["a"], ["b"], ["c"]];
+        let i = 0;
+        for (const [] of arr) {
+            ++i;
+        }
+        return i;
+    `;
+    expect(util.transpileAndExecute(code)).toBe(3);
+});
+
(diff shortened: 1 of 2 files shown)


test/unit/loops.spec.ts
```diff
@@ -754,6 +754,142 @@ test("forof forwarded lua iterator with tupleReturn", () => {
     expect(result).toBe("0a1b2c");
 });
 
+test("forof array empty destructuring declaration", () => {
```
ark120202: Declaration/assignment versions of each test take quite a lot of space, maybe it would be better to generate them if the only difference is `const []`/`[]`? Something like:

```ts
describe("for...of empty destructuring", () => {
    const declareTests = (destructuringPrefix: string) => {
        test("array", () => {
            const code = `
                const arr = [["a"], ["b"], ["c"]];
                let i = 0;
                for (${destructuringPrefix}[] of arr) {
                    ++i;
                }
                return i;
            `;
            expect(util.transpileAndExecute(code)).toBe(3);
        });

        // ...
    };

    describe("declaration", () => declareTests("const "));
    describe("assignment", () => declareTests(""));
});
```

PR #667, opened 2019-07-15
Preserve newlines in template literals

martinjlowm:
edit:

The only equivalent string type TS has to `[=*[ ... ]=*]` where nothing can be escaped is ``String.raw`...` ``. However, the added code complexity is not worth the benefits of supporting template literals as multiline strings in Lua (and assuming they are equivalent).

The only benefit of transpiling template literals into multiline strings is to have a pretty output.

For those, searching for a way to make things pretty, I recommend passing the output through some sort of prettier, not just for strings, but for the general output of tstl.

- ~~All template literals are now transpiled into `[=*[ ... ]=*]`~~

- ~~Newlines within template literals are tricky, ideally String.raw would be
  handy to distinguish between escaped- and non-escaped newlines (`\n`)~~

- ~~Added three tests to cover growing multiline string (increasing number of
  `=`'s)~~

- ~~NoSubstitutionTemplateLiteral is now covered by MultilineStringLiteral~~

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
             "integrity": "sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==",
             "dev": true,
             "requires": {
-                "@babel/highlight": "^7.0.0"
+                "@babel/highlight": "7.0.0"
             }
         },
         "@babel/core": {
@@ -19,20 +19,20 @@
             "integrity": "sha512-Dzl7U0/T69DFOTwqz/FJdnOSWS57NpjNfCwMKHABr589Lg8uX1RrlBIJ7L5Dubt/xkLsx0xH5EBFzlBVes1ayA==",
             "dev": true,
             "requires": {
-                "@babel/code-frame": "^7.0.0",
-                "@babel/generator": "^7.4.0",
-                "@babel/helpers": "^7.4.0",
-                "@babel/parser": "^7.4.0",
-                "@babel/template": "^7.4.0",
-                "@babel/traverse": "^7.4.0",
-                "@babel/types": "^7.4.0",
-                "convert-source-map": "^1.1.0",
-                "debug": "^4.1.0",
-                "json5": "^2.1.0",
-                "lodash": "^4.17.11",
-                "resolve": "^1.3.2",
-                "semver": "^5.4.1",
-                "source-map": "^0.5.0"
+                "@babel/code-frame": "7.0.0",
+                "@babel/generator": "7.4.0",
+                "@babel/helpers": "7.4.2",
+                "@babel/parser": "7.4.2",
+                "@babel/template": "7.4.0",
+                "@babel/traverse": "7.4.0",
+                "@babel/types": "7.4.0",
+                "convert-source-map": "1.6.0",
+                "debug": "4.1.1",
+                "json5": "2.1.0",
+                "lodash": "4.17.11",
+                "resolve": "1.11.0",
+                "semver": "5.5.0",
+                "source-map": "0.5.7"
             },
             "dependencies": {
                 "@babel/generator": {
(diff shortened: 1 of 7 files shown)


ark120202:
Transpiling

```ts
const lines = [];
lines.join("\n\n");
```

to

```lua
local lines = {}
table.concat(lines, [[


]])
```

doesn't sound good to me

martinjlowm:
true - I had to leave for work, I'll find a better solution tonight.

ark120202:
Some thoughts:
- String literal should get a `raw` property (to differentiate between `\n` in literal and actual line break) and some `originalNodeKind`-like field. Alternatively, a single `rawTemplateLiteral` property
- Just replacing backslash won't work for most of escaped values, so it should fallback to regular literal in these cases
- Literal may contain `]]`, or `]=]`, etc, so printer has to choose the right number of `=`s

I'm not entirely convinced that it would worth the added complexity, but I'm open to other opinions.

martinjlowm:
How about making a separate literal (kind) for `[=*[ ... ]=*]` that is directly converted from TypeScript's template literal?

src/LuaPrinter.ts
```diff
@@ -463,6 +465,23 @@ export class LuaPrinter {
         return this.createSourceNode(expression, `"${expression.value}"`);
     }
 
+    public printMultilineStringLiteral(expression: tstl.MultilineStringLiteral): SourceNode {
+        const chunks = [];
+
+        const braces = ["\\]"];
```
ark120202:
```ts
let braces = 0;
while (expression.value.includes("]".repeat(braces)) {
    braces++;
}

const bracesSeparators = "=".repeat(braces);
```

src/LuaTransformer.ts
```diff
@@ -4817,11 +4818,16 @@ export class LuaTransformer {
         return this.transformLuaLibFunction(LuaLibFeature.Spread, expression, innerExpression);
     }
 
-    public transformStringLiteral(literal: ts.StringLiteralLike): ExpressionVisitResult {
+    public transformStringLiteral(literal: ts.StringLiteral): ExpressionVisitResult {
         const text = tsHelper.escapeString(literal.text);
         return tstl.createStringLiteral(text, literal);
     }
 
+    public transformMultilineStringLiteral(literal: ts.NoSubstitutionTemplateLiteral): ExpressionVisitResult {
```
```suggestion
    public transformNoSubstitutionTemplateLiteral(literal: ts.NoSubstitutionTemplateLiteral): ExpressionVisitResult {
```

src/TSHelper.ts
```diff
@@ -659,11 +659,12 @@ export function getFunctionContextType(type: ts.Type, checker: ts.TypeChecker):
     return reduceContextTypes(signatureDeclarations.map(s => getDeclarationContextType(s, checker)));
 }
 
-export function escapeString(text: string): string {
+export function escapeString(text: string, exclude: RegExp[] = []): string {
     // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String
     const escapeSequences: Array<[RegExp, string]> = [
         [/[\\]/g, "\\\\"],
-        [/[\']/g, "\\'"],
+        [/\\'/g, "\\'"],
```
ark120202: This replacement does nothing.
I guess it can be safely removed, it wasn't ever required anyway.

src/TSHelper.ts
```diff
@@ -659,11 +659,12 @@ export function getFunctionContextType(type: ts.Type, checker: ts.TypeChecker):
...
+        [/[\`]/g, "\\\\`"],
```
ark120202: I don't get it, why is it required?

src/TSHelper.ts
```diff
@@ -659,11 +659,12 @@ export function getFunctionContextType(type: ts.Type, checker: ts.TypeChecker):
     return reduceContextTypes(signatureDeclarations.map(s => getDeclarationContextType(s, checker)));
 }
 
-export function escapeString(text: string): string {
+export function escapeString(text: string, exclude: RegExp[] = []): string {
```
ark120202: I don't really like ability to specify any unrelated regexps there.
Maybe just a `multiline` flag?

package-lock.json
```diff
@@ -10,7 +10,7 @@
             "integrity": "sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==",
```
ark120202: What's with these lockfile changes? Have you used `npm update`?

PR #668, opened 2019-07-15
Constrained generic fixes

### LINKED ISSUE #489: Array detection fails on constrained generics
## Input
```typescript
function f<A extends any[]>(a: A){
  a.forEach(v => console.log(v));
}
```
## Current Result
```lua
f = function(self, a)
    a:forEach(function(____, v)
        return print(v);
    end);
end;
```
## Expected Result
```lua
f = function(self, a)
    __TS__ArrayForEach((a), function(____, v)
        return print(v);
    end);
end;
```

hazzard993:
Closes #489

Also fixed it for strings too. Thanks @ark120202 for the solution idea

### DIFF
--- a/src/TSHelper.ts
+++ b/src/TSHelper.ts
@@ -157,9 +157,16 @@ export function isExplicitArrayType(type: ts.Type, checker: ts.TypeChecker, prog
         return true;
     }
 
+    if (type.symbol) {
+        const baseConstraint = checker.getBaseConstraintOfType(type);
+        if (baseConstraint) {
+            return isExplicitArrayType(baseConstraint, checker, program);
+        }
+    }
+
     const flags = ts.NodeBuilderFlags.InTypeAlias | ts.NodeBuilderFlags.AllowEmptyTuple;
     const typeNode = checker.typeToTypeNode(type, undefined, flags);
-    return typeNode !== undefined && (ts.isArrayTypeNode(typeNode) || ts.isTupleTypeNode(typeNode));
+    return typeNode !== undefined ? ts.isArrayTypeNode(typeNode) || ts.isTupleTypeNode(typeNode) : false;
 }
 
 export function isFunctionType(type: ts.Type, checker: ts.TypeChecker): boolean {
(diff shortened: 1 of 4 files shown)


src/TSHelper.ts
```diff
@@ -157,9 +164,16 @@ export function isExplicitArrayType(type: ts.Type, checker: ts.TypeChecker, prog
...
+        const baseConstraint = checker.getBaseConstraintOfType(type);
+        if (baseConstraint) {
+            return isExplicitArrayType(baseConstraint, checker, program);
+        }
+    }
+
     const flags = ts.NodeBuilderFlags.InTypeAlias | ts.NodeBuilderFlags.AllowEmptyTuple;
     const typeNode = checker.typeToTypeNode(type, undefined, flags);
-    return typeNode !== undefined && (ts.isArrayTypeNode(typeNode) || ts.isTupleTypeNode(typeNode));
+    return typeNode !== undefined ? ts.isArrayTypeNode(typeNode) || ts.isTupleTypeNode(typeNode) : false;
```
Perryvw: I prefer the previous version of this line, why the change?

hazzard993: Shouldn't have touched that line, going with preference

src/TSHelper.ts
```diff
@@ -157,6 +164,13 @@ export function isExplicitArrayType(type: ts.Type, checker: ts.TypeChecker, prog
         return true;
     }
 
+    if (type.symbol) {
```
ark120202: It shouldn't affect correctness, but would it maybe be better to make it a first check in the function, so that a technical detail is checked before the actual function logic?

test/unit/array.spec.ts
```diff
@@ -226,3 +226,36 @@ test.each([`["foo", "bar"].length`, `["foo", "bar"][0]`, `[() => "foo", () => "b
         expect(util.transpileAndExecute(code)).toBe(expectResult);
     }
 );
+
+const genericChecks = [
+    "function generic<T extends number[]>(array: T)",
+    "function generic<T extends [...number[]]>(array: T)",
+    "function generic<T extends any>(array: T[])",
+    "type ArrayType = number[]; function generic<T extends ArrayType>(array: T)",
+];
```
ark120202: Few cases that may worth to be tested:

```suggestion
    "function generic<T extends number[]>(array: T & {})",
    "function generic<T extends number[] & {}>(array: T)",
];
```

Issue #615, opened 2019-05-29
Pluggable transformer design


ark120202:
`LuaTransformer` class now has a number of issues:
* Ability to extend it isn't a plugin system - you can do it only once, you can override things you aren't meant to, we are limited with public/protected/private when designing visitor API
* If you want to extend it, you always have to use it with API, so you also have to re-implement `tstl` even when you don't want to change anything there
* 5500 lines is too much

The idea is to make whole transformation be composed of plugins. A plugin is a function that receives `context: TranspilerContext` and returns an object. Context is an object with some state that may be passed to utility functions.

Project structure:
- src/transformers
  - utils
    - index.ts
    - type-checks.ts
    - lua-ast-utils.ts
    - ...
  - class.ts
  - call.ts
  - try.ts
  - ...

Example transformer:

```ts
import { escapeString, isTupleReturnCall, isArrayType, createUnpackCall, transformLuaLibFunction } from "./utils";

export const someTransformer = (context: TranspilerContext): TranspilerPlugin => ({
    visitors: {
        [ts.SyntaxKind.StringLiteral]: node => escapeString(node.text),
        [ts.SyntaxKind.SpreadElement](node) {
            const innerExpression = context.transformExpression(node.expression);
            if (isTupleReturnCall(context, node.expression)) {
                return innerExpression;
            }

            const type = context.checker.getTypeAtLocation(node.expression);
            if (isArrayType(context, type)) {
                return createUnpackCall(innerExpression, node);
            }

            return transformLuaLibFunction(context, LuaLibFeature.Spread, node, innerExpression);
        },
    },
});
```

Perryvw:
I'm archiving some of the discussion related to this issue on discord:

A possible way to achieve this design with mixins:
[...]

PR #670, opened 2019-07-17
Export default and export equals

### LINKED ISSUE #606: Support default imports/exports

hazzard993:
Closes #606

As far as I know this PR should make TypeScriptToLua support all import and export statements. This doesn't include call expression `import()` (Promise).

Highlighted below are the new statements that can be written.

```diff
+ import defaultExport from "module-name";
  import * as name from "module-name";
  import { export } from "module-name";
  import { export as alias } from "module-name";
  import { export1 , export2 } from "module-name";
  import { foo , bar } from "module-name/path/to/specific/un-exported/file";
  import { export1 , export2 as alias2 , [...] } from "module-name";
+ import defaultExport, { export [ , [...] ] } from "module-name";
+ import defaultExport, * as name from "module-name";
  import "module-name";
  
  // Exporting individual features
  export let name1, name2, …, nameN; // also var, const
  export let name1 = …, name2 = …, …, nameN; // also var, const
  export function functionName(){...}
  export class ClassName {...}
  
  // Export list
  export { name1, name2, …, nameN };
  
  // Renaming exports
  export { variable1 as name1, variable2 as name2, …, nameN };
  
  // Exporting destructured assignments with renaming
  export const { name1, name2: bar } = o;
  
  // Default exports
+ export default expression;
+ export default function (…) { … }
+ export default function name1(…) { … }
+ export default class { … }
+ export default class Class { … }
+ export { name1 as default, … };
  
  // Aggregating modules
  export * from …;
  export { name1, name2, …, nameN } from …;
  export { import1 as name1, import2 as name2, …, nameN } from …;
+ export { default } from …;

  // Export only one value
+ export = 5;
```

Moved some path resolving logic over to tsHelper to allow it to be used by testing utilities. Using this I've made a new utility to functionally test module imports and exports.


src/LuaTransformer.ts
```diff
@@ -779,18 +826,35 @@ export class LuaTransformer {
...
+                result.push(
+                    tstl.createVariableDeclarationStatement(
+                        localClassName,
+                        this.createExportedIdentifier(tstl.cloneIdentifier(className), exportScope)
+                    )
+                );
+            }
         }
 
         // localClassName.name = className
```
ark120202: Unnamed default class declarations should get `default` name

src/LuaTransformer.ts
```diff
@@ -132,12 +133,13 @@ export class LuaTransformer {
             this.popScope();
 
             if (this.isModule) {
+                const exportsTable = !this.visitedExportEquals ? tstl.createTableExpression() : undefined;
+
                 // local exports = {}
+                // or
+                // local exports
```
ark120202: Considering that `export =` can be used only once and only on top level, maybe it would be better to declare a local on the export assignment itself?

src/LuaTransformer.ts
```diff
@@ -238,140 +242,159 @@ export class LuaTransformer {
         return tstl.createDoStatement(statements, block);
     }
 
+    public transformExportAssignment(statement: ts.ExportAssignment): StatementVisitResult {
```
ark120202: It's ellidable, so needs
```ts
if (!resolver.isValueAliasDeclaration(statement)) {
	return undefined;
}
```

src/TSHelper.ts
```diff
@@ -53,6 +56,51 @@ export function getExtendedType(node: ts.ClassLikeDeclarationBase, checker: ts.T
     return extendedTypeNode && checker.getTypeAtLocation(extendedTypeNode);
 }
 
+export function getExportable(exportSpecifiers: ts.NamedExports, resolver: EmitResolver): ts.ExportSpecifier[] {
+    return exportSpecifiers.elements.filter(exportSpecifier => isExportable(exportSpecifier, resolver));
+}
+
+export function isExportable(exportSpecifier: ts.ExportSpecifier, resolver: EmitResolver): boolean {
```
ark120202: It's not used anywhere except `getExportable`, so maybe it would be better to inline it?

test/unit/require.spec.ts
```diff
@@ -187,3 +187,165 @@ test("ImportEquals declaration require", () => {
...
+            `,
+            "module.ts": classDeclarationStatement,
+        },
+        "value"
+    );
+
+    expect(result).toBe(true);
+});
+
+test("Class exported by default has name `default`", () => {
```
ark120202: I think it also would be nice to have a test that ensures that named class retains it's specified name. Maybe reuse cases from the previous test or even merge them?

src/LuaTransformer.ts
```diff
@@ -238,140 +247,163 @@ export class LuaTransformer {
...
+        } else {
+            // export * from "...";
+            return this.transformExportAllFrom(statement);
+        }
+    }
 
-            // Wrap in block to prevent imports from hoisting out of `do` statement
-            const block = ts.createBlock([importDeclaration]);
-            const result = this.transformBlock(block).statements;
+    public transformExportSpecifier(node: ts.ExportSpecifier): tstl.AssignmentStatement {
```
Perryvw: All public statement visitor methods should return StatementVisitResult

hazzard993: ExportSpecifier is a `Declaration`, not a `Statement`.

```ts
export { x as default, y }; // export { ExportSpecifier, ExportSpecifier };
export { x } from "default"; // export { ExportSpecifier } from "default";
```

Does the return type `tstl.AssignmentStatement` not make a lot of sense?

ark120202: It's not really a visitor in this case - currently we have only expression and statement visitors, so it should be `protected` (see `transformEntityName` for a similar case).

During #615 refactor I think we'll add a basic `transformNode` + runtime cast that would cover this better (`transformExpression` and `transformStatements` would just call `transformNode` and cast).

src/LuaTransformer.ts
```diff
@@ -380,83 +412,103 @@ export class LuaTransformer {
...
+
+        if (scope.importStatements) {
+            scope.importStatements.push(...result);
+            return undefined;
+        } else {
+            return result;
+        }
+    }
+
+    public transformImportSpecifier(
```
Perryvw: This should probably also be protected, similar to the other comment.

src/LuaTransformer.ts
```diff
@@ -3324,10 +3412,14 @@ export class LuaTransformer {
...
 
-    public transformClassExpression(expression: ts.ClassExpression): ExpressionVisitResult {
+    public transformClassExpression(expression: ts.ClassLikeDeclaration): ExpressionVisitResult {
+        const isDefaultExport = tsHelper.hasDefaultExportModifier(expression.modifiers);
+
         const className =
             expression.name !== undefined
                 ? this.transformIdentifier(expression.name)
+                : isDefaultExport
+                ? this.createDefaultExportIdentifier(expression)
```
Perryvw: I don't like nested ternary conditions, or how this is formatted. In case of this nesting I'd expect an extra indent on the cases of the second conditional. But it might be nicer to refactor this to just an if statement or maybe another method.

src/LuaTransformer.ts
```diff
@@ -5162,38 +5262,6 @@ export class LuaTransformer {
         }
     }
 
-    protected getAbsoluteImportPath(relativePath: string): string {
```
Perryvw: Any reason you moved these? I think they were located correctly in this file.

hazzard993: I mainly moved them to tsHelper as it was logic that _transpileAndExecuteProjectReturningMainExport_ needed.

The methods were not available to convert TypeScript path strings to Lua import and export path strings without constructing a LuaTransformer only to convert a string using those methods.

PR #671, opened 2019-07-17
Fixed referencing named function expressions

### LINKED ISSUE #669: Function expressions can have names. This is not accounted for
```ts
const y = function x() {
    x();
}
```

```lua
local y
y = function(self)
    x(_G)
end
```

[Playground Link](https://typescripttolua.github.io/play.html#src=const%20y%20%3D%20function%20x()%20%7B%0A%20%20%20%20x()%3B%0A%7D%0A)

tomblind:
fixes #669

```ts
const y = function x() {
    x();
}
```
=>
```lua
local y
y = (function()
    local function x(self)
        x(_G)
    end
    return x
end)()
```

### DIFF
--- a/src/LuaTransformer.ts
+++ b/src/LuaTransformer.ts
@@ -3679,7 +3679,7 @@ export class LuaTransformer {
             const symbol = this.checker.getSymbolAtLocation(node.name);
             if (symbol) {
                 const symbolId = this.symbolIds.get(symbol);
-                if (symbolId && scope.referencedSymbols.has(symbolId)) {
+                if (symbolId !== undefined && scope.referencedSymbols.has(symbolId)) {
                     const nameIdentifier = this.transformIdentifier(node.name);
                     return this.createImmediatelyInvokedFunctionExpression(
                         [tstl.createVariableDeclarationStatement(nameIdentifier, functionExpression)],
--- a/src/LuaTransformer.ts
+++ b/src/LuaTransformer.ts
@@ -3675,10 +3675,12 @@ export class LuaTransformer {
             node
         );
 
+        //Handle named function expressions which reference themselves
         if (ts.isFunctionExpression(node) && node.name && scope.referencedSymbols) {
             const symbol = this.checker.getSymbolAtLocation(node.name);
             if (symbol) {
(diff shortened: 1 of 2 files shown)


src/LuaTransformer.ts
```diff
@@ -3664,16 +3664,32 @@ export class LuaTransformer {
...
             spreadIdentifier,
             flags,
             node
         );
+
+        if (ts.isFunctionExpression(node) && node.name && scope.referencedSymbols) {
+            const symbol = this.checker.getSymbolAtLocation(node.name);
+            if (symbol) {
+                const symbolId = this.symbolIds.get(symbol);
+                if (symbolId && scope.referencedSymbols.has(symbolId)) {
```
```suggestion
                if (symbolId !== undefined && scope.referencedSymbols.has(symbolId)) {
```

It can't happen in practice, since `genSymbolIdCounter` always starts from `1`, but I think it's better to keep things safe when checking numbers and strings.

We probably should enable `strict-boolean-expressions` after we'll migrate to eslint.

tomblind: Good catch. I'm definitely in favor of enabling `strict-boolean-expressions`.

src/LuaTransformer.ts
```diff
@@ -3664,16 +3664,32 @@ export class LuaTransformer {
...
+        const functionExpression = tstl.createFunctionExpression(
             tstl.createBlock(transformedBody),
             paramNames,
             dotsLiteral,
             spreadIdentifier,
             flags,
             node
         );
+
+        if (ts.isFunctionExpression(node) && node.name && scope.referencedSymbols) {
```
Perryvw: Would probably not be a bad idea if y ou added a comment above this block saying this handles named function assignments.

src/LuaTransformer.ts
```diff
@@ -3664,16 +3664,32 @@ export class LuaTransformer {
...
+            const symbol = this.checker.getSymbolAtLocation(node.name);
+            if (symbol) {
+                const symbolId = this.symbolIds.get(symbol);
+                if (symbolId !== undefined && scope.referencedSymbols.has(symbolId)) {
```
Perryvw: It is not clear to me what the purpose of this check is.

PR #675, opened 2019-07-20
Process a union or intersection type of all strings as a string

martinjlowm:
Prior to this PR, the following snippet would fail to transpile `toUpperCase` to `upper`.

Now, all subtypes of a union is checked and the following is transpiled from:
```
const union: string = "hello test";
if (union === 'hello' || union === 'test' || union === 'hello test') {
    return union.toUpperCase();
}
```
to
```
local union = "hello test"
if (union == 'hello' or union == 'test' or union == 'hello test') then
    return string.upper(union)
end
```

### DIFF
--- a/src/LuaTransformer.ts
+++ b/src/LuaTransformer.ts
@@ -3918,10 +3918,8 @@ export class LuaTransformer {
             });
         }
 
-        switch (ownerType.flags) {
-            case ts.TypeFlags.String:
-            case ts.TypeFlags.StringLiteral:
-                return this.transformStringCallExpression(node);
+        if (tsHelper.isStringType(ownerType, this.checker, this.program)) {
+            return this.transformStringCallExpression(node);
         }
 
         // if ownerType is a array, use only supported functions
--- a/test/unit/string.spec.ts
+++ b/test/unit/string.spec.ts
@@ -349,6 +349,22 @@ test.each([`"foobar".length`, `"foobar".repeat(2)`, "`foo${'bar'}`.length", "`fo
     }
 );
 
+test("scoped string-union inference", () => {
+    const inp = "foo";
+
+    const result = util.transpileAndExecute(`
+        const union: string = "${inp}";
+
+        if (union === "foo" || union === "bar") {
+            return union.length;
+        }
+
+        return 0;
+    `);
+
+    expect(result).toBe(inp.length);
+});
+
 test.each([
     "function generic<T extends string>(string: T)",
     "type StringType = string; function generic<T extends StringType>(string: T)",
--- a/src/TSHelper.ts
+++ b/src/TSHelper.ts
@@ -140,6 +140,14 @@ export function isStringType(type: ts.Type, checker: ts.TypeChecker, program: ts
         }
     }
 
+    if (type.isUnion()) {
+        return type.types.every(t => isStringType(t, checker, program));
+    }
+
+    if (type.isIntersection()) {
+        return type.types.some(t => isStringType(t, checker, program));
+    }
+
     return (
         (type.flags & ts.TypeFlags.String) !== 0 ||
         (type.flags & ts.TypeFlags.StringLike) !== 0 ||


src/LuaTransformer.ts
```diff
@@ -3895,6 +3895,14 @@ export class LuaTransformer {
         }
 
         switch (ownerType.flags) {
+            case ts.TypeFlags.Union:
```
Perryvw: Would be much cleaner if you replaced this switch statement with a `tsHelper.isStringType` call (already exists). You can expand that function to deal with union/intersection types. Something like this:
```ts
export function isStringType(type: ts.Type): boolean {
    if (type.flags & ts.TypeFlags.Union) {
        return (type as ts.UnionType).types.every(isStringType);
    }

    if (type.flags & ts.TypeFlags.Intersection) {
        return (type as ts.IntersectionType).types.some(isStringType);
    }

    return (
        (type.flags & ts.TypeFlags.String) !== 0 ||
        (type.flags & ts.TypeFlags.StringLike) !== 0 ||
        (type.flags & ts.TypeFlags.StringLiteral) !== 0
    );
}
```

ark120202: There is a `ts.isUnionOrIntersection` function that can make it even simpler:

```ts
export function isStringType(type: ts.Type, checker: ts.TypeChecker, program: ts.Program): boolean {
    if (type.symbol) {
        const baseConstraint = checker.getBaseConstraintOfType(type);
        if (baseConstraint) {
            return isStringType(baseConstraint, checker, program);
        }
    }

    if (type.isUnionOrIntersection()) {
        return type.types.some(t => isStringType(t, checker, program));
    }

    return (
        (type.flags & ts.TypeFlags.String) !== 0 ||
        (type.flags & ts.TypeFlags.StringLike) !== 0 ||
        (type.flags & ts.TypeFlags.StringLiteral) !== 0
    );
}
```

test/unit/string.spec.ts
```diff
@@ -233,6 +233,17 @@ test.each(["hello test"])("string.toUpperCase (%p)", inp => {
     expect(result).toBe(inp.toUpperCase());
 });
 
+test.each(["hello", "test", "hello test"])("string.toUpperCase (union) (%p)", inp => {
```
ark120202: 1. This test isn't really related to string.toUpperCase, but to general array string detection, so it should be be placed before `string constrained generic foreach (%p)`  and be renamed
2. There's no need for test.each, since it's not the main reason for this test
3. The code inside template is indented with 2 spaces instead of 4
4. Variables shouldn't begin with underscore, typescript uses it for it's unused diagnostics
5. I don't see much reason to over-complicate test code with inference logic, something like this would be enough to prove test's point:

```ts
const union: "foo" | "bar" = "foo";
return union.length;
```

martinjlowm: 1. I was mainly concerned with `string.*` being transpiled correctly, hence I thought it made sense to have it around here. I'll move it :)
2. I guess you are right.
3. Woopsie 👍 
4. Gotcha - fixed in the next push.
5. Your proposal doesn't cover this. `union` here is just a string literal which already supports `.length`. Enforcing union to be a string type triggers the type guard which appears to compare the string to a union of each or-condition. Inside the if-block, `union` is now inferred as an actual union of strings.
    ```
    const union: string = "foo";

    if (union === "foo" || union === "bar") {
        return union.length;
    }

    return 0;
    ```
    ^ does the trick, though!

martinjlowm:
I've just pushed new changes based off of your feedback - thanks :)

test/unit/string.spec.ts
ark120202: Oh, looks like TS ignores type annotation in this case and only something like `const union = "foo" as "foo" | "bar";` works

test/unit/string.spec.ts
```diff
@@ -349,6 +349,20 @@ test.each([`"foobar".length`, `"foobar".repeat(2)`, "`foo${'bar'}`.length", "`fo
     }
 );
 
+test.each(["foo"])("scoped string-union inference (%p)", inp => {
```
ark120202: `.each` with a single case can be removed completely and be inlined to code.

martinjlowm: Done :)

PR #677, opened 2019-07-21
Assignment binding patterns

### LINKED ISSUE #574: Binding Pattern Assignment Statements
The AST for this a bit odd

`[ObjectLiteralExpression] = [Expression]` = Syntax error
`([ObjectLiteralExpression] = [Expression])` = left is treated like an object binding pattern
`[ArrayLiteralExpression] = [Expression]` = left is treated like an array binding pattern
`([ObjectLiteralExpression | ArrayLiteralExpression]) = [Expression]` Syntax error

```ts
// assigns load() to love.load, update() to love.update and draw() to love.draw
({ load: love.load, update: love.update, draw: love.draw } = {

    load() {
        print("Load");
    },

    update(dt) {
        print("Update")
    },

    draw() {
        print("Draw");
    }

});
```

tsc output
```js
var _a;
(_a = {
    load: function () {
        print("Load");
    },
    update: function (dt) {
        print("Update");
    },
    draw: function () {
        print("Draw");
    }
}, love.load = _a.load, love.update = _a.update, love.draw = _a.draw);
```

tstl output (v0.19.0)
```lua
local ____ = ((function()
    { -- fails here
        load = love.load,
        update = love.update,
        draw = love.draw,
    } = {
        load = function(self)
            print("Load")
        end,
        update = function(self, dt)
            print("Update")
        end,
        draw = function(self)
            print("Draw")
        end,
    }
    return {
        load = love.load,
        update = love.update,
        draw = love.draw,
    }
end)())
```

### LINKED ISSUE #599: Destructuring pattern assignments don't use special assignment behaviour
Two ways to assign an array's length:

```ts
array.length = 0;
[array.length] = [0];
```

The first results in the `ArraySetLength` LuaLib being used to correctly shorten the array. The second becomes `#array = 0` which doesn't run.

Thinking a prior transformation may need to convert a destructuring pattern to TypeScript assignment statements.

hazzard993:
Closes #574

Closes #599

Before destructures were only made with a Lua statement.

```lua
a, b, c = 1, 2, 3
```

This still happens as an optimisation but if the left hand side contains initializers, special assignments or nested destructuring patterns it reverts to an alternate method similar to how binding patterns currently work but separated into more methods.

```lua
local ____ = ...
x = ____[1][1]
...
```


ark120202:
This PR doesn't appear to do anything with variables declarations and `transformBindingPattern`, Is it impossible to share some logic with it?

hazzard993:
The destructuring assignment AST shares no kinds of nodes with binding patterns.

I tried it initially but I kept it out of scope as I ended up overhauling all binding and destructuring behaviour which wasn't my intention.

src/LuaTransformer.ts
```diff
@@ -3039,7 +3061,7 @@ export class LuaTransformer {
         this.validateFunctionAssignment(expression.right, rightType, leftType);
         this.validatePropertyAssignment(expression);
 
-        if (tsHelper.isArrayLengthAssignment(expression, this.checker, this.program)) {
+        if (tsHelper.isArrayLength(expression.left, this.checker, this.program)) {
```
tomblind: Can we remove this here since it's called in `transformAssignment` now?

src/LuaTransformer.ts
```diff
@@ -3051,32 +3073,283 @@ export class LuaTransformer {
...
+                case ts.SyntaxKind.ArrayLiteralExpression:
+                    result.push(
+                        ...this.transformArrayLiteralAssignmentPattern(
+                            element as ts.ArrayLiteralExpression,
+                            indexedRoot
+                        )
+                    );
+                    break;
+                case ts.SyntaxKind.BinaryExpression:
+                    const assignedVariable = tstl.createIdentifier("____bindingAssignmentValue");
```
tomblind: I think we need a test for this case (I see one for assignments in object destructuring, but not arrays)

src/LuaTransformer.ts
```diff
@@ -3085,7 +3358,7 @@ export class LuaTransformer {
         const leftType = this.checker.getTypeAtLocation(expression.left);
         this.validateFunctionAssignment(expression.right, rightType, leftType);
 
-        if (tsHelper.isArrayLengthAssignment(expression, this.checker, this.program)) {
+        if (tsHelper.isArrayLength(expression.left, this.checker, this.program)) {
```
tomblind: Same as above - this shouldn't be needed here now, right?

hazzard993: I think this one is optimal. This turns an array length assignment into a LuaLib call instead of creating an iife that returns that LuaLib call's result.

test/unit/bindingpatterns.spec.ts
```diff
@@ -18,6 +18,7 @@ const testCases = [
 const testCasesDefault = [
     { bindingString: "{x = true}", objectString: "{}", returnVariable: "x" },
     { bindingString: "{x, y = true}", objectString: "{x: false}", returnVariable: "y" },
+    { bindingString: "[x = true, y = false]", objectString: "[undefined, undefined]", returnVariable: "x" },
```
tomblind: We should test the case where `objectString` has values in it as well.

test/unit/bindingpatterns.spec.ts
```diff
@@ -82,3 +83,103 @@ test.each([
...
+);
+
+test.each([
+    { bindingString: "{ x: x.prop = true } = {}", returnValue: "x.prop", expectedResult: true },
+    {
+        bindingString: "{ x: x.prop = true } = {}",
+        returnValue: "typeof y === 'object'",
+        expectedResult: true,
+    },
+])("Binding pattern assignment pass-through (%p)", ({ bindingString, returnValue, expectedResult }) => {
```
tomblind: Similar to the above, it would be good to test the case where the object literal has a value for `x`.

src/LuaTransformer.ts
```diff
@@ -3043,44 +3065,283 @@ export class LuaTransformer {
...
+            return tstl.createAssignmentStatement(left, right, node);
+        }
+
+        throw TSTLErrors.NonFlattenableDestructure(node);
+    }
+
+    protected transformDestructuringAssignment(
+        node: ts.DestructuringAssignment,
+        root: tstl.Expression
+    ): tstl.Statement[] {
```
Perryvw: Will need to discuss with @ark120202 and @tomblind if they are okay with this since this seems a major workaround. `ts.DestructuringAssignment` is a type of expression, so ideally this would be a public transform visitor returning a ExpressionVisitResult. I do however understand that without the preceding statements rewrite this might be pretty much impossible.

PR #679, opened 2019-07-23
Fix transform `string.indexOf` with out `plain` parameter.

lightrabbit:


src/LuaTransformer.ts
```diff
@@ -4352,7 +4352,14 @@ export class LuaTransformer {
             case "indexOf":
                 const stringExpression =
```
ark120202: Since both expressions now have same arity they can be collapsed to a single call
```ts
this.createStringCall(
    "find",
    node,
    caller,
    params[0],
    params[1] ? this.expressionPlusOne(params[1]) : tstl.createNilLiteral(),
    tstl.createBooleanLiteral(true)
)
```

PR #685, opened 2019-07-24
Simplify and improve enum transform

### LINKED ISSUE #683: Enum values shouldn't be assumed to be pure
Input:
```ts
let i = 0;
const sideEffect = () => i++;
enum Foo {
    X = sideEffect(),
    Y = X,
}
```

Current Result:
```lua
local i = 0
local sideEffect
sideEffect = function() return (function()
    local ____TS_tmp = i
    i = ____TS_tmp + 1
    return ____TS_tmp
end)() end
Foo = {}
Foo.X = sideEffect(_G)
Foo[sideEffect(_G)] = "X"
Foo.Y = sideEffect(_G)
Foo[sideEffect(_G)] = "Y"
```

Expected Result:
```lua
local i = 0
local sideEffect
sideEffect = function() return (function()
    local ____TS_tmp = i
    i = ____TS_tmp + 1
    return ____TS_tmp
end)() end
Foo = {}
Foo.X = sideEffect(_G)
Foo[Foo.X] = "X"
Foo.Y = Foo.X
Foo[Foo.Y] = "Y"
```

### LINKED ISSUE #684: Const enums with own member references are emitted
Input:
```ts
const enum Foo {
    A,
    B = A,
}
```

Current Result:
```lua
Foo = {}
Foo.A = 0
Foo[0] = "A"
Foo.B = 0
Foo[0] = "B"
```

Expected Result:
```lua
```

ark120202:
Fixes #683, fixes #684.
Fixes string literal references:
```ts
enum Foo {
    ["A"],
    B = A,
}
```

Tests depend on #609.


src/LuaTransformer.ts
```diff
@@ -1727,115 +1727,77 @@ export class LuaTransformer {
...
-                    }
+                if (!valueExpression) {
+                    valueExpression = this.transformExpression(member.initializer);
                 }
-            } else if (hasStringInitializers) {
-                throw TSTLErrors.HeterogeneousEnum(node);
             } else {
-                valueExpression = tstl.createNumericLiteral(numericValue);
-                numericValue++;
+                valueExpression = tstl.createNilLiteral();
```
Perryvw: Looking at this again, is this actually correct? In what case would an enum value be nil?

ark120202: Original transformer has the same branch: https://github.com/microsoft/TypeScript/blob/master/src/compiler/transformers/ts.ts#L2442.
I have thought it'd be used for some syntactically incorrect enums, but couldn't find any cases where TS goes there.

PR #688, opened 2019-07-25
Wrap expression statements based on transformed node

ark120202:
Instead of wrapping everything except for call-like source nodes now it wraps everything except transformed call nodes. Because of that when call transform returns non-call expression it would be handled correctly now, so [string indexOf statements](https://typescripttolua.github.io/play.html#src=%22%22.indexOf(%22%22)) are valid now and extra logic in `Math.log10` and `Math.log2` is no longer needed.


test/unit/expressions.spec.ts
```diff
@@ -97,37 +97,37 @@ test.each([
...
+    { input: "a^b", lua: "bit32.bxor(a, b)" },
     { input: "a^=b", lua: "a = bit32.bxor(a, b)" },
-    { input: "a<<b", lua: "local ____ = bit32.lshift(a, b)" },
+    { input: "a<<b", lua: "bit32.lshift(a, b)" },
     { input: "a<<=b", lua: "a = bit32.lshift(a, b)" },
-    { input: "a>>b", lua: "local ____ = bit32.arshift(a, b)" },
+    { input: "a>>b", lua: "bit32.arshift(a, b)" },
     { input: "a>>=b", lua: "a = bit32.arshift(a, b)" },
-    { input: "a>>>b", lua: "local ____ = bit32.rshift(a, b)" },
+    { input: "a>>>b", lua: "bit32.rshift(a, b)" },
```
Perryvw: I don't think this is valid lua without the assignment?

ark120202: Even though it's pure it's just a function call which is valid

PR #689, opened 2019-07-25
@noSelf changed to not affect merged namespaces

### LINKED ISSUE #686: noSelf for declaration merging
```

class A {
    functionA(): void {
        
    }
}

/**@noSelf */
module A {
    function functionB() : void {

    }
}
```
Expected :
functionA with self argument.
functionB without self argument.

tomblind:
fixes #686

Previously, if a namespace or class was decorated with `@noSelf`, this affected all declarations merged with it. This changes that behavior so, for example, methods in a class merged with a `@noSelf` decorated namespace will retain their `self` parameter.

Example:
```ts
declare class A {
    method(s: string): void;
}

/** @noSelf */
declare namespace A {
    export function noSelf(s: string): void;
}

declare namespace A {
    export function hasSelf(s: string): void;
}

const c = new A();
c.method("foo");
A.noSelf("bar");
A.hasSelf("baz");
```
=>
```lua
local c = A.new()
c:method("foo")
A.noSelf("bar")
A:hasSelf("baz")
```

### DIFF
--- a/test/unit/assignments/functionPermutations.ts
+++ b/test/unit/assignments/functionPermutations.ts
@@ -133,6 +133,22 @@ export const selfTestFunctions: TestFunction[] = [
         }
         const anonFunctionNestedInNoSelfClass = (new AnonFunctionNestedInNoSelfClass).method();`,
     },
+    {
+        value: "anonMethodClassMergedNoSelfNS.method",
+        definition: `class AnonMethodClassMergedNoSelfNS { method(s: string): string { return s; } }
+            /** @noSelf */ namespace AnonMethodClassMergedNoSelfNS { export function nsFunc(s: string) { return s; } }
+            const anonMethodClassMergedNoSelfNS = new AnonMethodClassMergedNoSelfNS();`,
+    },
+    {
+        value: "AnonFuncNSMergedNoSelfClass.nsFunc",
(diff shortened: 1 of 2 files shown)


test/unit/assignments/functionPermutations.ts
```diff
@@ -133,6 +133,22 @@ export const selfTestFunctions: TestFunction[] = [
         }
         const anonFunctionNestedInNoSelfClass = (new AnonFunctionNestedInNoSelfClass).method();`,
     },
+    {
+        value: "anonMethodClassMergedNoSelfNS.method",
+        definition: `class AnonMethodClassMergedNoSelfNS { method(s: string): string { return s; } }
+            /** @noSelf */ namespace AnonMethodClassMergedNoSelfNS { export function nsFunc(s: string) { return s; } }
```
Perryvw: Do these actually fail without the transformer change?

tomblind: Yes, because the class method will have it's `self` stripped due to the namespace using `@noSelf`

Perryvw: But the call to the function will adjust accordingly?

tomblind: Yes, but the assignment test would fail since it would try to assign the method to a type with a `self` parameter. Also, those tests actually call the function and verify parameters are in correct order. That could probably be moved out to separate tests, but that seems out of scope for this PR.

PR #690, opened 2019-07-25
Allow specifying transformer in tsconfig.json

danielpza:
- [x] old behavior if transformer is not specified 
- [x] tsconfig option: `tstl.luaTransformer`
- [ ] tests

related #615

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1887,7 +1887,8 @@
                 "ansi-regex": {
                     "version": "2.1.1",
                     "bundled": true,
-                    "dev": true
+                    "dev": true,
+                    "optional": true
                 },
                 "aproba": {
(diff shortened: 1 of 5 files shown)


ark120202:
I'm not entirely convinced that we should to expose it in configuration, but in case we'll do there are some comments on the implementation:

1. `tstl.ts` is a consumer of API so it shouldn't contain such logic. All custom options should be processed in the internal API (`tstl.transpile`).
2. It should reuse loading and resolution logic with ts transformers (resolve it from tsconfig directory instead of cwd and load .ts files with ts-node)

Perryvw:
This is covered by the compiler API, I don't see any use for this at the moment.

PR #693, opened 2019-07-28
SpreadAssignment support

### LINKED ISSUE #121: Spread operator
It would be nice to be able to use the spread operator. For arrays, it's pretty straightforward:

```typescript
const example = [ "a", ...[ "b", "c" ] ];
print(...example);
```

can be compiled to

```lua
local example = { "a", table.unpack("b", "c" }
print(table.unpack(example))
```

Objects, on the other hand, can't easily be implemented in a way that emulates their behavior in JS - the easiest way to implement it might be to add a function to `typescript_lualib`, so that something like this could be done:

```typescript
const example = { a: 1, ...{ a: 2, b: 3 } }
```
...might be compiled to...
```lua
-- typescript_lualib.lua
function TS_spreadobj(a, b)
	local c = {}

	for k,v in pairs(a) do c[k] = v end
	for k,v in pairs(b) do c[k] = v end

	return c
end

-- somewhere else
local example = TS_spreadobj({a = 1}, {a = 2, b = 3})
```

This introduces a bit of extra complexity - we'd need to separate the literal keys in the object from the spread operator and transpile them to separate expressions, and finally wrap it all in a call.

hazzard993:
Closes #121

SpreadAssignment elements can be used in object literal expressions.

```ts
const example = { a: 1, ...{ a: 2, b: 3 } };
```

```lua
local example = __TS__ObjectAssign({a = 1}, {a = 2, b = 3})
```

Strings can also be spread.

```ts
print(..."string"); // --> s t r i n g
```


src/lualib/MergeObjects.ts
```diff
@@ -0,0 +1,11 @@
+function __TS__MergeObjects(this: void, ...objects: object[]): object {
```
ark120202: Can't it reuse `Object.assign`?

src/lualib/Spread.ts
```diff
@@ -1,7 +1,13 @@
 function __TS__Spread<T>(this: void, iterable: Iterable<T>): T[] {
-    const arr: T[] = [];
-    for (const item of iterable) {
-        arr[arr.length] = item;
+    const arr = [];
+    if (typeof iterable === "string") {
+        for (const [character] of string.gmatch(iterable, ".")) {
```
ark120202: I assume it would be significantly slower than a regular for loop. Is some reason why it wouldn't work?

src/LuaTransformer.ts
```diff
@@ -3853,12 +3854,26 @@ export class LuaTransformer {
...
         });
 
-        return tstl.createTableExpression(properties, expression);
+        if (tableExpressions.length === 0) {
+            return tstl.createTableExpression(properties, expression);
+        } else {
+            const tableExpression = tstl.createTableExpression(properties, expression);
+            tableExpressions.push(tableExpression);
+
+            return this.transformLuaLibFunction(LuaLibFeature.ObjectAssign, expression, ...tableExpressions);
```
ark120202: `Object.assign` should always get a table literal as a first argument to avoid mutation of first spread element value. It probably would be better to check the type of the first element of `tableExpressions` to handle literal spread cases (check out [TypeScript output](https://www.typescriptlang.org/play/#code/BQbwBAdF4G4IYBsCuBTAXGALgJ1WAvgQJQDcAUKGPMumAGaIDOKANJNFYqhjnofqQrgoEWF1oMEzAm2rcsuFMXKURYmhknT8s8Zqat2ozhoV9lZMgGMA9gDtGmMNhR0wAXjAh8K9fN6GIi5uAr5GwRZAA)).

src/LuaTransformer.ts
```diff
@@ -3853,12 +3854,26 @@ export class LuaTransformer {
...
             } else {
                 throw TSTLErrors.UnsupportedKind("object literal element", element.kind, expression);
             }
         });
 
-        return tstl.createTableExpression(properties, expression);
+        if (tableExpressions.length === 0) {
+            return tstl.createTableExpression(properties, expression);
+        } else {
+            const tableExpression = tstl.createTableExpression(properties, expression);
```
ark120202: It would be nice to avoid adding empty table when `properties` is empty.

src/LuaTransformer.ts
hazzard993: Really good edge cases, thanks!

test/unit/spreadElement.spec.ts
```diff
@@ -64,3 +64,30 @@ test("Spread Element Iterable", () => {
...
+test.each([
+    "{ ...{ value: true } }",
+    "{ value: false, ...{ value: true } }",
+    "{ ...{ value: false }, value: true }",
+    "{ ...{ value: false }, value: false, ...{ value: true } }",
+])('SpreadAssignment "%s"', expression => {
+    const code = `
+        const obj = ${expression};
+        return obj.value;`;
+    expect(JSON.parse(util.transpileAndExecute(code))).toBe(true);
```
Perryvw: The execution result here should already be a boolean, so there is no reason to JSON.parse here.

test/unit/spreadElement.spec.ts
```diff
@@ -64,3 +64,30 @@ test("Spread Element Iterable", () => {
...
+        const arr = [..."${str}"];
+        return JSONStringify(arr)`;
+    expect(JSON.parse(util.transpileAndExecute(code))).toEqual([...str]);
+});
+
+test.each([
+    "{ ...{ value: true } }",
+    "{ value: false, ...{ value: true } }",
+    "{ ...{ value: false }, value: true }",
+    "{ ...{ value: false }, value: false, ...{ value: true } }",
```
Perryvw: Could you add a basic test where you destructure multiple different properties with different values and then expect JSON.parse toEqual?

Perryvw: Also some where you have multiple destructures adding different properties to the same object.

src/LuaTransformer.ts
```diff
@@ -3853,12 +3854,32 @@ export class LuaTransformer {
             } else if (ts.isMethodDeclaration(element)) {
                 const expression = this.transformFunctionExpression(element);
                 properties.push(tstl.createTableFieldExpression(expression, name, element));
+            } else if (ts.isSpreadAssignment(element)) {
+                if (properties.length > 0) {
+                    const tableExpression = tstl.createTableExpression(properties, expression);
+                    tableExpressions.push(tableExpression);
+                }
+                properties = [];
```
Perryvw: You already handle this situation (line 3858-3862) in lines 3872-3875. I think you can remove these 5 lines.

test/unit/spreadElement.spec.ts
Perryvw: And I would also like to see a test where you destructure an identifier of a previously made object, rather than an object literal.

ark120202:
It's also possible to spread arrays: `{ ...[0, 1, 2] }`, which should return `{ 0: 0, 1: 1, 2: 2 }`. It shouldn't be done in `Object.assign` since it has a different return type, so we probably need a new helper that would shift indexes of arrays.

src/LuaTransformer.ts
hazzard993: Alright so the current algorithm creates multiple tables in the order the properties are declared so if they are overwritten they are overwritten in the correct order.

Source

```ts
let obj = { x: 0, y: 1, z: 2 };
const object = { x: 0, ...obj, y: 1, z: 2 }
const o = { x: true, ...{ y: true, z: true } };
```

Output with this code:

```diff
  local obj = {x = 0, y = 1, z = 2}
+ local object = __TS__ObjectAssign({x = 0}, obj, {y = 1, z = 2})
+ local o = __TS__ObjectAssign({x = true}, {y = true, z = true})
```

Output without:

```diff
  local obj = {x = 0, y = 1, z = 2}
- local object = __TS__ObjectAssign({}, obj, {x = 0, y = 1, z = 2})
- local o = __TS__ObjectAssign({y = true, z = true}, {x = true})
```

Perryvw: I see, that makes sense. Could you place a comment on the first occurrence explaining that? Something like `// Create table for preceding properties to preserve order of elements.`?

src/LuaTransformer.ts
```diff
@@ -3853,12 +3854,44 @@ export class LuaTransformer {
...
+                // Create a table for preceding properties to preserve property order
+                // { x: 0, ...{ y: 2 }, y: 1, z: 2 } --> __TS__ObjectAssign({x = 0}, {y = 2}, {y = 1, z = 2})
+                if (properties.length > 0) {
+                    const tableExpression = tstl.createTableExpression(properties, expression);
+                    tableExpressions.push(tableExpression);
+                }
+                properties = [];
+
+                const type = this.checker.getTypeAtLocation(element.expression);
+                let tableExpression = this.transformExpression(element.expression);
```
Perryvw: Put this in the else branch of the following if, or behind a ternary conditional, I'm a bit wary of doing work twice after a major performance bug we solved recently.

test/unit/spreadElement.spec.ts
```diff
@@ -64,3 +64,53 @@ test("Spread Element Iterable", () => {
...
+    "{ ...{ x: true, y: true } }",
+    "{ x: true, y: true }",
```
ark120202: What's the point of a test without any spread elements?

test/unit/spreadElement.spec.ts
```diff
@@ -64,3 +64,53 @@ test("Spread Element Iterable", () => {
...
+        return JSONStringify(arr)`;
+    expect(JSON.parse(util.transpileAndExecute(code))).toEqual([...str]);
+});
+
+test.each([
+    "{ ...{ value: true } }",
+    "{ value: false, ...{ value: true } }",
+    "{ ...{ value: false }, value: true }",
+    "{ ...{ value: false }, value: false, ...{ value: true } }",
+    "{ ...{ x: true, y: true } }",
```
ark120202: Do we really need a single-property test if we have a multi-property one?

PR #700, opened 2019-08-11
class field initialization order

### LINKED ISSUE #697: Class properties should be initialized after constructor parameter properties
Input:
```ts
class Test {
  public foo = this.bar;
  constructor(public bar: string) {}
}
```

Current Result:
```lua
-- ...
function Test.prototype.____constructor(self, bar)
    self.foo = self.bar
    self.bar = bar
end
```

Expected Result:
```lua
-- ...
function Test.prototype.____constructor(self, bar)
    self.bar = bar
    self.foo = self.bar
end
```

tomblind:
fixes #697

This PR also addresses issues which arose with code like this:
```ts
class Test {
    constructor(public bar = true) {}
}
const result = (new Test(false)).bar;
console.log(result);
```
=>
```lua
Test = {}
Test.name = "Test"
Test.__index = Test
Test.prototype = {}
Test.prototype.__index = Test.prototype
Test.prototype.constructor = Test
function Test.new(...)
    local self = setmetatable({}, Test.prototype)
    self:____constructor(...)
    return self
end
function Test.prototype.____constructor(self, bar)
    self.bar = bar or true
    if bar == nil then
        bar = true
    end
end
local result = Test.new(false).bar
print(result)
```

There are 2 problems here:
- `bar` will incorrectly be assign `true`
- The default value assignment to `bar` is done twice

These are corrected by placing default assignments above field initializers and removing the `or true` assignment to constructor fields:
```lua
function Test.prototype.____constructor(self, bar)
    if bar == nil then
        bar = true
    end
    self.bar = bar
end
```

### DIFF
--- a/src/LuaTransformer.ts
+++ b/src/LuaTransformer.ts
@@ -1237,10 +1237,7 @@ export class LuaTransformer {
             return undefined;
         }
 
-        const bodyWithFieldInitializers: tstl.Statement[] = this.transformClassInstanceFields(
-            classDeclaration,
-            instanceFields
-        );
+        const bodyWithFieldInitializers: tstl.Statement[] = [];
 
         // Check for field declarations in constructor
         const constructorFieldsDeclarations = statement.parameters.filter(p => p.modifiers !== undefined);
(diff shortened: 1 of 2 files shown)


src/LuaTransformer.ts
```diff
@@ -1294,11 +1267,27 @@ export class LuaTransformer {
             ) {
                 const superCall = body.shift();
                 if (superCall) {
-                    bodyWithFieldInitializers.unshift(superCall);
+                    bodyWithFieldInitializers.push(superCall);
```
Perryvw: This looks strange in combination with the comment on line 1256.

PR #702, opened 2019-08-16
Fix LuaTable

### LINKED ISSUE #699: LuaTable as object member
Currently output is broken when `LuaTable` objects are used as object members.

# Input

```typescript
/** @LuaTable */
declare class LuaTable<K extends {} = {}, V = any> {
  readonly length: number;
  set(key: K, value: V): void;
  get(key: K): V;
}

const k = "k"
const t = { data: new LuaTable() }
const v = t.data.get(k)
t.data.set(k,v)
```

# Output

```lua
local t = {data = {}}
local v = .k
local k = "k"
[k] = v
```

# Expected

```lua
local t = {data = {}}
local v = t.data.k
local k = "k"
t.data[k] = v
```

hazzard993:
Closes #699

Also added a rule preventing LuaTables from being accessed via ElementAccessExpressions.

```diff
  /** @LuaTable */
  declare class LuaTable<K extends {} = {}, V = any> {
    readonly length: number;
    set(key: K, value: V): void;
    get(key: K): V;
  }
  
  const k = "k"
  const t = { data: new LuaTable() }
+ const v = t.data.get(k) // --> local v = t.data[k]
+ t.data.set(k,v) // --> t.data[k] = v
+ const v2 = t.data.get(k).x; // --> local v2 = t.data[k].x
- const v3 = t.data["get"](k)
- t.data["set"](k, v)
```

Removed a lot of casting to track down the bug.

### DIFF
--- a/src/LuaTransformer.ts
+++ b/src/LuaTransformer.ts
@@ -4665,12 +4665,9 @@ export class LuaTransformer {
 
         const argumentType = this.checker.getTypeAtLocation(expression.argumentExpression);
         const type = this.checker.getTypeAtLocation(expression.expression);
-        const decorators = new Map();
-        if (type && type.symbol) {
-            tsHelper.collectCustomDecorators(type.symbol, this.checker, decorators);
-            if (decorators.has(DecoratorKind.LuaTable)) {
-                throw TSTLErrors.UnsupportedKind("LuaTable access expression", expression.kind, expression);
-            }
+        const decorators = tsHelper.getCustomDecorators(type, this.checker);
+        if (decorators.has(DecoratorKind.LuaTable)) {
+            throw TSTLErrors.UnsupportedKind("LuaTable access expression", expression.kind, expression);
         }
 
         if (
--- a/src/LuaTransformer.ts
+++ b/src/LuaTransformer.ts
@@ -5450,18 +5450,13 @@ export class LuaTransformer {
     }
 
     protected getLuaTablePropertyName(node: ts.LeftHandSideExpression): string {
-        let methodName: string;
-        switch (node.kind) {
-            case ts.SyntaxKind.PropertyAccessExpression:
-                methodName = (node as ts.PropertyAccessExpression).name.text;
-                break;
-            case ts.SyntaxKind.Identifier:
-                methodName = (node as ts.Identifier).text;
-                break;
-            default:
-                throw TSTLErrors.UnsupportedKind("LuaTable access expression", node.kind, node);
+        if (ts.isPropertyAccessExpression(node)) {
+            return node.name.text;
+        } else if (ts.isIdentifier(node)) {
+            return node.text;
+        } else {
+            throw TSTLErrors.UnsupportedKind("LuaTable access expression", node.kind, node);
         }
-        return methodName;
     }
 
     protected transformLuaLibFunction(
--- a/src/LuaTransformer.ts
+++ b/src/LuaTransformer.ts
@@ -2165,17 +2165,16 @@ export class LuaTransformer {
             );
         }
 
-        if (ts.isCallExpression(expression) && ts.isPropertyAccessExpression(expression.expression)) {
-            const ownerType = this.checker.getTypeAtLocation(expression.expression.expression);
+        if (
+            ts.isExpressionStatement(statement) &&
+            ts.isCallExpression(statement.expression) &&
+            ts.isPropertyAccessExpression(statement.expression.expression)
+        ) {
+            const ownerType = this.checker.getTypeAtLocation(statement.expression.expression.expression);
             const classDecorators = tsHelper.getCustomDecorators(ownerType, this.checker);
             if (classDecorators.has(DecoratorKind.LuaTable)) {
-                this.validateLuaTableCall(
-                    expression as ts.CallExpression & { expression: ts.PropertyAccessExpression },
-                    true
-                );
-                return this.transformLuaTableExpressionStatement(statement as ts.ExpressionStatement & {
-                    expression: ts.CallExpression;
-                } & { expression: { expression: ts.PropertyAccessExpression } });
+                this.validateLuaTableCall(statement.expression);
+                return this.transformLuaTableExpressionAsExpressionStatement(statement.expression);
             }
         }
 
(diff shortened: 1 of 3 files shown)


src/LuaTransformer.ts
```diff
@@ -2165,17 +2165,16 @@ export class LuaTransformer {
             );
         }
 
-        if (ts.isCallExpression(expression) && ts.isPropertyAccessExpression(expression.expression)) {
-            const ownerType = this.checker.getTypeAtLocation(expression.expression.expression);
+        if (
+            ts.isExpressionStatement(statement) &&
```
ark120202: Do you want to disallow it in statement-like places like `() => luaTable.set('x', 'y')`?

src/LuaTransformer.ts
```diff
@@ -4656,6 +4665,14 @@ export class LuaTransformer {
 
         const argumentType = this.checker.getTypeAtLocation(expression.argumentExpression);
         const type = this.checker.getTypeAtLocation(expression.expression);
+        const decorators = new Map();
```
ark120202: Why not just `getCustomDecorators`?

src/LuaTransformer.ts
```diff
@@ -4995,57 +5009,48 @@ export class LuaTransformer {
...
-    ): tstl.VariableDeclarationStatement | tstl.AssignmentStatement {
-        const methodName = node.expression.expression.name.text;
-        const signature = this.checker.getResolvedSignature(node.expression);
-        const tableName = (node.expression.expression.expression as ts.Identifier).text;
-        const luaTable = tstl.createIdentifier(tableName);
-        const params = this.transformArguments((node.expression as ts.CallExpression).arguments, signature);
+    protected transformLuaTableExpressionAsExpressionStatement(expression: ts.CallExpression): tstl.Statement {
+        const methodName = this.getLuaTablePropertyName(expression.expression);
+        const signature = this.checker.getResolvedSignature(expression);
+        const tableAccessExpression = this.transformExpression(expression.expression);
```
ark120202: `AccessExpression` can be quite misleading since it can be an identifier as well

src/LuaTransformer.ts
```diff
@@ -5444,6 +5449,21 @@ export class LuaTransformer {
         return scope;
     }
 
+    protected getLuaTablePropertyName(node: ts.LeftHandSideExpression): string {
+        let methodName: string;
```
ark120202: I don't think switch statement looks good there

```ts
if (ts.isPropertyAccessExpression(node)) {
    return node.name.text;
} else if (ts.isIdentifier(node)) {
    return node.text;
} else {
    throw TSTLErrors.UnsupportedKind("LuaTable access expression", node.kind, node);
}
```

src/LuaTransformer.ts
hazzard993: Could be nice to give it some more features, it there any other statement-like places you can think of?

ark120202: Oh, I have assumed it worked, but it makes sense that it doesn't since it has to return expression, forget about this case then (we'll be able to support it easily with multi-statement expressions, I don't think it worth the complexity for now).

But what I was asking is why you have changed this condition. The way it was written before allowed it to work in cases where `transformExpressionStatement` is called directly with an expression, for example in for statement.

hazzard993: It must've been left over from when I was removing the casting from the statements below. This change doesn't seem to disallow LuaTable CallExpressions to be used in for loops and doesn't seem to break anything according to the tests so I'll remove this conditional change

src/LuaTransformer.ts
```diff
@@ -4995,57 +5006,48 @@ export class LuaTransformer {
                         expression
                     );
                 }
-                if (!isWithinExpressionStatement) {
+                if (expression.parent.kind !== ts.SyntaxKind.ExpressionStatement) {
```
ark120202: That won't work if it's gone through `transformExpressionStatement` as an expression

hazzard993: Seems like that would cause a bug but I can't seem to create some code to break this. What kind of code would do you think would break this?

ark120202: Just like above, anything that calls `transformExpressionStatement` directly with an expression

src/LuaTransformer.ts
```diff
@@ -4619,8 +4620,12 @@ export class LuaTransformer {
         }
     }
 
-    protected transformLuaTableProperty(node: ts.PropertyAccessExpression): tstl.UnaryExpression {
+    protected transformLuaTableProperty(node: ts.PropertyAccessExpression): tstl.Expression {
         switch (node.name.text) {
+            case "set":
+                return this.transformExpression(node.expression);
```
ark120202: Okay, took me some time to understand how it works and I don't like how transformation flows there. In general, property access isn't equivalent to it's root node, so returning it there isn't correct.

Maybe something like `const [luaTable, methodName] = this.parseLuaTableExpression(expression.expression);`?

src/LuaTransformer.ts
hazzard993: It looks it is working as expected. It doesn't break when it goes through `transformExpressionStatement` as an Expression and will throw the Exception appropriately

```ts
/** @LuaTable */
declare class LuaTable<K extends {} = {}, V = any> {
  readonly length: number;
  set(key: K, value: V): number;
  get(key: K): V;
}
let y;
let t = new LuaTable();
for (let x = 0; x < 10; t.set("x", 4)) {} // ❌
for (let x = 0; x < 10; x = t.set("x", 4)) {} // ❌
let x = 4 + (y = t.set('', 2)); // ❌
let xx = 4 + t.set('', 2); // ❌
let xxx = t.set('', 2); // ❌
let y = 4 + (y = t.get('')); // ✔
let yy = 4 + t.get(''); // ✔
let yyy = t.get(''); // ✔
```

`transformExpressionStatement` is only called with an ExpressionStatement from `transformStatement` and this would be the only case where `luaTable.set` would be allowed

ark120202: It shouldn't throw in these cases. The whole point of passing an expression to `transformExpressionStatement` is to assume that this expression is used in statement context, even if it's not a part of ExpressionStatement.

src/LuaTransformer.ts
```diff
@@ -4477,7 +4477,8 @@ export class LuaTransformer {
         }
 
         if (decorators.has(DecoratorKind.LuaTable)) {
-            return this.transformLuaTableProperty(expression);
+            const [luaTable] = this.parseLuaTableExpression(expression);
```
ark120202: That's not what I meant. `table.get` shouldn't be transformed just to `table`, because these aren't interchangeable in isolation. `transformLuaTableProperty` should allow only `length` property.
What I meant is that `getLuaTablePropertyName` can be renamed to `parseLuaTableExpression` and would split lua table reference and it's method name.

hazzard993: Thanks for the clarification, looks like it neatened up nicely

PR #705, opened 2019-08-19
Transformation Pipeline Refactor

### LINKED ISSUE #559: Rename Decorator to Directive
Decorator is used internally in TS to mean something else

### LINKED ISSUE #615: Pluggable transformer design

ark120202:
Resolves #615, closes #559.


src/transformation/context/visitors.ts
```diff
@@ -0,0 +1,157 @@
...
+    : T extends ts.SourceFile
+    ? tstl.Block
+    : OneToManyVisitorResult<tstl.Node>;
+
+export type FunctionVisitor<T extends ts.Node> = (node: T, context: TransformationContext) => VisitorResult<T>;
+export type ObjectVisitor<T extends ts.Node> = { transform: FunctionVisitor<T>; priority?: number };
+export type Visitor<T extends ts.Node> = FunctionVisitor<T> | ObjectVisitor<T>;
+export type Visitors = { [P in keyof NodesBySyntaxKind]?: Visitor<NodesBySyntaxKind[P]> };
+
+export interface TransformerPlugin {
```
Perryvw: We could consider moving this to its own file if this interface grows, for now I don't know if there is a better location for this interface.

src/transformation/index.ts
```diff
@@ -0,0 +1,72 @@
...
+    messageText: error.message,
+});
+
+export function createVisitorMap(customPlugins: TransformerPlugin[]): VisitorMap {
+    const visitorMap: VisitorMap = new Map();
+    for (const plugin of [...standardPlugins, ...customPlugins]) {
+        for (const [syntaxKindKey, visitor] of Object.entries(plugin.visitors)) {
+            if (!visitor) continue;
+
+            const syntaxKind = Number(syntaxKindKey) as ts.SyntaxKind;
```
Perryvw: This number cast seems strange, is it because keys are typed as strings?

src/transformation/index.ts
```diff
@@ -0,0 +1,72 @@
...
+            const visitors = setIfMissing(visitorMap, syntaxKind, () => []);
+            visitors.push((typeof visitor === "function" ? { transform: visitor } : visitor) as any);
```
Perryvw: This seems like some magic, should we even allow this?

src/transformation/transformers/binary/destructuring-assignments.ts
```diff
@@ -0,0 +1,280 @@
...
+import * as tstl from "../../../LuaAST";
+import { flatMap } from "../../../utils";
+import { TransformationContext } from "../../context";
+import { UnsupportedKind } from "../../utils/errors";
+import { LuaLibFeature, transformLuaLibFunction } from "../../utils/lualib";
+import { isArrayType, isAssignmentPattern } from "../../utils/typescript";
+import { transformIdentifier } from "../identifier";
+import { transformPropertyName } from "../literal";
+import { transformAssignment, transformAssignmentStatement } from "./assignments";
+
```
Perryvw: This binary directory name really confused me for a bit there, I was thinking of binary as bits (since I saw the bitwise operators first), not binary expressions. Would be nice if this can be ranamed to something like binary-expressions. Taking that a bit broader, should we in general indicate whether a folders transformers concern expressions or assignments?

src/transformation/transformers/class/members/accessors.ts
```diff
@@ -0,0 +1,91 @@
...
+import { transformIdentifier } from "../../identifier";
+import { findInClassOrAncestor, isStaticNode } from "../utils";
+
+export function hasSetAccessorInClassOrAncestor(
+    context: TransformationContext,
+    classDeclaration: ts.ClassLikeDeclarationBase,
+    isStatic: boolean
+): boolean {
+    return (
+        findInClassOrAncestor(context, classDeclaration, c =>
```
Perryvw: Don't know if we can do anything about this, but this formatting looks strange. Maybe extract the callback to its own const?

src/transformation/transformers/class/members/accessors.ts
```diff
@@ -0,0 +1,91 @@
...
+    );
+}
+
+export function hasGetAccessorInClassOrAncestor(
+    context: TransformationContext,
+    classDeclaration: ts.ClassLikeDeclarationBase,
+    isStatic: boolean
+): boolean {
+    return (
+        findInClassOrAncestor(context, classDeclaration, c =>
```
Perryvw: Same as the previous function

src/transformation/transformers/class/new.ts
```diff
@@ -0,0 +1,64 @@
+import * as ts from "typescript";
+import * as tstl from "../../../LuaAST";
+import { FunctionVisitor, TransformationContext } from "../../context";
+import { DecoratorKind, getCustomDecorators } from "../../utils/decorators";
+import { InvalidDecoratorArgumentNumber, InvalidNewExpressionOnExtension } from "../../utils/errors";
+import { importLuaLibFeature, LuaLibFeature } from "../../utils/lualib";
+import { transformArguments } from "../call";
+
```
Perryvw: I think it's a little strange we have both class/creation and class/new, why are they not the same thing?

src/transformation/transformers/jump.ts
```diff
@@ -0,0 +1,42 @@
+import * as ts from "typescript";
+import { LuaTarget } from "../../CompilerOptions";
+import * as tstl from "../../LuaAST";
+import { FunctionVisitor, TransformerPlugin } from "../context";
+import { UndefinedScope, UnsupportedForTarget } from "../utils/errors";
+import { findScope, ScopeType } from "../utils/scope";
+
+// TODO: File name?
```
Perryvw: maybe just break-continue.ts?

src/transformation/utils/assignment-validation.ts
```diff
@@ -0,0 +1,125 @@
...
+    if (!ts.isPropertyAccessExpression(node.left)) return;
+
+    const leftType = context.checker.getTypeAtLocation(node.left.expression);
+    const decorators = getCustomDecorators(context, leftType);
+    if (decorators.has(DecoratorKind.LuaTable) && node.left.name.text === "length") {
+        throw ForbiddenLuaTableUseException(`A LuaTable object's length cannot be re-assigned.`, node);
+    }
+}
+
+// TODO: Clear if types are reused between compilations
```
Perryvw: Should be in context?

src/transformation/builtins/number.ts
```diff
@@ -0,0 +1,23 @@
+import * as tstl from "../../LuaAST";
+import { TransformationContext } from "../context";
+import { PropertyCallExpression, transformArguments } from "../transformers/call";
+import { UnsupportedProperty } from "../utils/errors";
+import { LuaLibFeature, transformLuaLibFunction } from "../utils/lualib";
+
+// Transpile a Number._ property
+export function transformNumberConstructorCall(
+    context: TransformationContext,
+    expression: PropertyCallExpression
```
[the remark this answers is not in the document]
ark120202: That's possible, but quite pointless. Such a type won't be statically validated anyway, so I don't see any point to make function contract more complex without getting any advantages at usage side.

src/transformation/index.ts
ark120202: Yes. It also can be written as `ts.SyntaxKind[ts.SyntaxKind[syntaxKindKey]]` but that looks even more confusing

src/transformation/transformers/access.ts
```diff
@@ -0,0 +1,89 @@
...
+    const type = context.checker.getTypeAtLocation(expression.expression);
+    const argumentType = context.checker.getTypeAtLocation(expression.argumentExpression);
+    if (isNumberType(context, argumentType) && isArrayType(context, type)) {
+        return createExpressionPlusOne(index);
+    }
+
+    return index;
+}
+
+const transformElementAccessExpression: FunctionVisitor<ts.ElementAccessExpression> = (expression, context) => {
```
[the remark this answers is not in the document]
ark120202: Unlike declarations, function expressions can have a value type annotation.
Using a value level annotation is shorted (declaration would look like `function transformElementAccessExpression(expression: ts.ElementAccessExpression, context: TransformationContext): tstl.Expression`) and reveals possible errors at a declaration time (instead of at usage for declaration).

src/transformation/transformers/binary/destructuring-assignments.ts
ark120202: I've renamed n-ary transformers directories.

(Assuming you meant expressions or statements) I don't think that would make it better. Transformers are grouped by their semantic meaning, so a lot of them handle both expressions and statements (and sometimes declarations or other node kinds).

src/transformation/transformers/class/new.ts
ark120202: `creation` contains the `createClassCreationMethods` function, which creates some common class setup boilerplate. Any suggestions for a better name?

test/unit/loops.spec.ts
```diff
@@ -588,11 +592,11 @@ test.each([
     const luajit = { luaTarget: tstl.LuaTarget.LuaJIT };
 
     expect(() => util.transpileString(loop, lua51)).toThrowExactError(
-        TSTLErrors.UnsupportedForTarget("Continue statement", tstl.LuaTarget.Lua51, ts.createContinue())
+        UnsupportedForTarget("Continue statement", tstl.LuaTarget.Lua51, ts.createContinue())
     );
-    expect(util.transpileString(loop, lua52).indexOf("::__continue1::") !== -1).toBe(true);
-    expect(util.transpileString(loop, lua53).indexOf("::__continue1::") !== -1).toBe(true);
-    expect(util.transpileString(loop, luajit).indexOf("::__continue1::") !== -1).toBe(true);
+    expect(util.transpileString(loop, lua52).indexOf("::__continue2::") !== -1).toBe(true);
```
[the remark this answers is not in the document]
ark120202: It's caused by a small internal change: scope ids now start from 1 instead of 0.

test/unit/typeof.spec.ts
```diff
@@ -37,8 +37,6 @@ test.each([
     { expression: "{}", operator: "!==", compareTo: "object", expectResult: false },
     { expression: "{}", operator: "==", compareTo: "object", expectResult: true },
     { expression: "{}", operator: "!=", compareTo: "object", expectResult: false },
-    { expression: "{}", operator: "<=", compareTo: "object", expectResult: true },
```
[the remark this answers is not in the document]
ark120202: I have actually originally just removed them as unnecessary during refactoring, but actually old behavior was invalid: https://typescripttolua.github.io/play.html#src=console.log(typeof%20'foo'%20%3E%20'object')%3B%0Aconsole.log((typeof%20'foo')%20%3E%20'object')%3B

I probably should add a new test for that.

src/transformation/utils/decorators/collect.ts
```diff
@@ -0,0 +1,121 @@
+import * as ts from "typescript";
+import { TransformationContext } from "../../context";
+
+export enum DecoratorKind {
```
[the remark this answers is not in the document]
ark120202: Looks like TypeScript calls these annotations: https://github.com/microsoft/TypeScript/blob/f41472b427ba99b075123e092548910277d41f55/src/compiler/transformers/declarations.ts#L12-L15, so maybe `custom-annotations`?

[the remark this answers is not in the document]
Perryvw: That sounds good to me.

src/transformation/transformers/todo-move-somewhere.ts
```diff
@@ -0,0 +1,59 @@
...
+    }
+}
+
+// TODO: Consider handling without a visitor since it depends on context too much
+const transformOmittedExpression: FunctionVisitor<ts.OmittedExpression> = node => {
+    const isWithinBindingAssignmentStatement = isWithinLiteralAssignmentStatement(node);
+    return isWithinBindingAssignmentStatement ? lua.createAnonymousIdentifier() : lua.createNilLiteral(node);
+};
+
+export const todoMoveSomewhereVisitors: Visitors = {
```
Perryvw: Maybe split this up in statements and expressions and put them in their top-level transformer files?

src/transformation/utils/annotations/collect.ts
```diff
@@ -0,0 +1,138 @@
...
+    LuaIterator = "LuaIterator",
+    LuaTable = "LuaTable",
+    NoSelf = "NoSelf",
+    NoSelfInFile = "NoSelfInFile",
+    Vararg = "Vararg",
+    ForRange = "ForRange",
+}
+
+function getAnnotationKindByName(name: AnnotationName): AnnotationKind;
+function getAnnotationKindByName(name: string): AnnotationKind | undefined;
```
Perryvw: This line seems a duplicate to the one below, can be removed?

src/transformation/utils/annotations/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from "./collect";
+export * from "./utils";
```
Perryvw: Might as well just put the contents of collect and/or utils in here? This seems a little useless.

src/transformation/utils/lualib.ts
```diff
@@ -0,0 +1,28 @@
...
+    getUsedLuaLibFeatures(context).add(feature);
+}
+
+export function transformLuaLibFunction(
+    context: TransformationContext,
+    feature: LuaLibFeature,
+    tsParent?: ts.Expression,
+    ...params: lua.Expression[]
+): lua.CallExpression {
+    importLuaLibFeature(context, feature);
```
Perryvw: Am I missing something here? Didnt we have a mechanism for lualib dependencies?

src/transformation/transformers/access.ts
```diff
@@ -0,0 +1,103 @@
...
+import * as lua from "../../LuaAST";
+import { transformBuiltinPropertyAccessExpression } from "../builtins";
+import { FunctionVisitor, TransformationContext } from "../context";
+import { AnnotationKind, getTypeAnnotations } from "../utils/annotations";
+import { createExpressionPlusOne } from "../utils/lua-ast";
+import { isArrayType, isNumberType, isStringType } from "../utils/typescript";
+import { tryGetConstEnumValue } from "./enum";
+import { transformLuaTableElementAccessExpression, transformLuaTablePropertyAccessExpression } from "./lua-table";
+
+export function transformElementAccessArgument(
```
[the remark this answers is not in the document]
ark120202: It's used in context-preserving calls and computed assignments

src/transformation/transformers/call.ts
```diff
@@ -0,0 +1,218 @@
...
+import { LuaLibFeature, transformLuaLibFunction } from "../utils/lualib";
+import { isValidLuaIdentifier, luaKeywords } from "../utils/safe-names";
+import { isArrayType, isExpressionWithEvaluationEffect, isInDestructingAssignment } from "../utils/typescript";
+import { transformElementAccessArgument } from "./access";
+import { transformIdentifier } from "./identifier";
+import { transformLuaTableCallExpression } from "./lua-table";
+
+export type PropertyCallExpression = ts.CallExpression & { expression: ts.PropertyAccessExpression };
+
+export function transformArguments(
```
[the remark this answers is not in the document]
ark120202: Yes, we have quite many usages of that

src/transformation/transformers/delete.ts
```diff
@@ -0,0 +1,25 @@
...
+};
+
+export function transformDeleteExpressionStatement(
+    context: TransformationContext,
+    node: ts.ExpressionStatement
+): lua.Statement | undefined {
+    const expression = ts.isExpressionStatement(node) ? node.expression : node;
+    if (ts.isDeleteExpression(expression)) {
+        return lua.createAssignmentStatement(
+            cast(context.transformExpression(expression.expression), lua.isAssignmentLeftHandSideExpression),
```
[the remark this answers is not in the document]
ark120202: For me, `verify` implies that it's returning an error on failure, not throwing.

Maybe something like `assertWith(value, cast)` (which is just `assert(cast(value))`, but with a better failure message)? We'd have to upgrade TS to 3.7 for that though

src/transformation/transformers/enum.ts
```diff
@@ -0,0 +1,94 @@
+import * as ts from "typescript";
+import * as lua from "../../LuaAST";
+import { FunctionVisitor, TransformationContext } from "../context";
+import { AnnotationKind, getTypeAnnotations } from "../utils/annotations";
+import { getSymbolExportScope } from "../utils/export";
+import { createLocalOrExportedOrGlobalDeclaration } from "../utils/lua-ast";
+import { transformIdentifier } from "./identifier";
+import { transformPropertyName } from "./literal";
+
+export function tryGetConstEnumValue(
```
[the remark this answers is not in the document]
ark120202: It wasn't exported when we had multi-pass transformers, but now it's required for access visitors

src/transformation/utils/annotations/collect.ts
ark120202: It's required, last overload is used for implementation, but doesn't contribute to the type

src/transformation/utils/lualib.ts
ark120202: It's done during printing

src/plugins.ts
```diff
@@ -0,0 +1,7 @@
+import { PrinterFactory } from "./LuaPrinter";
+import { Visitors } from "./transformation/context";
+
+export interface Plugin {
```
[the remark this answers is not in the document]
ark120202: It also would include code for loading plugins from options

src/transformation/transformers/variable.ts
```diff
@@ -0,0 +1,226 @@
...
+import { isTupleReturnCall } from "../utils/annotations";
+import { validateAssignment } from "../utils/assignment-validation";
+import { UnsupportedKind } from "../utils/errors";
+import { addExportToIdentifier } from "../utils/export";
+import { createLocalOrExportedOrGlobalDeclaration, createUnpackCall } from "../utils/lua-ast";
+import { LuaLibFeature, transformLuaLibFunction } from "../utils/lualib";
+import { transformIdentifier } from "./identifier";
+import { transformPropertyName } from "./literal";
+
+export function transformArrayBindingElement(
```
Perryvw: maybe variable-declaration would be a better name for this file?

src/transformation/utils/symbols.ts
```diff
@@ -0,0 +1,70 @@
...
+import { TransformationContext } from "../context";
+import { ReferencedBeforeDeclaration } from "./errors";
+import { markSymbolAsReferencedInCurrentScopes } from "./scope";
+import { getFirstDeclarationInFile } from "./typescript";
+
+const symbolIdCounters = new WeakMap<TransformationContext, number>();
+function nextSymbolId(context: TransformationContext): lua.SymbolId {
+    const symbolId = (symbolIdCounters.get(context) || 0) + 1;
+    symbolIdCounters.set(context, symbolId);
+    return symbolId as lua.SymbolId;
```
Perryvw: Does the location of  SymbolId in lua make sense? Should we move it here?

ark120202: In that case we wouldn't be able to store it on `Identifier` node and have to use a WeakMap. That actually might be a good thing, but might require quite a lot of changes, so maybe leave it for a different PR?

src/plugins.ts
```diff
@@ -0,0 +1,27 @@
...
+     * An augmentation to the map of visitors that transform TypeScript AST to Lua AST.
+     *
+     * Key is a `SyntaxKind` of a processed node.
+     */
+    visitors?: Visitors;
+
+    /**
+     * A function that converts Lua AST to a string.
+     *
+     * Only one plugin can provide `printer`.
```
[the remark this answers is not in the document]
ark120202: It's supposed to mean that it can be provided either by 0 or 1 plugins passed to the the compilation. Any wording I can think of sounds a bit confusing, could you make a suggestion?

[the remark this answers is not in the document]
Perryvw: I have some doubts with this entire mechanism. If this can only be supplied once, why give it as an option on all plugins. Would it be an idea to do something like `type Plugin = TransformerPlugin | PrinterPlugin`?

src/transformation/utils/symbols.ts
Perryvw: That's fine with me.

PR #706, opened 2019-08-24
Preserving order in Map and Set using doubly linked lists

Perryvw:
Code complicates a bit with the extra list bookkeeping. This does preserve O(1) insert and deletes. Iteration is still O(n), although it does include some extra table lookups, so probably more like O(2n) compared to the O(n) it was before.

### DIFF
--- a/src/lualib/Set.ts
+++ b/src/lualib/Set.ts
@@ -4,11 +4,10 @@ Set = class Set<T> {
 
     public size = 0;
 
-    // Key-order variables
     private firstKey: T | undefined;
     private lastKey: T | undefined;
-    private nextKey: { [key: string]: T | undefined } = {};
-    private previousKey: { [key: string]: T | undefined } = {};
+    private nextKey: LuaTable<T, T> = new LuaTable();
+    private previousKey: LuaTable<T, T> = new LuaTable();
 
     constructor(values?: Iterable<T> | T[]) {
         if (values === undefined) return;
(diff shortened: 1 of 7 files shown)


src/lualib/Set.ts
```diff
@@ -6,6 +6,12 @@ Set = class Set<T> {
     private items: { [key: string]: boolean } = {};
```
tomblind: Does `Set` even need `items` anymore? `has` could just check `nextKey` or `previousKey`, couldn't it?

tomblind: Wait nevermind... I just realized that won't work for a one-item set.

test/unit/builtins/map.spec.ts
```diff
@@ -135,3 +135,45 @@ test("map size", () => {
     expect(util.transpileAndExecute(`let m = new Map([[1,2],[3,4]]); m.clear(); return m.size;`)).toBe(0);
     expect(util.transpileAndExecute(`let m = new Map([[1,2],[3,4]]); m.delete(3); return m.size;`)).toBe(1);
 });
+
+const testMapConstructionCode = `
```
ark120202: Have you considered using .each?

```ts
test.each(["keys", "values", "entries"])("map.%s() preserves insertion order", (method) => {
    util.testFunction`
        const mymap = new Map();

        mymap.set("x", 1);
        mymap.set("a", 2);
        mymap.set(4, 3);
        mymap.set("b", 6);
        mymap.set(1, 4);
        mymap.set("a", 5);
        
        mymap.delete("b");

        return [...mymap.${method}()];
    `.expectToMatchJsResult();
});
```

Perryvw: That would mean I can no longer also check for the actual expected value. I have not made my mind up on if that should be there or not yet.

src/lualib/Set.ts
Perryvw: You do have a point, I think checking nextKey and lastKey would be enough to get rid of items and would reduce storage space required.

Perryvw: I made this change, seems fine.

PR #711, opened 2019-08-28
Fix broken for-in declaration

Perryvw:


src/LuaTransformer.ts
```diff
@@ -2628,9 +2628,20 @@ export class LuaTransformer {
...
+        // Transform iteration variable
+        let iterationVariable: tstl.Identifier;
+        if (
+            ts.isVariableDeclarationList(statement.initializer) &&
+            !ts.isArrayBindingPattern(statement.initializer.declarations[0].name) &&
+            !ts.isObjectBindingPattern(statement.initializer.declarations[0].name)
+        ) {
+            iterationVariable = this.transformIdentifier(statement.initializer.declarations[0].name);
+        } else if (ts.isIdentifier(statement.initializer)) {
+            iterationVariable = this.transformIdentifier(statement.initializer);
```
ark120202: That way it would create a new variable instead of reassigning existing one. It should do it similarly to for...of:
```lua
for ____key in pairs(expression) do
    initializer = ____key
end
```

Perryvw: I'm not sure what you mean, line 260 will just directly translate the variable name, `for <x> in y` -> `for <x> in pairs(y)`?

src/LuaTransformer.ts
```diff
@@ -2628,9 +2628,20 @@ export class LuaTransformer {
...
 
     public transformForInStatement(statement: ts.ForInStatement): StatementVisitResult {
-        // Get variable identifier
-        const variable = (statement.initializer as ts.VariableDeclarationList).declarations[0];
-        const identifier = variable.name as ts.Identifier;
+        // Transform iteration variable
+        let iterationVariable: tstl.Identifier;
+        if (
+            ts.isVariableDeclarationList(statement.initializer) &&
+            !ts.isArrayBindingPattern(statement.initializer.declarations[0].name) &&
```
ark120202: Wouldn't it be cleaner to include the only one expected node kind instead of filtering out unexpected ones?
```ts
ts.isIdentifier(statement.initializer.declarations[0].name)
````

File: src/LuaTransformer.ts [hunk 3]
```diff
@@ -2643,7 +2639,30 @@ export class LuaTransformer {
...
+            iterationVariable = this.transformIdentifier(statement.initializer.declarations[0].name);
+        } else if (ts.isIdentifier(statement.initializer)) {
+            // Iteration variable becomes ____key
+            iterationVariable = tstl.createIdentifier("____key");
+            // Push variable = ____key to the start of the loop body to match TS scoping
+            const initializer = tstl.createAssignmentStatement(
+                this.transformIdentifier(statement.initializer),
+                iterationVariable
+            );
+            body.statements.splice(0, 0, initializer);
```
```suggestion
            body.statements.unshift(initializer);
```

test/unit/loops.spec.ts
```diff
@@ -630,3 +630,27 @@ test("do...while double-negation", () => {
...
+test("for...in with pre-defined variable", () => {
+    util.testFunction`
+        const obj = { x: "y", foo: "bar" };
+
+        let x = "";
+        let result = [];
+        for (x in obj) {
+            result.push(x);
+        }
+        return result;
```
ark120202: Since objects are compared deeply these 2 tests may be merged to `return { x, result }`

PR #714, opened 2019-09-02
Upgrade TypeScript to v3.6

### LINKED ISSUE #713: TypeScript 3.6 breaks how JSDoc comments work
## source

```typescript
/** @noSelfInFile */

/**
 * This is the require docstring
 */
declare function require(modname: string): any;

require("test")
```

## compiled
```lua
require(nil, "test")
```

## expected
```lua
require("test")
```

ark120202:
Closes #713.

### DIFF
--- a/src/TSHelper.ts
+++ b/src/TSHelper.ts
@@ -404,7 +404,7 @@ export function getCustomDecorators(type: ts.Type, checker: ts.TypeChecker): Map
     return decMap;
 }
 
-export function getCustomNodeDirectives(node: ts.Node): Map<DecoratorKind, Decorator> {
+function getCustomNodeDirectives(node: ts.Node): Map<DecoratorKind, Decorator> {
     const directivesMap = new Map<DecoratorKind, Decorator>();
 
     ts.getJSDocTags(node).forEach(tag => {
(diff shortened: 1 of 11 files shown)


Perryvw:
Do we have a test that fails in TS 3.6 because of this change? If not we shoudl probably add one, a snapshot test with `/** @noSelfInFile */` and some other JSDoc on the first statement would suffice.

src/TSHelper.ts
```diff
@@ -419,10 +419,25 @@ export function getCustomNodeDirectives(node: ts.Node): Map<DecoratorKind, Decor
 }
 
 export function getCustomFileDirectives(file: ts.SourceFile): Map<DecoratorKind, Decorator> {
+    const directivesMap = new Map<DecoratorKind, Decorator>();
+
     if (file.statements.length > 0) {
-        return getCustomNodeDirectives(file.statements[0]);
+        // Manually collect jsDoc because `getJSDocTags` includes tags only from closest comment
+        const jsDoc = file.statements[0].jsDoc;
```
tomblind: Is it possible for a file to have no statements?

ark120202: Yes, for empty file `SourceFile.statements` is an empty array, that's why there is a length check above.

tomblind: Sorry - I completely missed that

PR #718, opened 2019-09-06
No self option

### LINKED ISSUE #716: Add an option to disable implicit `self` argument in project
Implementation in `getDeclarationContextType` should check that `signatureDeclaration.getSourceFile()` is included in `program.getRootFileNames()` and then make `ContextType.Void` default.
It also requires a new `afterDeclarations` transformer, that should add a `@noSelfInFile` synthetic comment to generated declaration files.

hazzard993:
Closes #716

New option: `--noImplicitSelf` (boolean) (`false` by default)

Implicitly enables `@noSelfInFile` for all root files and will add the `@noSelfInFile` annotation to the top of generated declaration files.


src/NoSelfTransformer.ts
```diff
@@ -0,0 +1,12 @@
+import * as ts from "typescript";
+
+export function noSelfTransformerFactory<T extends ts.Node>(_context: ts.TransformationContext): ts.Transformer<T> {
```
ark120202: - Making transformer handle any kind of node defines looser constraint than it could and moves compile-time check to runtime
- You can use function expressions to specify full signature type
- This transformer doesn't handle bundles
- Even though it's a part of a type, IMO `Factory` is unnecessary there, such name better fits a function with `(program: ts.Program) => ts.TransformerFactory` signature (even though it actually is `FactoryFactory`).

```ts
const transformSourceFile: ts.Transformer<ts.SourceFile> = node => {
  const empty = ts.createEmptyStatement();
  ts.addSyntheticLeadingComment(empty, ts.SyntaxKind.MultiLineCommentTrivia, "* @noSelfInFile ", true);
  return ts.updateSourceFileNode(node, [empty, ...node.statements], node.isDeclarationFile);
};

export const noSelfTransformer: ts.TransformerFactory<ts.SourceFile | ts.Bundle> = () => node =>
    ts.isBundle(node) ? ts.updateBundle(node, node.sourceFiles.map(transformSourceFile)) : transformSourceFile(node);
```

For bundles it generates declarations like

```ts
 /** @noSelfInFile */
declare module "foo" { }
/** @noSelfInFile */
declare module "bar" { }
```

Not sure how it works now,

src/Transpile.ts
```diff
@@ -104,7 +104,8 @@ export function transpile({
             const isDeclaration = fileName.endsWith(".d.ts");
             const isDeclarationMap = fileName.endsWith(".d.ts.map");
             if (isDeclaration) {
-                updateTranspiledFile(sourceFile.fileName, { declaration: data });
+                const dataWithoutEmptyStatements = data.replace(/^;/gm, "");
```
ark120202: This looks quite hacky and likely breaks source maps (which is the main reason to use transformer anyway). Have you tried `ts.createNotEmittedStatement`? It requires original node reference, but `sourceFile` or `undefined!` should work. I've also tried to attach it to first first statement, but it doesn't look good when it already has some comments (though probably works)

hazzard993: Brilliant! Didn't know about `NotEmittedStatement`. No semicolon is emitted so this change has been reverted

src/NoSelfTransformer.ts
hazzard993: Thanks a bunch for these notes and code

src/NoSelfTransformer.ts
```diff
@@ -0,0 +1,10 @@
+import * as ts from "typescript";
+
+const transformSourceFile: ts.Transformer<ts.SourceFile> = node => {
+    const empty = ts.createEmptyStatement();
```
ark120202:
```suggestion
    const empty = ts.createNotEmittedStatement(undefined!);
```

(I've left transformer comment before not emitted statement one)

test/unit/functions/noSelfOption.spec.ts
```diff
@@ -0,0 +1,52 @@
+import * as util from "../../util";
+
+test.each([["foo: string, bar: string", `this: void, foo: string, bar: string`]])(
```
ark120202: Do you plan to add more cases? If not, I don't see any reason to use .each

test/unit/functions/noSelfOption.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+    'enables noSelfInFile behaviour for functions ("%s" equivalent to "%s")',
+    (expectedParameters, assignmentParameters) => {
+        util.testFunction`
+            function fooBar(${expectedParameters}) {}
+
+            const test: (${assignmentParameters}) => void = fooBar;
```
ark120202: Since function are assignable to types with fewer parameters you can just use `this: void`. Though there's probably no much reason to add any parameters to these functions anyway, I think it's better to keep test code as simple as possible

test/unit/functions/noSelfOption.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+
+            const test: (${assignmentParameters}) => void = fooBar.fooBar;
+        `
+            .setOptions({ noSelf: true })
+            .expectToHaveNoDiagnostics();
+    }
+);
+
+test("generates declaration files with @noSelfInFile", () => {
+    const result = util.transpileStringsAsProject(
```
ark120202: You can use new builder for that test as well (see next comment)

test/unit/functions/noSelfOption.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+        {
+            noSelf: true,
+            declaration: true,
+        }
+    );
+
+    const declarationFile = result.transpiledFiles.find(transpiledFile => transpiledFile.declaration);
+
+    if (util.expectToBeDefined(declarationFile) && util.expectToBeDefined(declarationFile.declaration)) {
+        expect(declarationFile.declaration).toMatch(/^\/\*\* \@noSelfInFile \*\//);
```
ark120202: Actually, just checking string might be not enough to ensure that it works (for example now it doesn't fail, but we have `;` because of empty statement), it would be nicer to test it like that:
- Build a project with `noSelf: true`
- Take `declarationFile` from it's result
- Build a second project without `noSelf: true` and with a .d.ts file from previous project
- Test assignability like in previous tests

Something like that (haven't tested it though):

```ts
    export function bar() {}
`
    .setOptions({ declaration: true, noSelf: true })
    .expectToHaveNoDiagnostics();

const declarationFile = builder.getLuaResult().transpiledFiles.find(f => f.declaration);
if (!util.expectToBeDefined(declarationFile) || !util.expectToBeDefined(declarationFile.declaration)) return;

util.testModule`
    import { bar } from "./foo";
    const test: (this: void) => void = bar;
`
    .addExtraFile("foo.d.ts", declarationFile.declaration)
    .expectToHaveNoDiagnostics();

```

src/TSHelper.ts
```diff
@@ -712,7 +715,13 @@ export function getDeclarationContextType(
         return ContextType.NonVoid;
     }
 
-    // Walk up to find @noSelf or @noSelfOnFile
+    // If using --noSelf @noSelfInFile should be implicitly enabled
+    const options = program.getCompilerOptions() as CompilerOptions;
+    if (options.noSelf && program.getRootFileNames().includes(signatureDeclaration.getSourceFile().fileName)) {
```
Perryvw: What is the purpose of the second clause here? Maybe add a comment to clarify.

test/unit/functions/noSelfOption.spec.ts
hazzard993: I've been trying to do this but I think I found a bug with with `util.testModule`.

```ts
test("test", () => {
    util.testModule`
        import { bar } from "./foo";
        bar();
    `
        .addExtraFile("foo.d.ts", "export function bar(): void;")
        .expectToHaveNoDiagnostics();
});
```

```
    main.ts:2:29 - error TS2307: Cannot find module './foo'.

    2         import { bar } from "./foo";
```

When renaming `"foo.d.ts"` to `"foo.ts"` and changing the contents of its file to `"export function foo() {}` this works but specifically declarations don't seem to work

hazzard993: Resolved by using `import { bar } from "./foo.d";`

src/TSHelper.ts
```diff
@@ -727,7 +730,13 @@ export function getDeclarationContextType(
         return ContextType.NonVoid;
     }
 
-    // Walk up to find @noSelf or @noSelfOnFile
+    // When using --noSelf and the signature being used is in a file targeted by the program apply the --noSelf rule.
```
ark120202: - Option name in this comment is outdated now
- `being used` -> `is defined`?
- `apply the --noSelf rule` part sounds confusing, did you mean `@noSelf rule`?

src/CommandLineParser.ts
```diff
@@ -38,6 +38,11 @@ const optionDeclarations: CommandLineOption[] = [
         type: "enum",
         choices: Object.values(LuaTarget),
     },
+    {
+        name: "noImplicitSelf",
+        description: "Implicitly enable @noSelfInFile for all target files.",
```
ark120202: I think it would be nice to describe what it actually does and keep `--help` useful without making user look up for annotations documentation

Issue #676, opened 2019-07-20
Migrate to ESLint


ark120202:
I have upgraded few own projects (see [typescript-tstl-plugin](https://github.com/TypeScriptToLua/typescript-tstl-plugin/commit/5a7200b72e18db41ebc74f0ac2fe0d420e082915) as an example) and it it looks quite stable at this point. I have gone through typescript-eslint and few other plugins and made a strict [shareable config](https://github.com/ark120202/eslint-config) which can be used as a base.

ark120202:
Resolved by #839.

PR #724, opened 2019-09-24
Allow throwing arbitrary types and add builtin error classes

TheLartians:
This PR enables throwing custom error objects and adds the builtin error classes as [defined by JavaScript](https://basarat.gitbooks.io/typescript/docs/types/exceptions.html). Also fixes a bug in `LuaTransformer.transformClassExpression()` where local class definitions had global scope.  

### Arbitrary errors

```typescript
try {
  throw {x: "Hello error object!"};
} catch (error) {
  console.log(error.x); // -> Hello error object!
}
```

### Builtin errors

```typescript
try {
  throw RangeError("message"); // or new RangeError("message")
} catch (error) {
  if (error instanceof Error) {
    console.log(error); // -> RangeError: message
    console.log(error.name); // -> RangeError
    console.log(error.message); // -> message
    console.log(error.stack); // prints stack trace from where the error was thrown
  }
}
```

### Inheritance

```typescript
class MyError extends Error{
  name: "MyError"
}

try {
  throw new MyError("message");
} catch (error) {
  // same error properties as builtin errors
} 
```

### DIFF
--- a/src/LuaLib.ts
+++ b/src/LuaLib.ts
@@ -64,7 +64,6 @@ export enum LuaLibFeature {
 const luaLibDependencies: { [lib in LuaLibFeature]?: LuaLibFeature[] } = {
     ArrayFlat: [LuaLibFeature.ArrayConcat],
     ArrayFlatMap: [LuaLibFeature.ArrayConcat],
-    Error: [LuaLibFeature.ArrayMap],
     InstanceOf: [LuaLibFeature.Symbol],
     Iterator: [LuaLibFeature.Symbol],
     ObjectFromEntries: [LuaLibFeature.Iterator, LuaLibFeature.Symbol],
--- a/src/LuaLib.ts
+++ b/src/LuaLib.ts
@@ -64,6 +64,7 @@ export enum LuaLibFeature {
 const luaLibDependencies: { [lib in LuaLibFeature]?: LuaLibFeature[] } = {
     ArrayFlat: [LuaLibFeature.ArrayConcat],
     ArrayFlatMap: [LuaLibFeature.ArrayConcat],
+    Error: [LuaLibFeature.FunctionCall],
     InstanceOf: [LuaLibFeature.Symbol],
     Iterator: [LuaLibFeature.Symbol],
     ObjectFromEntries: [LuaLibFeature.Iterator, LuaLibFeature.Symbol],
--- a/src/LuaLib.ts
+++ b/src/LuaLib.ts
@@ -64,6 +64,7 @@ export enum LuaLibFeature {
 const luaLibDependencies: { [lib in LuaLibFeature]?: LuaLibFeature[] } = {
     ArrayFlat: [LuaLibFeature.ArrayConcat],
     ArrayFlatMap: [LuaLibFeature.ArrayConcat],
+    Error: [LuaLibFeature.ArrayPush, LuaLibFeature.ArrayMap, LuaLibFeature.Index],
     InstanceOf: [LuaLibFeature.Symbol],
     Iterator: [LuaLibFeature.Symbol],
     ObjectFromEntries: [LuaLibFeature.Iterator, LuaLibFeature.Symbol],
--- a/src/LuaLib.ts
+++ b/src/LuaLib.ts
@@ -25,6 +25,7 @@ export enum LuaLibFeature {
     ClassIndex = "ClassIndex",
     ClassNewIndex = "ClassNewIndex",
     Decorate = "Decorate",
+    Error = "Error",
     FunctionApply = "FunctionApply",
     FunctionBind = "FunctionBind",
     FunctionCall = "FunctionCall",
--- a/src/LuaLib.ts
+++ b/src/LuaLib.ts
@@ -64,7 +64,7 @@ export enum LuaLibFeature {
 const luaLibDependencies: { [lib in LuaLibFeature]?: LuaLibFeature[] } = {
     ArrayFlat: [LuaLibFeature.ArrayConcat],
     ArrayFlatMap: [LuaLibFeature.ArrayConcat],
-    Error: [LuaLibFeature.ArrayPush, LuaLibFeature.ArrayMap, LuaLibFeature.Index],
+    Error: [LuaLibFeature.ArrayMap],
     InstanceOf: [LuaLibFeature.Symbol],
     Iterator: [LuaLibFeature.Symbol],
     ObjectFromEntries: [LuaLibFeature.Iterator, LuaLibFeature.Symbol],
--- a/src/LuaTransformer.ts
+++ b/src/LuaTransformer.ts
@@ -2847,15 +2847,14 @@ export class LuaTransformer {
     }
 
     public transformThrowStatement(statement: ts.ThrowStatement): StatementVisitResult {
-        const error = tstl.createIdentifier("error");
-        if (statement.expression === undefined) {
-            return tstl.createExpressionStatement(tstl.createCallExpression(error, []), statement);
-        } else {
-            return tstl.createExpressionStatement(
-                tstl.createCallExpression(error, [this.transformExpression(statement.expression)]),
-                statement
-            );
+        const parameters: tstl.Expression[] = [];
+        if (statement.expression) {
+            parameters.push(this.transformExpression(statement.expression));
         }
+        return tstl.createExpressionStatement(
+            tstl.createCallExpression(tstl.createIdentifier("error"), parameters),
+            statement
+        );
     }
 
     public transformContinueStatement(statement: ts.ContinueStatement): StatementVisitResult {
--- a/src/LuaTransformer.ts
+++ b/src/LuaTransformer.ts
@@ -2848,9 +2848,11 @@ export class LuaTransformer {
 
     public transformThrowStatement(statement: ts.ThrowStatement): StatementVisitResult {
         const parameters: tstl.Expression[] = [];
+
         if (statement.expression) {
             parameters.push(this.transformExpression(statement.expression));
         }
+
         return tstl.createExpressionStatement(
             tstl.createCallExpression(tstl.createIdentifier("error"), parameters),
             statement
--- a/src/LuaTransformer.ts
+++ b/src/LuaTransformer.ts
@@ -2843,19 +2843,14 @@ export class LuaTransformer {
     }
 
     public transformThrowStatement(statement: ts.ThrowStatement): StatementVisitResult {
+        const error = tstl.createIdentifier("error");
         if (statement.expression === undefined) {
-            throw TSTLErrors.InvalidThrowExpression(statement);
-        }
-
-        const type = this.checker.getTypeAtLocation(statement.expression);
-        if (tsHelper.isStringType(type, this.checker, this.program)) {
-            const error = tstl.createIdentifier("error");
+            return tstl.createExpressionStatement(tstl.createCallExpression(error, []), statement);
+        } else {
             return tstl.createExpressionStatement(
                 tstl.createCallExpression(error, [this.transformExpression(statement.expression)]),
                 statement
             );
-        } else {
-            throw TSTLErrors.InvalidThrowExpression(statement.expression);
         }
     }
 
@@ -4208,6 +4203,7 @@ export class LuaTransformer {
 
         const expressionType = this.checker.getTypeAtLocation(expression.expression);
         if (tsHelper.isStandardLibraryType(expressionType, undefined, this.program)) {
+            this.checkForLuaLibType(expressionType);
             const result = this.transformGlobalFunctionCall(expression);
             if (result) {
                 return result;
@@ -5516,6 +5512,42 @@ export class LuaTransformer {
                     return;
                 case "WeakSet":
                     this.importLuaLibFeature(LuaLibFeature.WeakSet);
+                    return;
+                case "Error":
+                    this.importLuaLibFeature(LuaLibFeature.Error);
+                    return;
+                case "ErrorConstructor":
+                    this.importLuaLibFeature(LuaLibFeature.Error);
+                    return;
+                case "RangeError":
+                    this.importLuaLibFeature(LuaLibFeature.Error);
+                    return;
+                case "RangeErrorConstructor":
+                    this.importLuaLibFeature(LuaLibFeature.Error);
+                    return;
+                case "ReferenceError":
+                    this.importLuaLibFeature(LuaLibFeature.Error);
+                    return;
+                case "ReferenceErrorConstructor":
+                    this.importLuaLibFeature(LuaLibFeature.Error);
+                    return;
+                case "SyntaxError":
+                    this.importLuaLibFeature(LuaLibFeature.Error);
+                    return;
+                case "SyntaxErrorConstructor":
+                    this.importLuaLibFeature(LuaLibFeature.Error);
+                    return;
+                case "TypeError":
+                    this.importLuaLibFeature(LuaLibFeature.Error);
+                    return;
+                case "TypeErrorConstructor":
+                    this.importLuaLibFeature(LuaLibFeature.Error);
+                    return;
+                case "URIError":
+                    this.importLuaLibFeature(LuaLibFeature.Error);
+                    return;
+                case "URIErrorConstructor":
+                    this.importLuaLibFeature(LuaLibFeature.Error);
                     return;
             }
         }
--- a/src/LuaTransformer.ts
+++ b/src/LuaTransformer.ts
@@ -5518,6 +5518,7 @@ export class LuaTransformer {
                     this.importLuaLibFeature(LuaLibFeature.WeakSet);
                     return;
             }
+
             if (tsHelper.isBuiltinErrorTypeName(name)) {
                 this.importLuaLibFeature(LuaLibFeature.Error);
             }
--- a/src/LuaTransformer.ts
+++ b/src/LuaTransformer.ts
@@ -2851,6 +2851,7 @@ export class LuaTransformer {
 
         if (statement.expression) {
             parameters.push(this.transformExpression(statement.expression));
+            parameters.push(tstl.createNumericLiteral(0));
         }
 
         return tstl.createExpressionStatement(
--- a/src/LuaTransformer.ts
+++ b/src/LuaTransformer.ts
@@ -3679,7 +3679,10 @@ export class LuaTransformer {
             className = tstl.createAnonymousIdentifier();
         }
 
+        this.pushScope(ScopeType.Function);
         const classDeclaration = this.transformClassDeclaration(expression, className);
+        this.popScope();
+
         return this.createImmediatelyInvokedFunctionExpression(
             this.statementVisitResultToArray(classDeclaration),
             className,
--- a/src/LuaTransformer.ts
+++ b/src/LuaTransformer.ts
@@ -5503,7 +5503,8 @@ export class LuaTransformer {
 
     protected checkForLuaLibType(type: ts.Type): void {
         if (type.symbol) {
-            switch (this.checker.getFullyQualifiedName(type.symbol)) {
+            const name = this.checker.getFullyQualifiedName(type.symbol);
+            switch (name) {
                 case "Map":
                     this.importLuaLibFeature(LuaLibFeature.Map);
                     return;
@@ -5516,42 +5517,9 @@ export class LuaTransformer {
                 case "WeakSet":
                     this.importLuaLibFeature(LuaLibFeature.WeakSet);
                     return;
-                case "Error":
-                    this.importLuaLibFeature(LuaLibFeature.Error);
-                    return;
-                case "ErrorConstructor":
-                    this.importLuaLibFeature(LuaLibFeature.Error);
-                    return;
-                case "RangeError":
-                    this.importLuaLibFeature(LuaLibFeature.Error);
-                    return;
-                case "RangeErrorConstructor":
-                    this.importLuaLibFeature(LuaLibFeature.Error);
-                    return;
-                case "ReferenceError":
-                    this.importLuaLibFeature(LuaLibFeature.Error);
-                    return;
-                case "ReferenceErrorConstructor":
-                    this.importLuaLibFeature(LuaLibFeature.Error);
-                    return;
-                case "SyntaxError":
-                    this.importLuaLibFeature(LuaLibFeature.Error);
-                    return;
-                case "SyntaxErrorConstructor":
-                    this.importLuaLibFeature(LuaLibFeature.Error);
-                    return;
-                case "TypeError":
-                    this.importLuaLibFeature(LuaLibFeature.Error);
-                    return;
-                case "TypeErrorConstructor":
-                    this.importLuaLibFeature(LuaLibFeature.Error);
-                    return;
-                case "URIError":
-                    this.importLuaLibFeature(LuaLibFeature.Error);
-                    return;
-                case "URIErrorConstructor":
-                    this.importLuaLibFeature(LuaLibFeature.Error);
-                    return;
+            }
+            if (tsHelper.isBuiltinErrorTypeName(name)) {
+                this.importLuaLibFeature(LuaLibFeature.Error);
             }
         }
     }
--- a/src/LuaTransformer.ts
+++ b/src/LuaTransformer.ts
@@ -613,6 +613,10 @@ export class LuaTransformer {
         // Get type that is extended
         const extendsType = tsHelper.getExtendedType(statement, this.checker);
 
+        if (extendsType) {
+            this.checkForLuaLibType(extendsType);
+        }
+
         if (!(isExtension || isMetaExtension) && extendsType) {
             // Non-extensions cannot extend extension classes
             const extendsDecorators = tsHelper.getCustomDecorators(extendsType, this.checker);
--- a/src/TSHelper.ts
+++ b/src/TSHelper.ts
@@ -33,6 +33,7 @@ const defaultArrayCallMethodNames = new Set<string>([
     "flatMap",
 ]);
 
+// TODO [2019-09-27/Perry]: Refactor lualib detection to consistent map
 const builtinErrorTypeNames = new Set([
     "Error",
     "ErrorConstructor",
--- a/src/TSHelper.ts
+++ b/src/TSHelper.ts
@@ -33,6 +33,21 @@ const defaultArrayCallMethodNames = new Set<string>([
     "flatMap",
 ]);
 
+const builtinErrorTypeNames = new Set([
+    "Error",
+    "ErrorConstructor",
+    "RangeError",
+    "RangeErrorConstructor",
+    "ReferenceError",
+    "ReferenceErrorConstructor",
+    "SyntaxError",
+    "SyntaxErrorConstructor",
+    "TypeError",
+    "TypeErrorConstructor",
+    "URIError",
+    "URIErrorConstructor",
+]);
+
 export function getExtendedTypeNode(
     node: ts.ClassLikeDeclarationBase,
     checker: ts.TypeChecker
@@ -1040,4 +1055,8 @@ export function formatPathToLuaPath(filePath: string): string {
         filePath = filePath.replace(/\.\\/g, "").replace(/\\/g, ".");
     }
     return filePath.replace(/\.\//g, "").replace(/\//g, ".");
+}
+
+export function isBuiltinErrorTypeName(name: string): boolean {
+    return builtinErrorTypeNames.has(name);
 }
--- a/src/lualib/Error.ts
+++ b/src/lualib/Error.ts
@@ -1,6 +1,6 @@
-interface ErrorType<T> extends Function {
+interface ErrorType {
     name: string;
-    new (...args: any[]): T;
+    new (...args: any[]): Error;
 }
 
 function __TS__GetErrorStack(constructor: Function): string {
@@ -32,7 +32,7 @@ function __TS__WrapErrorToString<T extends Error>(getDescription: (this: T) => s
     };
 }
 
-function __TS__InitErrorClass<T>(Type: ErrorType<T>, name: string): any {
+function __TS__InitErrorClass(Type: ErrorType, name: string): any {
     Type.name = name;
     return setmetatable(Type, {
         __call: (_self: any, message: string) => new Type(message),
@@ -40,7 +40,7 @@ function __TS__InitErrorClass<T>(Type: ErrorType<T>, name: string): any {
 }
 
 Error = __TS__InitErrorClass(
-    class {
+    class implements Error {
         public name = "Error";
         public stack: string;
 
--- a/src/lualib/Error.ts
+++ b/src/lualib/Error.ts
@@ -1,93 +1,76 @@
-Error = class {
-    public static captureStackTrace(targetObject: {}, depth: number): void {
-        targetObject["stack"] = debug.traceback(undefined, depth);
-    }
+type TSTLCapturedErrorStack = Array<{
+    namewhat: string;
+    name?: string;
+    source: string;
+    short_src: string;
+    currentline: number;
+    func: Function;
+}>;
 
-    public message: string;
-    public name: string;
-    public stack?: string;
+function __TS__GetErrorStack(constructor: Function): TSTLCapturedErrorStack {
+    const functionFrames = [];
+    let level = 1;
+    while (true) {
+        const info = debug.getinfo(level, "f");
+        level += 1;
+        if (!info || info.func === constructor) {
+            break;
+        }
+    }
+    while (true) {
+        const info = debug.getinfo(level, "Snl");
+        if (!info) break;
+        if (info.currentline !== -1) {
+            functionFrames.push(info);
+        }
+        level += 1;
+    }
+    return functionFrames;
+}
 
-    constructor(message?: string, name?: string) {
-        this.message = message || "";
-        this.name = name || "Error";
-        Error["captureStackTrace"](this, name ? 5 : 4);
-    }
+function __TS__ConvertErrorStack(stack: TSTLCapturedErrorStack): string {
+    const info = stack
+        .map(v => {
+            if (v.namewhat === "") {
+                return `${v.short_src}:${v.currentline}`;
+            } else {
+                return `${v.short_src}:${v.currentline} in ${v.namewhat} ${v.name}`;
+            }
+        })
+        .join("\n");
+    const transform = (globalThis as any).__TS__SourceMapTransform;
+    return transform ? transform(info) : info;
+}
 
-    public __tostring(): string {
-        return `${this.name}: ${this.message}`;
-    }
-} as any;
+function __TS__GetErrorString(this: void, error: Error): string {
+    return error.message !== "" ? `${error.name}: ${error.message}` : error.name;
+}
 
-setmetatable(Error, { __call: (_self: any, message: string) => new Error(message) });
+function __TS__InitErrorClass(Type: any): any {
+    Type.prototype.__tostring = __TS__GetErrorString;
+    return setmetatable(Type, {
+        __index: getmetatable(Type),
+        __call: (_self: any, message: string) => new Type(message),
+    });
+}
 
-/** Standard error types */
+Error = __TS__InitErrorClass(
+    class {
+        public message: string;
+        public name = "Error";
+        public stack: string;
 
-RangeError = class extends Error {
-    constructor(message?: string) {
-        // @ts-ignore
-        super(message, "RangeError");
+        constructor(message = "") {
+            this.message = message;
+            this.stack = __TS__ConvertErrorStack(__TS__GetErrorStack((this.constructor as any).new));
+        }
     }
-    public __tostring(): string {
-        return `${this.name}: ${this.message}`;
-    }
-} as any;
-setmetatable(RangeError, {
-    __call: (_self: any, message: string) => new RangeError(message),
-    __index: getmetatable(RangeError),
-});
+);
 
-ReferenceError = class extends Error {
-    constructor(message?: string) {
-        // @ts-ignore
-        super(message, "ReferenceError");
-    }
-    public __tostring(): string {
-        return `${this.name}: ${this.message}`;
-    }
-} as any;
-setmetatable(ReferenceError, {
-    __call: (_self: any, message: string) => new ReferenceError(message),
-    __index: getmetatable(ReferenceError),
-});
-
-SyntaxError = class extends Error {
-    constructor(message?: string) {
-        // @ts-ignore
-        super(message, "SyntaxError");
-    }
-    public __tostring(): string {
-        return `${this.name}: ${this.message}`;
-    }
-} as any;
-setmetatable(SyntaxError, {
-    __call: (_self: any, message: string) => new SyntaxError(message),
-    __index: getmetatable(SyntaxError),
-});
-
-TypeError = class extends Error {
-    constructor(message?: string) {
-        // @ts-ignore
-        super(message, "TypeError");
-    }
-    public __tostring(): string {
-        return `${this.name}: ${this.message}`;
-    }
-} as any;
-setmetatable(TypeError, {
-    __call: (_self: any, message: string) => new TypeError(message),
-    __index: getmetatable(TypeError),
-});
-
-URIError = class extends Error {
-    constructor(message?: string) {
-        // @ts-ignore
-        super(message, "URIError");
-    }
-    public __tostring(): string {
-        return `${this.name}: ${this.message}`;
-    }
-} as any;
-setmetatable(URIError, {
-    __call: (_self: any, message: string) => new URIError(message),
-    __index: getmetatable(URIError),
-});
+for (const errorName of ["RangeError", "ReferenceError", "SyntaxError", "TypeError", "URIError"]) {
+    globalThis[errorName] = __TS__InitErrorClass(
+        class extends Error {
+            public name = errorName;
+        }
+    );
+}
--- a/src/lualib/Error.ts
+++ b/src/lualib/Error.ts
@@ -6,6 +6,11 @@ type TSTLCapturedErrorStack = Array<{
     currentline: number;
 }>;
 
+interface ErrorType<T> extends Function {
+    name: string;
+    new (...args: any[]): T;
+}
+
 function __TS__GetErrorStack(constructor: Function): TSTLCapturedErrorStack {
     const functionFrames = [];
     let level = 1;
@@ -49,7 +54,7 @@ function __TS__GetErrorString(this: void, error: Error): string {
     return error.message !== "" ? `${error.name}: ${error.message}` : error.name;
 }
 
-function __TS__InitErrorClass(Type: any, name?: string): any {
+function __TS__InitErrorClass<T>(Type: ErrorType<T>, name?: string): any {
     if (name) {
         Type.name = name;
     }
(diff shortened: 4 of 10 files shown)


src/lualib/Error.ts
```diff
@@ -0,0 +1,93 @@
+Error = class {
+    public static captureStackTrace(targetObject: {}, depth: number): void {
+        targetObject["stack"] = debug.traceback(undefined, depth);
+    }
+
+    public message: string;
+    public name: string;
+    public stack?: string;
+
+    constructor(message?: string, name?: string) {
```
ark120202: Why `name` is a second argument there? In JS you'd usually just set `name` field in a subclass.

src/lualib/Error.ts
```diff
@@ -0,0 +1,93 @@
...
+        this.message = message || "";
```
ark120202: We can use a constructor field there

```ts
constructor(public message = "")
```

src/lualib/Error.ts
```diff
@@ -0,0 +1,93 @@
...
+        this.name = name || "Error";
+        Error["captureStackTrace"](this, name ? 5 : 4);
+    }
+
+    public __tostring(): string {
+        return `${this.name}: ${this.message}`;
```
ark120202:
```suggestion
        return _VERSION === "Lua 5.1" ? `${this.name}: ${this.message}` : this.stack;
```

- Lua 5.1 doesn't use tostring on errors
- LuaJIT uses tostring and appends stack to it
- Lua 5.2 and 5.3 uses tostring and uses it's raw result

Because of that we have to show a short form for JIT and add stack for > 5.2. I'm not aware of any good way to differentiate between JIT and 5.1, but since 5.1 would require some top-level handler anyway we can just use `_VERSION === "Lua 5.1"`.

src/lualib/Error.ts
```diff
@@ -0,0 +1,93 @@
+Error = class {
+    public static captureStackTrace(targetObject: {}, depth: number): void {
+        targetObject["stack"] = debug.traceback(undefined, depth);
```
ark120202: If we want to keep non-standard extension exposed let's keep the same function signature as in V8. I think it should be possible by walking through getinfo.

Also, `stack` should contain type and message as well.

test/unit/error.spec.ts
```diff
@@ -292,3 +285,76 @@ test("return from nested finally", () => {
     `;
     expect(util.transpileAndExecute(code)).toBe("finally AB");
 });
+
+test("throw and catch custom error object", () => {
```
ark120202: You can use builder API for these tests, search for `util.testFunction` usage.

test/unit/error.spec.ts
```diff
@@ -292,3 +285,76 @@ test("return from nested finally", () => {
...
+    const code = `
+      try {
+          throw {x: "Hello error object!"};
```
ark120202:
```suggestion
          throw { x: "Hello error object!" };
```

src/lualib/Error.ts
```diff
@@ -0,0 +1,93 @@
...
+setmetatable(Error, { __call: (_self: any, message: string) => new Error(message) });
+
+/** Standard error types */
+
+RangeError = class extends Error {
+    constructor(message?: string) {
+        // @ts-ignore
+        super(message, "RangeError");
+    }
+    public __tostring(): string {
```
ark120202: Would custom subclasses also have to redefine it?

src/lualib/Error.ts
TheLartians: Do we really want `tostring(error)` to emit the full stacktrace? AFAICT in JavaScript printing the error usually gives `${name}: ${message}`.

src/lualib/Error.ts
TheLartians: Good point, that was actually my original intention but I wasn't aware that we can actually get the calling function through `debug.getinfo`. So it may make sense to implement the intended behaviour. `debug.getinfo` has the advantage of respecting source maps though.

src/lualib/Error.ts
TheLartians: Unfortunately yes, as Lua basically uses `rawget` to find metamethods.

src/lualib/Error.ts
TheLartians: As the implementation gets hidden away behind the default `Error` type, my idea was to use the name parameter to avoid redundant work and to choose and appropriate stack trace here. But if we do use `captureStackTrace` as intended it would make sense to stick with the standard definition here too.

ark120202: Actually I guess it's not that simple. In JS extended error classes don't have to use `captureStackTrace` (it has to be used only as a hack for es5 downleveling), so that's something that would be nice to avoid.

Looks like .stack in JS is a lazy getter:

```js
const error = new Error();
error.stack // Error...
error.name = 'MyError';
error.stack // Error...
```

```js
const error = new Error();
error.name = 'MyError';
error.stack // MyError...
```

src/lualib/Error.ts
ark120202: `__tostring` there is generally added not for `tostring(error)`, but for standard top-level lua error handler, which usually attaches stack trace to thrown strings, but doesn't do that for objects in some versions. Indeed, it would be nice to make `tostring(error)` consistent as well, maybe it would be possible with some call stack inspection?

TBH at this point all of that sounds pretty complex, maybe we should consider always using `${name}: ${message}` for `__tostring` and provide some custom top-level handler to attach stack.

src/lualib/Error.ts
TheLartians: Lazy getter would be nice, however unavailable here due to #725 as it would break subclassing `Error`.

src/lualib/Error.ts
TheLartians: For now I've renamed the function to `__TS__GetErrorStack(constructor: Function)` that actually behaves similarly as `captureStackTrace` should, but simply returns the stack trace up to the `constructor` provided.

TheLartians: > Also, stack should contain type and message as well

According to my tests on [TypeScript Playground](https://www.typescriptlang.org/play/#code/FAMwrgdgxgLglgewgAgOYAoCUBvYz-ICGANgKYBOM6AoueQuegEQAWpxxCyF95TmAOgDOMQlADWmYAF9goSLEQoQWXATRYA3DLkrMQA) the stack does not contain type and message.

src/lualib/Error.ts
TheLartians: I would also rather have such functionality in the top-level handler rather than the error class itself.

src/lualib/Error.ts
TheLartians: I've been testing on Safari. In Chrome/V8 the stack does include the error name and message before the traceback. I prefer `Error.stack` to only contain the traceback, but we might just as well accept the V8 as ground truth here.

TheLartians: However, as the stack property is set in the constructor and not lazily evaluated due to #725, we cannot append the error name and message as they are still unknown here. So for now let's simply go with Safari's interpretation.

src/LuaTransformer.ts
```diff
@@ -2843,19 +2847,14 @@ export class LuaTransformer {
     }
 
     public transformThrowStatement(statement: ts.ThrowStatement): StatementVisitResult {
```
```ts
public transformThrowStatement(statement: ts.ThrowStatement): StatementVisitResult {
    const parameters: tstl.Expression[] = [];
    if (statement.expression) {
        parameters.push(this.transformExpression(statement.expression));
    }
    
    return tstl.createExpressionStatement(
        tstl.createCallExpression(tstl.createIdentifier("error"), parameters),
        statement
    );
}
```

src/lualib/Error.ts
```diff
@@ -0,0 +1,78 @@
...
+        .join("\n");
+    const transform = globalThis.__TS__SourceMapTransform;
+    return transform ? transform(info) : info;
+}
+
+function __TS__GetErrorString(this: void, error: Error): string {
+    return error.message !== "" ? `${error.name}: ${error.message}` : error.name;
+}
+
+function __TS__InitErrorClass(Type: any): any {
```
ark120202: We also need `Type.name = name` to add names to all these classes

src/lualib/Error.ts
```diff
@@ -0,0 +1,78 @@
+type TSTLCapturedErrorStack = Array<{
```
ark120202: It's not really tstl-specific, it's just a result of debug.getinfo, so it should be added to debug declarations (see https://github.com/ark120202/lua-types/blob/master/core/debug.d.ts#L39).

src/lualib/Error.ts
ark120202: Hmm, maybe `getmetatable(this).__tostring = ...` in Error constructor?

ark120202:
> Lazy getter would be nice, however unavailable here due to #725 as it would break subclassing Error.

> I've been testing on Safari. In Chrome/V8 the stack does include the error name and message before the traceback. I prefer Error.stack to only contain the traceback, but we might just as well accept the V8 as ground truth here.

Actually, looks like `Error#stack` isn't part of a standard at all, so we aren't ought to follow any of existing implementations. But doing what Safari does there isn't good, since `.stack` is usually used to print the full error text in the same way you'd see it after throwing.

There is a [stage 1 proposal](https://github.com/tc39/proposal-error-stacks) ([rendered](https://tc39.es/proposal-error-stacks)) which I think we should follow. It defines `stack` as a getter with that logic: https://tc39.es/proposal-error-stacks/#sec-getstackstring.

src/lualib/Error.ts
TheLartians: Hm that would work for now.

src/LuaLib.ts
```diff
@@ -63,6 +64,7 @@ export enum LuaLibFeature {
 const luaLibDependencies: { [lib in LuaLibFeature]?: LuaLibFeature[] } = {
     ArrayFlat: [LuaLibFeature.ArrayConcat],
     ArrayFlatMap: [LuaLibFeature.ArrayConcat],
+    Error: [LuaLibFeature.ArrayPush, LuaLibFeature.ArrayMap, LuaLibFeature.Index],
```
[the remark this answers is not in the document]
TheLartians: I've reduced the dependencies to `[ArrayMap]`. That one adds about as much code as implementing a custom solution, so I think it makes sense to keep it.

src/lualib/Error.ts
TheLartians: This case it is specific though as it represents the exact parameters requested from `debug.getinfo(level, "Snl");`.

src/lualib/Error.ts
```diff
@@ -0,0 +1,78 @@
...
+        if (!info) break;
+        if (info.currentline !== -1) {
+            functionFrames.push(info);
+        }
+        level += 1;
+    }
+    return functionFrames;
+}
+
+function __TS__ConvertErrorStack(stack: TSTLCapturedErrorStack): string {
```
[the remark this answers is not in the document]
TheLartians: Imo the error stack should match `debug.getinfo`. As far as I can tell there is also no standard way for users to retrieve the sourcemaps, so it makes sense to directly convert here as well.

TheLartians:
> There is a stage 1 proposal (rendered) which I think we should follow.

I agree, but for that to work we would need to use accessors, which don't work with subclassing at the moment. We can update the functionality as soon as #725 is fixed.

src/lualib/Error.ts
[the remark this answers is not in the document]
TheLartians: Also in the future, the call to `__TS__ConvertErrorStack` can be moved to the getter so it won't be invoked during the snapshot.

src/LuaTransformer.ts
```diff
@@ -5514,6 +5516,42 @@ export class LuaTransformer {
                 case "WeakSet":
                     this.importLuaLibFeature(LuaLibFeature.WeakSet);
                     return;
+                case "Error":
```
Perryvw: I don't really like this, could we move all error types to a const set of strings and just check `.has()`?

src/lualib/Error.ts
```diff
@@ -0,0 +1,82 @@
...
+    name?: string;
+    source: string;
+    short_src: string;
+    currentline: number;
+}>;
+
+function __TS__GetErrorStack(constructor: Function): TSTLCapturedErrorStack {
+    const functionFrames = [];
+    let level = 1;
+    while (true) {
```
Perryvw: This looks dodgy to me, maybe we should set a maximum level limit for this while loop to prevent infinite loops?

src/lualib/Error.ts
```diff
@@ -0,0 +1,82 @@
...
+    const info = stack
+        .map(v => {
+            if (v.namewhat === "") {
+                return `${v.short_src}:${v.currentline}`;
+            } else {
+                return `${v.short_src}:${v.currentline} in ${v.namewhat} ${v.name}`;
+            }
+        })
+        .join("\n");
+    const transform = globalThis.__TS__SourceMapTransform;
```
Perryvw: Would this not require a sourcemap traceback dependency?

src/lualib/Error.ts
```diff
@@ -0,0 +1,82 @@
...
+    return transform ? transform(info) : info;
+}
+
+function __TS__GetErrorString(this: void, error: Error): string {
+    return error.message !== "" ? `${error.name}: ${error.message}` : error.name;
+}
+
+function __TS__InitErrorClass(Type: any, name?: string): any {
```
Perryvw: I don't like this `any`, there must be a way to improve this type.

test/unit/error.spec.ts
```diff
@@ -292,3 +285,101 @@ test("return from nested finally", () => {
...
+          return error.x;
+      }
+    `.expectToMatchJsResult();
+});
+
+test.each(["Error", "RangeError", "ReferenceError", "SyntaxError", "TypeError", "URIError", "MyError"])(
+    "throw builtin Errors as classes",
+    errorType => {
+        util.testFunction`
+            class MyError extends Error {
```
Perryvw: This looks strange, split this into two tests -> throw builtin errors as classes & throw custom Error as class (does this already exist as test 'subclass error'?)

src/lualib/Error.ts
TheLartians: The loop is already limited by the maximum Lua stack size as `debug.getinfo` [will eventually](https://www.lua.org/pil/23.1.html) return `nil`. We could set an additional limit on the returned stack size, but I'd prefer keeping the full stack.

src/lualib/Error.ts
TheLartians: Not really. `globalThis.__TS__SourceMapTransform` is defined by default if and only if source maps are enabled. If source maps are enabled the function exists and we transform the traceback. Otherwise we keep the original stack unmodified.

test/unit/error.spec.ts
TheLartians: You're right, I was misusing the test to check if the string conversion works for subclasses. I've extended the `subclass error` test.

src/LuaTransformer.ts
```diff
@@ -5515,6 +5518,9 @@ export class LuaTransformer {
                     this.importLuaLibFeature(LuaLibFeature.WeakSet);
                     return;
             }
+            if (tsHelper.isBuiltinErrorTypeName(name)) {
```
Perryvw: Whiteline above this please :D

test/unit/error.spec.ts
```diff
@@ -292,3 +285,97 @@ test("return from nested finally", () => {
...
+        class MyError extends Error {
+            name: "MyError"
+        }
+        
+        try {
+            throw new MyError();
+        } catch (error) {
+            if (error instanceof Error) {
+                return \'\$error\';
+            } else {
```
Perryvw: Only one branch of this if/else seems to do anything in this test?

src/lualib/Error.ts
Perryvw: Not true, you defined __TS__SourceMapTransform in the SourceMapTraceBack lualib entry, which is only defined if sourceMapTraceBack is enabled in options (regardless of source maps). That said, I think the way you are handling this is correct since you only want to apply it at the moment sourceMapTraceBack is defined. That said, I would like to see a test for this with both sourceMapTraceBack enabled and disabled.

TheLartians: Ah, so _sourceMapTraceBack_ is what I meant to say then :D

test/unit/error.spec.ts
TheLartians: The else branch is implicitly stopping the test from giving false positives as I'm using `.expectNoExecutionError().expectToMatchJsResult()` to validate the result.

Perryvw: Would be nice if you added a comment with that explanation in the code here, maybe in combination with a better error, something like `Error("This error serves to prevent false positives from .expectNoExecutionError()")`

test/unit/error.spec.ts
```diff
@@ -292,3 +286,101 @@ test("return from nested finally", () => {
...
+            outerFunctionThatThrows();
+        } catch (error) {
+            if (error instanceof Error) {
+                return error.stack;
+            } else {
+                throw TypeError();
+            }
+        }
+    `
+            .setOptions({ sourceMapTraceback, luaLibImport: tstl.LuaLibImportKind.Inline })
```
Perryvw: Could we somehow assert the stack to see the traceback is correctly applied?

TheLartians: I'm not sure, AFAICT we don't have any tests that check the sourcemap transformation. Do they even work with non-file input?

src/lualib/Error.ts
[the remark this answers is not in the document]
ark120202: In any case, builtin source map handling is a kind of temporary thing:

> // TODO: In the future, change this to __TS__RegisterFileInfo and provide tstl interface to
> // get some metadata about transpilation.

In the future traceback patch would be packed in a separate module, which would have better module and bundling support and would allow to use sourcemaps for environments that don't call custom `debug.traceback`.

So what is needed there is some extension point that would allow to tap into stack formatting and allow to change it in environment-specific way. But actually, if we want stack to have the same format as `debug.traceback` has can't we just use `debug.traceback`?

test/unit/error.spec.ts
ark120202: If it wouldn't throw an error it should fail anyway since result wouldn't be the same as in JS

test/unit/error.spec.ts
ark120202: https://github.com/TypeScriptToLua/TypeScriptToLua/blob/master/test/unit/printer/sourcemaps.spec.ts

test/unit/error.spec.ts
```diff
@@ -292,3 +286,100 @@ test("return from nested finally", () => {
...
+        .expectNoExecutionError()
+        .expectToMatchJsResult();
+});
+
+test.each([["Error", false], ["Error", true], ["RangeError", false], ["MyError", false]] as Array<[string, boolean]>)(
+    "get stack from error",
+    (errorType, sourceMapTraceback) => {
+        const stack = util.testFunction`
+        class MyError extends Error {
+            name: "MyError"
```
ark120202:
```suggestion
            public name = "MyError";
```

test/unit/error.spec.ts
```diff
@@ -292,3 +286,100 @@ test("return from nested finally", () => {
...
+    `
+        .expectNoExecutionError()
+        .expectToMatchJsResult();
+});
+
+test.each([["Error", false], ["Error", true], ["RangeError", false], ["MyError", false]] as Array<[string, boolean]>)(
+    "get stack from error",
+    (errorType, sourceMapTraceback) => {
+        const stack = util.testFunction`
+        class MyError extends Error {
```
ark120202: It's not indented

test/unit/error.spec.ts
```diff
@@ -292,3 +286,100 @@ test("return from nested finally", () => {
...
+            } else {
+                throw Error("This error serves to prevent false positives from .expectNoExecutionError()");
+            }
+        }
+    `
+        .expectNoExecutionError()
+        .expectToMatchJsResult();
+});
+
+test.each([["Error", false], ["Error", true], ["RangeError", false], ["MyError", false]] as Array<[string, boolean]>)(
```
ark120202: Inline casts aren't safe, it wouldn't fail even when some element doesn't have a correct type. A better solution would be `as const`, but current jest types don't allow readonly tuples since it supports lower TypeScript versions.

Anyway, I think sourcemap tests should go to `sourcemaps.spec.ts`.

src/LuaTransformer.ts
ark120202: We probably should enforce it after #676 with https://eslint.org/docs/rules/padding-line-between-statements

src/lualib/declarations/debug.d.ts
```diff
@@ -2,4 +2,5 @@
 
 declare namespace debug {
     function traceback(...args: any[]): string;
+    function getinfo(i: number, what?: string): any;
```
Perryvw: This return value can be typed properly from https://www.lua.org/pil/23.1.html

src/lualib/declarations/global.d.ts
```diff
@@ -10,6 +10,7 @@ declare function type(
     value: any
 ): "nil" | "number" | "string" | "boolean" | "table" | "function" | "thread" | "userdata";
 declare function setmetatable<T extends object>(table: T, metatable: any): T;
+declare function getmetatable<T extends object>(table: T): any;
```
Perryvw: I'm considering adding a second type argument, ie ` TInput` and `TMetaTable` and having this return typed as ` `TMetaTable`, thoughts?

ark120202: What do you think about using lua-types in lualib? I have originally wanted to avoid it as it may slow down some PRs and we haven't used standard library too much, but now I think it may give more advantages making lualib more sound

test/unit/error.spec.ts
TheLartians: Assuming the test is properly written and works as expected. At this point It’s probably better to also check the actual result.

ark120202: We have discussed that with @Perryvw some time ago in Discord. The thing is that we don't really care about actual result of executed code, but only that it matches what JS would give. Ensuring that test code is correct was one of my worries about that, but adding expected result just gives one more point to make a mistake. Also, to check test code during development there is a `.debug` utility.

test/unit/error.spec.ts
TheLartians: Yeah I noticed that, but it’s not obvious to me if the output of debug.traceback is actually checked anywhere.

src/lualib/declarations/global.d.ts
TheLartians: About to suggest the same thing. I think that makes much more sense to have common, well-maintained definitions than to partially reimplement them for Lualib.

src/lualib/declarations/debug.d.ts
TheLartians: True, though in this case it would only partially help because the exact type depends on the input argument, so we would still rely on external assumptions.

src/lualib/Error.ts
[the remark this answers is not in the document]
TheLartians: Yeah more or less how it was in the beginning. I think now it makes sense to actually rollback to `debug.traceback` for printing the stack.

test/unit/error.spec.ts
TheLartians: Having an additional sanity check inside the test code is probably won't hurt then.

src/LuaTransformer.ts
```diff
@@ -2843,20 +2847,14 @@ export class LuaTransformer {
...
-        if (tsHelper.isStringType(type, this.checker, this.program)) {
-            const error = tstl.createIdentifier("error");
-            return tstl.createExpressionStatement(
-                tstl.createCallExpression(error, [this.transformExpression(statement.expression)]),
-                statement
-            );
-        } else {
-            throw TSTLErrors.InvalidThrowExpression(statement.expression);
+        const parameters: tstl.Expression[] = [];
+        if (statement.expression) {
```
Perryvw: Please add some whitelines before and after this if block

src/TSHelper.ts
```diff
@@ -33,6 +33,21 @@ const defaultArrayCallMethodNames = new Set<string>([
     "flatMap",
 ]);
 
+const builtinErrorTypeNames = new Set([
```
Perryvw: Could you mark this with a `// TODO [2019-09-27/Perry]: Refactor lualib detection to consistent map` :)

src/lualib/declarations/debug.d.ts
Perryvw: Yes, but if you type them all as optional properties it will already be a lot more readable/maintainable for other developers that might not know the details of this function.

src/lualib/declarations/global.d.ts
Perryvw: I think that's probably out of scope for this PR, but we should discuss it. I'm currently leaning to it's alright, but maybe we should consider moving lua-types to the TypeScriptToLua org in that case.

test/unit/error.spec.ts
Perryvw: It's probably nicer if these were just objects `{error: "RangeError", sourceMapTraceback: false}` anyway. Why not test all combinations by the way?

test/unit/error.spec.ts
Perryvw: Seems like this is a gap in our tests actually, I'm fine with it if you add a TODO here to add that to this test once we figure out how to do it. I made an issue to track this: #726

TheLartians:
For simplicity I've decided to rollback to `debug.traceback` for obtaining the error stack, so any source map transformations aren't anymore in the scope of this PR and are handled by the default or overwritten traceback function.

src/lualib/Error.ts
```diff
@@ -0,0 +1,56 @@
...
+        level += 1;
+        if (!info) {
+            // constructor is not in call stack
+            level = 1;
+            break;
+        } else if (info.func === constructor) {
+            break;
+        }
+    }
+    return debug.traceback(undefined, level);
```
ark120202:
```suggestion

    return debug.traceback(undefined, level);
```

src/lualib/Error.ts
```diff
@@ -0,0 +1,56 @@
...
+    });
+}
+
+Error = __TS__InitErrorClass(
+    class Error {
+        public name = "Error";
+        public message: string;
+        public stack: string;
+
+        constructor(message = "") {
```
ark120202:
```suggestion
        constructor(public message = "") {
```

src/lualib/Error.ts
```diff
@@ -0,0 +1,56 @@
...
+        }
+    }
+    return debug.traceback(undefined, level);
+}
+
+function __TS__GetErrorString(this: void, error: Error): string {
+    return error.message !== "" ? `${error.name}: ${error.message}` : error.name;
+}
+
+function __TS__InitErrorClass<T>(Type: ErrorType<T>, name?: string): any {
```
ark120202: Type of `Type` there doesn't make any code more sound since generic isn't used in return type, so I don't see any advantages plain `any` there. Actually it could be typed correctly:

```ts
type AddCallSignature<T> = T extends new (...args: infer TArgs) => infer TReturn
    ? T & ((...args: TArgs) => TReturn)
    : never;

function __TS__InitErrorClass<T>(Type: T, name?: string): AddCallSignature<T> {
    if (name) {
        Type.name = name;
    }

    return setmetatable(Type, {
        __call: (_self: any, message: string) => new Type(message),
    });
}

__TS__InitErrorClass(class Error {}); // => typeof Error & (() => Error)
```

But I'm not really sure if adding a quite non trivial type there worth it.

src/lualib/Error.ts
```diff
@@ -0,0 +1,56 @@
...
+Error = __TS__InitErrorClass(
+    class Error {
+        public name = "Error";
+        public message: string;
+        public stack: string;
+
+        constructor(message = "") {
+            this.message = message;
+            this.stack = __TS__GetErrorStack((this.constructor as any).new);
+            getmetatable(this).__tostring = __TS__GetErrorString;
```
ark120202: I wanted to suggest `metatable.__tostring = metatable.__tostring || __TS__GetErrorString`, but actually that would break inheritance:

```ts
class CustomError { toString() { return '...' } }
class ExtendedCustomError extends CustomError {}
```

Not sure how that would affect performance, but to support this case we'd have to walk through prototype chain or reconsider doing it during transformation.

src/lualib/Error.ts
```diff
@@ -0,0 +1,56 @@
...
+    if (name) {
+        Type.name = name;
+    }
+    return setmetatable(Type, {
+        __call: (_self: any, message: string) => new Type(message),
+    });
+}
+
+Error = __TS__InitErrorClass(
+    class Error {
```
ark120202: Might be a bit cleaner to make `name` in `__TS__InitErrorClass` non-optional and provide `"Error"` there.

test/unit/error.spec.ts
```diff
@@ -292,3 +285,88 @@ test("return from nested finally", () => {
     `;
     expect(util.transpileAndExecute(code)).toBe("finally AB");
 });
+
+test("throw and catch custom error object", () => {
+    util.testFunction`
+        try {
+            throw { x: "Hello error object!" };
```
ark120202: Would it make sense to add tests for other thrown types (`boolean`, `number` and `undefined`)?

test/unit/error.spec.ts
```diff
@@ -292,3 +285,88 @@ test("return from nested finally", () => {
...
+        } catch (error) {
+            return error.x;
```
ark120202:
```suggestion
            return error;
```

It would be easier to test other types with `.each` and also would check that we don't change thrown value in any way

test/unit/error.spec.ts
```diff
@@ -292,3 +285,88 @@ test("return from nested finally", () => {
...
+    util.testFunction`
+        class MyError extends Error {
+            name: "MyError"
+        }
+        
+        try {
+            throw new MyError();
+        } catch (error) {
+            if (error instanceof Error) {
+                return \'\$error\';
```
ark120202: Should it be `${error}` like in other cases?

test/unit/error.spec.ts
```diff
@@ -292,3 +285,88 @@ test("return from nested finally", () => {
...
+
+test.each(["Error", "RangeError", "ReferenceError", "SyntaxError", "TypeError", "URIError"])(
+    "throw builtin Errors as classes",
+    errorType => {
+        util.testFunction`
+            try {
+                throw new ${errorType}("message")
+            } catch (error) {
+                if (error instanceof Error) {
+                    return \`\${error}\`;
```
ark120202: It would be cleaner to just do `error.toString()` which is also transformed to `tostring(error)`

test/unit/error.spec.ts
ark120202: Unresolved it since it's still an issue.
Test probably didn't fail because of `return \'\$error\';`

src/lualib/Error.ts
ark120202: Looking into this more, I don't think there is a way to have a top-level error handler on all platforms. And anyway, depending on some extra thing user has to do for a quite basic feature like stacktraces doesn't sound like a good enough user experience.

I have played around it a bit, and this seems to consistently have stacks attached when thrown (tested on Lua 5.3 and JIT), but not when stringified manually:

```ts
const caller = debug.getinfo(3)
return _VERSION === "Lua 5.1" || (caller && caller.func !== error) ? `${this.name}: ${this.message}` : this.stack;
```

TheLartians: @Perryvw what are your thoughts on this?

src/lualib/Error.ts
TheLartians: I'm not sure that would work. How would TS know that a call method has been added by `setmetatable` without an explicit cast?

src/lualib/Error.ts
TheLartians: Actually I was thinking about adding `metatable.__tostring = metatable.__tostring || __TS__GetErrorString`. Why would that break inheritance?

test/unit/error.spec.ts
TheLartians: Thats actually a good idea. Especially `string` should be tested as I just noticed that Lua adds the source's location by default.

test/unit/error.spec.ts
TheLartians: Yep, good catch!

test/unit/error.spec.ts
TheLartians: Oh yeah I there was another one. Tests did't fail as we use `expectToMatchJsResult()` which doesn't care about the type definition.

src/lualib/Error.ts
TheLartians: Thinking some more about your code, I think it's actually a good way to handle error printing.

src/lualib/Error.ts
ark120202: Oh, I have assumed `setmetatable` returns `any`. Yeah, now it wouldn't work without a cast on returned value. It might make sense to add it to `setmetatable` signature instead (in `lua-types` if we'll switch to it).

src/lualib/Error.ts
ark120202: Because `ExtendedCustomError` prototype wouldn't have `__tostring` I think? I haven't tested it though, so if it works that's fine. We can add this as a test though

src/lualib/Error.ts
```diff
@@ -0,0 +1,63 @@
...
+    const caller = debug.getinfo(3, "f");
+    if (_VERSION === "Lua 5.1" || (caller && caller.func !== error)) {
+        return description;
+    } else {
+        return `${description}\n${errorObj.stack}`;
+    }
+}
+
+function __TS__InitErrorClass<T>(Type: ErrorType<T>, name: string): any {
+    if (name) {
```
ark120202: This condition can be removed now

test/unit/error.spec.ts
```diff
@@ -292,3 +285,109 @@ test("return from nested finally", () => {
...
+                throw Error("This error serves to prevent false positives from .expectNoExecutionError()");
+            }
+        }
+    `.expectToMatchJsResult();
+});
+
+test("extending from Error with custom toString()", () => {
+    util.testFunction`
+        class MyError extends Error {
+            toString(){ return "Custom error message";  }
```
ark120202:
```suggestion
            toString() {
            	return "Custom error message";
            }
```

test/unit/error.spec.ts
```diff
@@ -292,3 +285,109 @@ test("return from nested finally", () => {
...
+            if (error instanceof Error) {
+                return error.toString();
+            } else {
+                throw Error("This error serves to prevent false positives from .expectNoExecutionError()");
+            }
+        }
+    `.expectToMatchJsResult();
+});
+
+test.each(["Error", "RangeError", "MyError"])("get stack from error", errorType => {
```
ark120202:
```suggestion
test.each(["Error", "RangeError", "MyError"])("get stack from %s", errorType => {
```

test/unit/error.spec.ts
```diff
@@ -292,3 +285,109 @@ test("return from nested finally", () => {
...
+                } else {
+                    throw Error("This error serves to prevent false positives from .expectNoExecutionError()");
+                }
+            }
+        `.expectToMatchJsResult();
+    }
+);
+
+test.each(["Error", "RangeError", "ReferenceError", "SyntaxError", "TypeError", "URIError"])(
+    "throw builtin Errors as functions",
```
ark120202: `throwing as functions` doesn't sound as a good description. Actually IMO it would be better to test throwing anything and `Error` builtin separately.

So `throw and catch custom error object` would actually test that throw (and catch) pass anything as it is, and builtin class features are tested without any throws (i.e. just `return ${errorType}("message").toString()` there)

src/lualib/Error.ts
TheLartians: It works, because it's `__index` metatable has `__tostring`. tbh though I would rather remove this solution as we are currently over-engineering the error class to work around a more general issue of tstl's class system (see #728).

TheLartians:
I've rolled back to supporting `Error.toString()` only for the built-in error types as more general support for derived types will be available once #728 gets resolved.

src/lualib/Error.ts
ark120202: Actually, we'll have to hack it in any case, since user's `toString` implementation should have only description part. We'll have to wrap it with something like that:

```ts
function __TS__WrapErrorToString<T extends Error>(this: void, errorObj: T, getDescription: (this: T) => string): string {
    const description = getDescription.call(errorObj);
    const caller = debug.getinfo(3, "f");
    if (_VERSION === "Lua 5.1" || (caller && caller.func !== error)) {
        return description;
    } else {
        return `${description}\n${errorObj.stack}`;
    }
}

Error = __TS__InitErrorClass(
    class {
        public name = "Error";
        public stack: string;

        constructor(public message = "") {
            this.stack = __TS__GetErrorStack((this.constructor as any).new);
            const metatable = getmetatable(this);
            if (!metatable.__stackPatched) {
                metatable.__stackPatched = true;
                metatable.__tostring = __TS__WrapErrorToString(this, (this as any).__tostring);
            }
        }

        public toString(): string {
            return this.message !== "" ? `${this.name}: ${this.message}` : this.name;
        }
    },
    "Error"
);
```

test/unit/error.spec.ts
```diff
@@ -292,3 +285,94 @@ test("return from nested finally", () => {
...
+    const stack = util.testFunction`
+        class CustomError extends Error {
+            public name = "CustomError";
+        }
+
+        function innerFunctionThatThrows() { throw ${errorType}(); }
+        function outerFunctionThatThrows() { innerFunctionThatThrows(); }
+
+        try {
+            outerFunctionThatThrows();
```
ark120202: Stack catching is also not really related to throwing, so we can just return error object from these test functions

test/unit/error.spec.ts
```diff
@@ -292,3 +285,94 @@ test("return from nested finally", () => {
...
+    "new Error",
+    "new RangeError",
+    "new ReferenceError",
+    "new SyntaxError",
+    "new TypeError",
+    "new URIError",
+])("test builtin error properties", errorType => {
+    util.testFunction`
+        const error = ${errorType}();
+        return error.name;
```
ark120202:
```suggestion
        return { name: error.name, message: error.message, string: error.toString() };
```

test/unit/error.spec.ts
```diff
@@ -292,3 +285,94 @@ test("return from nested finally", () => {
     `;
     expect(util.transpileAndExecute(code)).toBe("finally AB");
 });
+
+test.each([
+    `"error string"`,
```
ark120202: Might be a bit cleaner to pass actual values there and convert them with util.valueToString

test/unit/error.spec.ts
```diff
@@ -292,3 +285,94 @@ test("return from nested finally", () => {
...
+                    return error();
+                } else {
+                    return error;
+                }
+            }
+        `.expectToMatchJsResult();
+});
+
+test.each([
+    "Error",
```
ark120202: Maybe

```ts
const errorTypes = ["Error", "RangeError", "ReferenceError", "SyntaxError", "TypeError", "URIError"];
test.each([...errorTypes, errorTypes.map(type => `new ${type}`)])
```

test/unit/error.spec.ts
TheLartians: Due to Lua's tail call optimisation, it actually won't work for simply returning the error object (from Lua's perspective the stack is at the outer call). We could work around this by storing the object in temporary variables, but I'd prefer to just keep the exception stack check.

test/unit/error.spec.ts
TheLartians: True, won't atm due to  `util.valueToString(undefined) = undefined` (#733).

test/unit/error.spec.ts
```diff
@@ -292,3 +285,57 @@ test("return from nested finally", () => {
...
+    `42`,
+    `3.141`,
+    `true`,
+    `false`,
+    `undefined`,
+    `{ x: "error object" }`,
+    `() => "error function"`,
+])("throw and catch arbitrary types", error => {
+    util.testFunction`
+            try {
```
ark120202: There should be one less indentation level

test/unit/error.spec.ts
```diff
@@ -292,3 +285,57 @@ test("return from nested finally", () => {
...
+        `.expectToMatchJsResult();
+});
+
+const builtinErrors = ["Error", "RangeError", "ReferenceError", "SyntaxError", "TypeError", "URIError"];
+
+test.each([...builtinErrors, ...builtinErrors.map(type => `new ${type}`)])(
+    "test builtin error properties",
+    errorType => {
+        util.testFunction`
+        const error = ${errorType}();
```
ark120202: And one more there

test/unit/error.spec.ts
```diff
@@ -292,3 +285,57 @@ test("return from nested finally", () => {
...
+        return { name: error.name, message: error.message, string: error.toString() };
+    `.expectToMatchJsResult();
+    }
+);
+
+test.each([...builtinErrors, `CustomError`])("get stack from %s", errorType => {
```
ark120202:
```suggestion
test.each([...builtinErrors, "CustomError"])("get stack from %s", errorType => {
```

test/unit/error.spec.ts
```diff
@@ -292,3 +285,54 @@ test("return from nested finally", () => {
...
+        }
+
+        let stack: string | undefined;
+
+        function innerFunction() { stack = new ${errorType}().stack; }
+        function outerFunction() { innerFunction(); }
+        outerFunction();
+        
+        return stack;
+        `.getLuaExecutionResult();
```
ark120202:
```suggestion
    `.getLuaExecutionResult();
```

test/unit/error.spec.ts
```diff
@@ -292,3 +285,54 @@ test("return from nested finally", () => {
...
+            }
+        }
+        `.expectToMatchJsResult();
+});
+
+const builtinErrors = ["Error", "RangeError", "ReferenceError", "SyntaxError", "TypeError", "URIError"];
+
+test.each([...builtinErrors, ...builtinErrors.map(type => `new ${type}`)])("test builtin error %s", errorType => {
+    util.testFunction`
+            const error = ${errorType}();
```
ark120202: One less now

test/unit/error.spec.ts
```diff
@@ -292,3 +285,54 @@ test("return from nested finally", () => {
...
+        try {
+            throw ${error};
+        } catch (error) {
+            if (typeof error == 'function') {
+                return error();
+            } else {
+                return error;
+            }
+        }
+        `.expectToMatchJsResult();
```
ark120202:
```suggestion
    `.expectToMatchJsResult();
```

test/unit/error.spec.ts
```diff
@@ -292,3 +285,54 @@ test("return from nested finally", () => {
...
+});
+
+const builtinErrors = ["Error", "RangeError", "ReferenceError", "SyntaxError", "TypeError", "URIError"];
+
+test.each([...builtinErrors, ...builtinErrors.map(type => `new ${type}`)])("test builtin error %s", errorType => {
```
ark120202: Maybe
```suggestion
test.each([...builtinErrors, ...builtinErrors.map(type => `new ${type}`)])("%s properties", errorType => {
```

PR #730, opened 2019-10-03
Added support for array.find

Perryvw:
Somehow we missed this and never noticed, but now it's in.


src/lualib/ArrayFind.ts
```diff
@@ -0,0 +1,14 @@
+// https://www.ecma-international.org/ecma-262/10.0/index.html#sec-array.prototype.find
+function __TS__ArrayFind<T>(this: void, arr: T[], callbackFn: (element: T, index?: number, array?: T[]) => boolean): T {
```
ark120202:
```suggestion
function __TS__ArrayFind<T>(this: void, arr: T[], callbackFn: (element: T, index: number, array: T[]) => unknown): T {
```

lolleko: and the return type should probably be `T | undefined`

PR #731, opened 2019-10-03
Fixed out of spec behavior of array.reduce

### LINKED ISSUE #727: Array.reduce should accept undefined as an initial value
```ts
console.log([].reduce(() => {}, undefined)); // should be undefined
```

[TypescriptToLua Playground](https://typescripttolua.github.io/play.html#src=console.log(%5B%5D.reduce(()%20%3D%3E%20%7B%7D%2C%20undefined))%3B)

[TS Playground](https://www.typescriptlang.org/play/#code/MYewdgziA2CmB00QHMAUBtAuvATrAJgK7CyqoCUABALwB8lA3gL4A0lhY+sAZgJZgFy5ANxA)

This is the more correct code roblox-ts uses:

```lua
function TS.array_reduce(list, callback, ...)
	local first = 1
	local last = #list
	local accumulator
	-- support `nil` initialValues
	if select("#", ...) == 0 then
		if last == 0 then
			error("Reduce of empty array with no initial value at Array.reduce", 2)
		end
		accumulator = list[first]
		first = first + 1
	else
		accumulator = ...
	end
	for i = first, last do
		accumulator = callback(accumulator, list[i], i - 1, list)
	end
	return accumulator
end

function TS.array_reduceRight(list, callback, ...)
	local first = #list
	local last = 1
	local accumulator
	-- support `nil` initialValues
	if select("#", ...) == 0 then
		if first == 0 then
			error("Reduce of empty array with no initial value at Array.reduceRight", 2)
		end
		accumulator = list[first]
		first = first - 1
	else
		accumulator = ...
	end
	for i = first, last, -1 do
		accumulator = callback(accumulator, list[i], i - 1, list)
	end
	return accumulator
end
```

Perryvw:
Fixes #727

### DIFF
--- a/src/lualib/ArrayReduce.ts
+++ b/src/lualib/ArrayReduce.ts
@@ -1,5 +1,7 @@
 /** @vararg */
 interface Vararg<T> extends Array<T> {}
+/** @forRange */
+declare function forRange(start: number, limit: number, step?: number): number[];
 
 // https://www.ecma-international.org/ecma-262/9.0/index.html#sec-array.prototype.reduce
 function __TS__ArrayReduce<T>(
@@ -26,9 +28,8 @@ function __TS__ArrayReduce<T>(
         k = 1;
     }
 
-    while (k < len) {
-        accumulator = callbackFn(accumulator, arr[k], k, arr);
-        k = k + 1;
+    for (const i of forRange(k, len - 1)) {
+        accumulator = callbackFn(accumulator, arr[i], i, arr);
     }
 
     return accumulator;
--- a/src/lualib/ArrayReduce.ts
+++ b/src/lualib/ArrayReduce.ts
@@ -1,5 +1,3 @@
-declare function select<T>(this: void, index: number, ...args: T[]): T;
-declare function select<T>(this: void, index: "#", ...args: T[]): number;
 /** @vararg */
 interface VarArg<T> extends Array<T> {}
 
--- a/src/lualib/ArrayReduce.ts
+++ b/src/lualib/ArrayReduce.ts
@@ -1,8 +1,3 @@
-/** @vararg */
-interface Vararg<T> extends Array<T> {}
-/** @forRange */
-declare function forRange(start: number, limit: number, step?: number): number[];
-
 // https://www.ecma-international.org/ecma-262/9.0/index.html#sec-array.prototype.reduce
 function __TS__ArrayReduce<T>(
     this: void,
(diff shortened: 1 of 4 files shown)


src/lualib/ArrayReduce.ts
```diff
@@ -1,20 +1,23 @@
+/** @vararg */
+interface VarArg<T> extends Array<T> {}
```
ark120202:
```suggestion
interface Vararg<T> extends Array<T> {}
```

src/lualib/ArrayReduce.ts
```diff
@@ -1,20 +1,23 @@
...
+    if (len === 0 && select("#", ...initial) === 0) {
         // tslint:disable-next-line: no-string-throw
         throw "Reduce of empty array with no initial value";
     }
 
     let k = 0;
-    let accumulator = initial;
-    if (initial === undefined) {
+    let accumulator = select(1, ...initial);
+    if (accumulator === undefined) {
```
ark120202: It doesn't look like that would fix that issue. The test case doesn't cover it, but:

```ts
['a', 'b'].reduce((a, b) => console.log(a, b));
// "a" "b"
['a', 'b'].reduce((a, b) => console.log(a, b), undefined);
// undefined "a"
// undefined "b"
```

src/lualib/ArrayReduce.ts
```diff
@@ -1,20 +1,23 @@
...
     this: void,
     arr: T[],
     callbackFn: (accumulator: T, currentValue: T, index: number, array: T[]) => T,
-    initial?: T
+    ...initial: VarArg<T>
 ): T {
     const len = arr.length;
 
-    if (len === 0 && initial === undefined) {
+    if (len === 0 && select("#", ...initial) === 0) {
```
ark120202: That's out of scope of this PR, but we probably want to transform `vararg.length` to `select("#", ...)` in general

src/lualib/ArrayReduce.ts
```diff
@@ -1,27 +1,35 @@
+/** @vararg */
+interface Vararg<T> extends Array<T> {}
```
ark120202: I think we should move it to something like declarations/tstl.d.ts

src/lualib/ArrayReduce.ts
```diff
@@ -1,27 +1,35 @@
...
     arr: T[],
     callbackFn: (accumulator: T, currentValue: T, index: number, array: T[]) => T,
-    initial?: T
+    ...initial: Vararg<T>
 ): T {
     const len = arr.length;
 
-    if (len === 0 && initial === undefined) {
+    const initialValuePresent = select("#", ...initial) !== 0;
+    if (len === 0 && !initialValuePresent) {
```
ark120202: Thoughts on moving that check to a condition below and removing `initialValuePresent` variable?

test/unit/builtins/array.spec.ts
```diff
@@ -486,6 +486,22 @@ test.each<[[(total: number, currentItem: number, index: number, array: number[])
...
+    util.testExpression`[].reduce(() => {}, undefined)`.expectToMatchJsResult();
+});
+
+test("array.reduce empty no initial", () => {
+    util.testExpression`[].reduce(() => {})`.expectToMatchJsResult(true);
+});
+
+test("array.reduce undefined returning callback", () => {
+    util.testFunction`
+        const calls: {a: void, b: string}[] = [];
```
ark120202:
```suggestion
        const calls: Array<{ a: void, b: string }> = [];
```

PR #733, opened 2019-10-05
Update valueToString for undefined values and fix failing tests

TheLartians:


test/util.ts
```diff
@@ -49,10 +49,15 @@ export function expectToBeDefined<T>(subject: T | null | undefined): subject is
...
 }
 
-export const valueToString = (value: unknown) =>
-    (typeof value === "number" && (!Number.isFinite(value) || Number.isNaN(value))) || typeof value === "function"
-        ? String(value)
-        : JSON.stringify(value);
+export function valueToString(value: unknown): string {
+    if (typeof value === "number" && (!Number.isFinite(value) || Number.isNaN(value))) {
+        return String(value);
+    } else if (typeof value === "function" || value === undefined) {
```
ark120202: Not a big issue, but grouping it that way looks kind of arbitrary. On a side note, we may consider using a full-blown solution like https://www.npmjs.com/package/javascript-stringify

TheLartians: I initially separated it as I wasn't happy the way prettier reformatted a large if statement but I agree this makes it look very arbitrary.

Perryvw:
Solved by #734

PR #734, opened 2019-10-06
Fixed issue in test code formatting, fixed array.splice and array.join

Perryvw:
Makes #733 obsolete.

- Merged test `util.valueToString` and `util.valuesToString` into single `util.formatCode`
- Fixed issue in case array.join was called with undefined argument.
- Fixed inconsistency between lua and js with regards to array.splice arguments.


src/LuaTransformer.ts
```diff
@@ -5134,8 +5134,18 @@ export class LuaTransformer {
             case "splice":
                 return this.transformLuaLibFunction(LuaLibFeature.ArraySplice, node, caller, ...params);
             case "join":
+                const defaultSeparatorLiteral = tstl.createStringLiteral(",");
                 const parameters =
```
```ts
                const parameters = [
                    caller,
                    node.arguments.length === 0
                        ? defaultSeparatorLiteral
                        : tstl.createBinaryExpression(params[0], defaultSeparatorLiteral, tstl.SyntaxKind.OrOperator),
                ];
```

src/lualib/ArraySplice.ts
```diff
@@ -1,6 +1,11 @@
-function __TS__ArraySplice<T>(this: void, list: T[], start: number, deleteCount: number, ...items: T[]): T[] {
+// https://www.ecma-international.org/ecma-262/9.0/index.html#sec-array.prototype.splice
+function __TS__ArraySplice<T>(this: void, list: T[], ...args: Vararg<unknown>): T[] {
     const len = list.length;
 
+    const actualArgumentCount = select("#", ...args);
+    const start = select(1, ...args) as number;
```
ark120202: Can't `start` be a regular argument? Though if it's just to keep `args` name semantics I'm fine with that

Perryvw: It cannot because of line 22, the ECMAScript spec prescribes a check 'if no arguments are provided', which includes `start`

Issue #262, opened 2018-10-23
Add custom Array type to lualib


lolleko:
This Array type should be used instead of Lua default table implementation to account for things like sparse arrays.

The added Array type will only be available via a compilerOption. 
If this options is not set, default Lua tables will be used for arrays.

This custom Array will have a rather big runtime performance impact, so it will be optional.

ark120202:
<details>
<summary>Unfinished version in Lua I worked on for my project</summary>

```Lua
function Array(self, ...)
  local args = {...}
  local max = select('#', ...)
  self.length = max

  for i = 1, max do
    self[i - 1] = args[i]
  end
end
Array.__js = true

function Array:from(arrayLike, mapFn)
  if arrayLike[Symbol.iterator] then
    arrayLike = arrayLike[Symbol.iterator](arrayLike)

    if arrayLike == nil or arrayLike.next == nil then
      error("Result of the Symbol.iterator method is not an iterable")
    end
  elseif arrayLike.next == nil then
    error("Array.from argument is not iterable")
  end

  local target = {}
  while true do
    local result = arrayLike.next()
    if result.done then
      break
    end
    if mapFn then
      result.value = mapFn(result.value)
    end
    table.insert(target, result.value)
  end

  return Reflect:__new(Array, Reflect:unpack(target))
end

Array.prototype[Symbol.iterator] = function(self)
  local i = -1
  return {
    next = function()
      i = i + 1

      return {
        value = self[i],
        done = i >= self.length
      }
    end
  }
end

function Array.prototype:forEach(callback)
  for i = 0, self.length - 1 do
    callback:call(nil, self[i], i, self)
  end
end

function Array.prototype:map(callback)
  local target = {}

  for i = 0, self.length - 1 do
    target[i + 1] = callback:call(nil, self[i], i, self)
  end

  return Reflect:__new(Array, Reflect:unpack(target))
end

function Array.prototype:filter(callback)
  local target = {}
  local n = 1

  for i = 0, self.length - 1 do
    if callback:call(nil, self[i], i, self) then
      target[n] = self[i]
      n = n + 1
    end
  end
  return Reflect:__new(Array, Reflect:unpack(target))
end

function Array.prototype:reduce(callback, initialValue)
  for i = 0, self.length - 1 do
    initialValue = callback:call(nil, initialValue, self[i], i, self)
  end

  return initialValue
end

function Array.prototype:every(callback)
  for i = 0, self.length - 1 do
[...]

PR #737, opened 2019-10-21
Fixed array.flat[Map] on [[]] (fixes #719).

### LINKED ISSUE #719: Array.flat doesn't flatten empty arrays
```ts
[[]].flat()
```

should be:

```
[]
```

in Lua:
```
{ 1: { } }
```

radekmie:
This change fixes a bug reported in #719: empty array elements (e.g. `[[]]`) were incorrectly handled by `array.flat` and `array.flatMap`.

### DIFF
--- a/src/lualib/ArrayFlat.ts
+++ b/src/lualib/ArrayFlat.ts
@@ -1,8 +1,11 @@
 function __TS__ArrayFlat(this: void, array: any[], depth = 1): any[] {
     let result: any[] = [];
     for (const value of array) {
-        // A hack to concat only true "arrays" and empty tables.
-        if (depth > 0 && type(value) === "table" && (1 in value || next(value, undefined) === undefined)) {
+        if (
+            depth > 0 &&
+            type(value) === "table" &&
+            (1 in value || (next as NextEmptyCheck)(value, undefined) === undefined)
+        ) {
             result = result.concat(__TS__ArrayFlat(value, depth - 1));
         } else {
             result[result.length] = value;
--- a/src/lualib/ArrayFlat.ts
+++ b/src/lualib/ArrayFlat.ts
@@ -1,7 +1,7 @@
 function __TS__ArrayFlat(this: void, array: any[], depth = 1): any[] {
     let result: any[] = [];
     for (const value of array) {
-        if (depth > 0 && type(value) === "table" && 1 in value) {
+        if (depth > 0 && type(value) === "table") {
             result = result.concat(__TS__ArrayFlat(value, depth - 1));
         } else {
             result[result.length] = value;
--- a/src/lualib/ArrayFlat.ts
+++ b/src/lualib/ArrayFlat.ts
@@ -4,6 +4,8 @@ function __TS__ArrayFlat(this: void, array: any[], depth = 1): any[] {
         if (
             depth > 0 &&
             type(value) === "table" &&
+            // Workaround to determine if value is an array or not (fails in case of objects without keys)
+            // See discussion in: https://github.com/TypeScriptToLua/TypeScriptToLua/pull/737
             (1 in value || (next as NextEmptyCheck)(value, undefined) === undefined)
         ) {
             result = result.concat(__TS__ArrayFlat(value, depth - 1));
--- a/src/lualib/ArrayFlat.ts
+++ b/src/lualib/ArrayFlat.ts
@@ -1,7 +1,8 @@
 function __TS__ArrayFlat(this: void, array: any[], depth = 1): any[] {
     let result: any[] = [];
     for (const value of array) {
-        if (depth > 0 && type(value) === "table") {
+        // A hack to concat only true "arrays" and empty tables.
+        if (depth > 0 && type(value) === "table" && (1 in value || next(value, undefined) === undefined)) {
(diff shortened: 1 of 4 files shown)


test/unit/builtins/array.spec.ts
```diff
@@ -487,6 +487,7 @@ test.each([
 });
 
 test.each([
+    { array: [[]], expected: [] },
```
Perryvw: Could you also add testcase `{ array: [{ a: 1 }, { a: 2 }, { a: 3 }], expected: [{ a: 1 }, { a: 2 }, { a: 3 }] },` to this list (passes on current master)

test/unit/builtins/array.spec.ts
```diff
@@ -497,6 +498,7 @@ test.each([
 });
 
 test.each([
+    { array: [[]], map: <T>(v: T) => v, expected: [] },
```
Perryvw: Could you also add testcase `{ array: [1, 2, 3], map: (v: number) => ({ a: v * 2 }), expected: [{ a: 2 }, { a: 4 }, { a: 6 }] },` to this list (passes on current master)

radekmie:
I've added these tests. Of course, they are failing. I'm just thinking, what is better to do here:
* Add a special case for an empty table (not very efficient).
* Add a `isArray`-like test (same as in [`__TS__ArrayConcat`](https://github.com/TypeScriptToLua/TypeScriptToLua/blob/53be6c32388d75e91a8ae1858bc0330094980372/src/lualib/ArrayConcat.ts#L7-L8)).

Perryvw:
The problem is there is not really a way to distinguish an empty table from an empty object in lua (they are both translated to `{}`). I am thinking we could assume for this case objects won't be empty and extend the existing check to something like `... && (1 in value || (next(value, undefined) == undefined))` To efficiently check the object has no keys. I would like to discuss your and my suggestions with some other contributors in the discord before making a final decision. I will come back to this after.

Perryvw:
Alright, we discussed it and came to the following conclusion: For now we would be fine if the way you solve this is by assuming that if an object has no keys, it is an array. This is not entirely correct, but there is no straightforward (or even a complex one) way to deal with this perfectly, so we would rather go for the very simple non-perfect fix than to go for a very complex non-perfect fix. So then the check could indeed be modified as I suggested above by adding a special case to the if statement. Please add a comment to the LuaLib explaining what the reasoning behind `(1 in value || (next(value, undefined) == undefined))` is and in which case it will fail (objects without keys in list).

radekmie:
It didn't help as `next(value, undefined) == undefined` (using `===` works the same) translates to

```lua
({ next(value, nil) }) == nil
```
and not, as expected, to
```lua
next(value, nil) == nil
```

Any idea why?

Perryvw:
That sounds very strange, can you just push your changes so I can have a look?

radekmie:
It seems to be caused by `@tupleReturn` directive on `next` definition - removing it fixes the problem (but it's there for a reason).

Perryvw:
That does explain the behavior. This turned out a bit more complicated than I expected.  To get rid of the wrapping we need a bit of typing. I feel like there must be a better way but for now you can put this in declarations/global.d.ts:
```ts
// Override next declaration so we can omit extra return values
type NextEmptyCheck<T> = (this: void, table: any, index: undefined) => unknown | undefined;
```
Then you change the check to
```ts
if (type(value) === "table" && (1 in value || (next as NextEmptyCheck)(value, undefined) === undefined)) {
```

src/lualib/ArrayFlat.ts
```diff
@@ -1,7 +1,8 @@
 function __TS__ArrayFlat(this: void, array: any[], depth = 1): any[] {
     let result: any[] = [];
     for (const value of array) {
-        if (depth > 0 && type(value) === "table" && 1 in value) {
+        // A hack to concat only true "arrays" and empty tables.
```
Perryvw:
```suggestion
        // Workaround to determine if value is an array or not (fails in case of objects without keys)
        // See discussion in: https://github.com/TypeScriptToLua/TypeScriptToLua/pull/737
```
(Also change the other comment)

lolleko:
I used `if (pcall(() => (arg as any[]).length)` in [ArrayConcat](https://github.com/TypeScriptToLua/TypeScriptToLua/blob/master/src/lualib/ArrayConcat.ts) to determine if a table is an array.
Maybe it could be used here as well.
Im not so sure tho since this new workaround might be better, because it doesn't rely on pcall.

In any case we should keep it consistent and use one of the methods for an isArray workaround across lualib.

radekmie:
I also thought about that at first, but `{a: 1}` also pass that `pcall` test. In general this may be a bigger problem, as `[].concat({a: 1}) === []`.

:+1: for a general `isArray`.

lolleko:
You are right! The array concat test do not include a case with an object... yikes.
So good thing we found that issue #738 here.

Anyway I think we should merge this and address isArray and array.concat in a different issue/pr.

Also related: https://github.com/TypeScriptToLua/TypeScriptToLua/issues/262

Perryvw:
I agree @lolleko, maybe we implement Array.isArray in the lualib and add some lualib dependencies? We can discuss it a bit more. That's a PR for another day.

PR #743, opened 2019-10-29
Refactor CLI file structure and fix `--project` path resolution on Windows

### LINKED ISSUE #736: Incorrect path normalization in TSTL.ts
Reproduces for me on windows, execute following commands in cmd

```cmd
mkdir tstl-bug-repro
cd tstl-bug-repro
npm init -yes
npm install -S typescript-to-lua
npm install -S typescript
npm install -S npx

git clone https://github.com/Perryvw/ExampleTSDotaAddon.git

REM Verify that tsconfig is present
type ExampleTSDotaAddon\game\scripts\tsconfig.json

REM Outputs a bunch of errors for me, that's ok
npx tsc -p ../tstl-bug-repro/ExampleTSDotaAddon/game/scripts/tsconfig.json

REM Outputs 'The specified path does not exist'
npx tstl -p ../tstl-bug-repro/ExampleTSDotaAddon/game/scripts/tsconfig.json
```

This seems to be caused by using posix.normalize when figuring out the tsconfig path in TSTL.ts

ark120202:
Fixes #736.


src/CommandLineParser.ts
```diff
@@ -256,3 +257,34 @@ export function createDiagnosticReporter(pretty: boolean, system = ts.sys): ts.D
         reporter(diagnostic);
     };
 }
+
+export function locateConfigFile(commandLine: ParsedCommandLine): ts.Diagnostic | string | undefined {
```
Perryvw: As discussed on discord, I don't think this belongs in here, especially since it's not even called form this file. I would like to either keep it close to where it is called, or maybe in a CLIHelpers.ts file or something similar.

src/utils.ts
```diff
@@ -0,0 +1 @@
+export const normalizeSlashes = (filePath: string) => filePath.replace(/\\/g, "/");
```
Perryvw: This also looks a little strange to be alone in this file. Maybe combine this and locateConfigFile into ProjectDirHelpers.ts?

ark120202: It can't go with `locateConfigFile` since it's also required for `Emit.ts`, and importing it from `cli` directory would kind of break isolation.

`src/utils.ts` for generic utils also exists in #705, so it wouldn't be alone.

PR #747, opened 2019-11-13
Playing with initial lua transformer setup

Perryvw:
Added initial idea for lua transformers setup with outFile bundler example/outline.


Perryvw:
Closing in favor of #750

PR #750, opened 2019-11-17
Alternative bundling

Perryvw:
Alternative approach to bundling from #648 and #747.

Took test cases from #648.

Instead of using outFile, instead uses `luaBundle` and `luaBundleEntry` tstl options.


src/bundle.ts
```diff
@@ -0,0 +1,59 @@
...
+
+function moduleTableNode(transpiledFiles: TranspiledFileWithSourceNode[]): SourceNode {
+    const tableHead = `__Lua_BundleModules = {\n`;
+    const tableEnd = `}\n`;
+    const tableEntries = transpiledFiles.map(moduleSourceNode);
+
+    return joinSourceChunks([tableHead, ...tableEntries, tableEnd]);
+}
+
+function mergeDeclarations(transpiledFiles: TranspiledFileWithSourceNode[]): string {
```
ark120202: This is almost guaranteed to generate incorrect code, I'd rather keep `declaration` undefined.

src/bundle.ts
```diff
@@ -0,0 +1,59 @@
+import { TranspiledFileWithSourceNode } from "./Transpile";
+import { SourceNode } from "source-map";
+
+const trimExtension = (fileName: string) => (fileName.endsWith(".ts") ? fileName.slice(0, -3) : fileName);
```
ark120202: There are a lot of other extensions TypeScript supports, we should trim all of them.
In refactor PR I've already moved `trimExt` from `Emit.ts` to `utils.ts` with `trimExtension` name.

src/bundle.ts
```diff
@@ -0,0 +1,59 @@
...
+const formatPath = (fileName: string) => trimExtension(fileName).replace("/", ".");
```
ark120202: Any reason we don't use the same code as we're using to resolve imports?

src/bundle.ts
```diff
@@ -0,0 +1,59 @@
...
+
+    if (transpiledFile.lua && transpiledFile.sourceMapNode) {
+        return joinSourceChunks([tableEntryHead, transpiledFile.sourceMapNode, tableEntryTail]);
+    } else {
+        return joinSourceChunks([tableEntryHead, tableEntryTail]);
+    }
+}
+
+function moduleTableNode(transpiledFiles: TranspiledFileWithSourceNode[]): SourceNode {
+    const tableHead = `__Lua_BundleModules = {\n`;
```
ark120202: `__Lua_BundleModules` looks quite inconsistent, can't we use names from https://github.com/TypeScriptToLua/TypeScriptToLua/pull/648/files#diff-7a7518949add62497c7fd54cdf9293e5?
 Also it probably should be local.

```suggestion
    const tableHead = `local ____modules = {\n`;
```

src/Transpile.ts
```diff
@@ -57,7 +61,11 @@ export function transpile({
     };
 
     if (options.noEmitOnError) {
-        const preEmitDiagnostics = [...program.getOptionsDiagnostics(), ...program.getGlobalDiagnostics()];
+        const preEmitDiagnostics = [
+            ...diagnostics,
```
ark120202: At this point it's always empty

Perryvw: What is? diagnostics can have diagnostics from validateOptions at line 51.

src/bundle.ts
```diff
@@ -0,0 +1,79 @@
...
+        const lualibBundle = emitHost.readFile(path.resolve(__dirname, "../dist/lualib/lualib_bundle.lua"));
+        moduleTableEntries.push(`["lualib_bundle"] = function() ${lualibBundle} end,\n`);
+    }
+
+    // Create ____modules table containing all entries from moduleTableEntries
+    const moduleTable = createModuleTableNode(moduleTableEntries);
+
+    // Override `require` to read from ____modules table.
+    const requireOverride =
+        `function require(file) if ____modules[file] then return ____modules[file]() ` +
```
ark120202: We should either fall back to original `require` if module is not found, or use a different function name

src/utils.ts
```diff
@@ -1 +1,14 @@
+import * as path from "path";
+
 export const normalizeSlashes = (filePath: string) => filePath.replace(/\\/g, "/");
+
+export const trimExt = (filePath: string) => filePath.slice(0, -path.extname(filePath).length);
```
ark120202: In #705 I've also renamed it to `trimExtension`

src/utils.ts
```diff
@@ -1 +1,14 @@
...
+
+export const formatPathToLuaPath = (filePath: string) => {
```
ark120202: Should it be an arrow function?

test/unit/outfile.spec.ts
```diff
@@ -0,0 +1,126 @@
+import * as util from "../util";
```
ark120202: It should be renamed to `bundl(e|ing).spec.ts`

test/util.ts
```diff
@@ -330,21 +333,41 @@ export abstract class TestBuilder {
         return this;
     }
 
-    public expectToHaveDiagnostics(): this {
-        expect(this.getLuaDiagnostics()).toHaveDiagnostics();
+    public expectToHaveDiagnostic(matcher: DiagnosticMatcher): this {
+        expect(this.getLuaDiagnostics().find(matcher)).toBeDefined();
```
ark120202: Is that really something we'd need that often to add a new method there?

Perryvw: We don't really have an alternative if we don't want to specify the exact diagnostic each time (which looks really messy and is not really required imo).

src/bundle.ts
```diff
@@ -0,0 +1,79 @@
...
+
+export function bundleTranspiledFiles(
+    bundleFile: string,
+    entryModule: string,
+    transpiledFiles: TranspiledFileWithSourceNode[],
+    emitHost: EmitHost
+): [Diagnostic[], TranspiledFileWithSourceNode] {
+    const diagnostics: Diagnostic[] = [];
+
+    if (transpiledFiles.find(f => f.fileName === entryModule) === undefined) {
```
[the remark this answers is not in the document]
Perryvw: I will add a full project integration test to catch this.

src/bundle.ts
```diff
@@ -0,0 +1,91 @@
...
+    return joinSourceChunks([tableHead, ...fileChunks, tableEnd]);
+}
+
+type SourceChunk = string | SourceNode;
+function joinSourceChunks(chunks: SourceChunk[]): SourceNode {
+    // tslint:disable-next-line:no-null-keyword
+    return new SourceNode(null, null, null, chunks);
+}
+
+const getProjectRootDir = (program: Program) => program.getCommonSourceDirectory();
```
ark120202: Shouldn't it take `options.rootDir` into account?

src/bundle.ts
```diff
@@ -0,0 +1,91 @@
...
+
+const resolveAbsolutePath = (program: Program, pathToResolve: string) =>
```
ark120202: Maybe `resolveFromRootDir`?

src/bundle.ts
```diff
@@ -0,0 +1,91 @@
...
+    path.isAbsolute(pathToResolve)
+        ? pathToResolve
+        : path.normalize(path.join(getProjectRootDir(program), pathToResolve));
```
ark120202:
```suggestion
        : normalizeSlashes(path.resolve(getProjectRootDir(program), pathToResolve));
```

test/util.ts
ark120202: I'd rather use diagnostic codes for that. In the future we should probably use an enum, or copy TypeScript's diagnostic messages concept to avoid magic numbers.

test/util.ts
```diff
@@ -235,24 +263,7 @@ export abstract class TestBuilder {
...
-                const returnType = to_jsstring(lua.lua_typename(L, lua.lua_type(L, -1)));
-                throw new Error(`Unsupported Lua return type: ${returnType}`);
-            }
-        } else {
-            // Filter out control characters appearing on some systems
-            const luaStackString = lua.lua_tostring(L, -1).filter(c => c >= 20);
-            const message = to_jsstring(luaStackString).replace(/^\[string "--\.\.\."\]:\d+: /, "");
-            return new ExecutionError(message);
-        }
+        return executeLua(code);
```
ark120202: IMO variable looks a bit unnecessary there, function name is already clear enough and there are no parameters that would make line look cluttered

test/util.ts
```diff
@@ -330,21 +341,41 @@ export abstract class TestBuilder {
...
     }
 
-    public expectToHaveNoDiagnostics(): this {
-        expect(this.getLuaDiagnostics()).not.toHaveDiagnostics();
+    public expectToHaveNoErrorDiagnostics(): this {
+        expect(this.getLuaDiagnostics()).not.toHaveErrorDiagnostics();
+        return this;
+    }
+
+    public expectExecutionError(message: string): this {
```
ark120202: Can't you already do `expectToEqual(new util.ExecutionError("message"))`?

test/unit/bundle.spec.ts
```diff
@@ -0,0 +1,131 @@
+import * as util from "../util";
```
ark120202: I think it also should be moved to `transpile`

test/transpile/bundle/tsconfig.json
```diff
@@ -0,0 +1,16 @@
+{
+    "compilerOptions": {
+        "strict": true,
+        "noUnusedLocals": true,
+        "noUnusedParameters": true,
+        "target": "es2017",
+        "lib": ["es2017"],
```
ark120202:
```json
        "target": "esnext",
        "lib": ["esnext"],
```

test/transpile/basic.spec.ts
```diff
@@ -0,0 +1,18 @@
+import * as path from "path";
```
ark120202: I'm kind of missing the point of that test, don't we already test all of that stuff in other tests? Also IMO it tests too much things

test/cli/parse.spec.ts
```diff
@@ -166,39 +169,68 @@ describe("tsconfig", () => {
...
+            ["luaLibImport", "require", { luaLibImport: tstl.LuaLibImportKind.Require }],
+
+            ["luaTarget", "5.1", { luaTarget: tstl.LuaTarget.Lua51 }],
+            ["luaTarget", "5.2", { luaTarget: tstl.LuaTarget.Lua52 }],
+            ["luaTarget", "5.3", { luaTarget: tstl.LuaTarget.Lua53 }],
+            ["luaTarget", "jit", { luaTarget: tstl.LuaTarget.LuaJIT }],
+
+            ["luaBundle", "foo", { luaBundle: "foo" }],
+            ["luaBundleEntry", "bar", { luaBundleEntry: "bar" }],
+        ])("--%s %s", (optionName, value, expected) => {
```
ark120202:
```suggestion
        ])("{ %p: %p }", (optionName, value, expected) => {
```

src/CompilerOptions.ts
```diff
@@ -41,3 +43,42 @@ export enum LuaTarget {
...
+    }
+
+    if (options.luaBundle && options.luaLibImport === LuaLibImportKind.Inline) {
+        diagnostics.push(
+            configWarningDiagnostic(
+                `Using 'luaBundle' with 'luaLibImport: "inline"' might generate duplicate code. ` +
+                    `It is recommended to use 'luaLibImport: "require"'`
+            )
+        );
+    }
```
ark120202: Can you think of any cases where someone actually would want to keep using inline? I'm thinking we might just error in such case

src/CompilerOptions.ts
```diff
@@ -41,3 +43,42 @@ export enum LuaTarget {
     Lua53 = "5.3",
     LuaJIT = "JIT",
 }
+
+export function validateOptions(options: CompilerOptions): ts.Diagnostic[] {
+    const diagnostics: ts.Diagnostic[] = [];
+
+    if (options.luaBundle && (options.luaBundleEntry === "" || !options.luaBundleEntry)) {
```
ark120202: `!options.luaBundleEntry` already checks for empty string

src/CompilerOptions.ts
```diff
@@ -41,3 +43,42 @@ export enum LuaTarget {
...
+                `Using 'luaBundle' with 'luaLibImport: "inline"' might generate duplicate code. ` +
+                    `It is recommended to use 'luaLibImport: "require"'`
+            )
+        );
+    }
+
+    return diagnostics;
+}
+
+const configErrorDiagnostic = (message: string): ts.Diagnostic => ({
```
ark120202: IMO diagnostic factories should be specified, that would look better once we'll have diagnostic codes

src/bundle.ts
```diff
@@ -0,0 +1,91 @@
+import * as path from "path";
+import { TranspiledFile, EmitHost } from "./Transpile";
+import { SourceNode } from "source-map";
+import { formatPathToLuaPath, trimExtension, normalizeSlashes } from "./utils";
+import { Diagnostic, Program } from "typescript";
```
ark120202: For consistency:

```suggestion
import * as ts from "typescript";
```

src/Transpile.ts
```diff
@@ -1,11 +1,13 @@
...
-import { CompilerOptions } from "./CompilerOptions";
+import { CompilerOptions, validateOptions } from "./CompilerOptions";
 import * as diagnosticFactories from "./diagnostics";
 import { Block } from "./LuaAST";
 import { LuaPrinter } from "./LuaPrinter";
 import { LuaTransformer } from "./LuaTransformer";
 import { TranspileError } from "./TranspileError";
 import { getCustomTransformers } from "./TSTransformers";
+import { bundleTranspiledFiles } from "./bundle";
+import { SourceNode } from "source-map";
```
ark120202: Can you run TypeScript's `Organize Imports` command (Ctrl+Shift+O in vscode)?

src/Emit.ts
```diff
@@ -17,7 +15,7 @@ export function emitTranspiledFiles(
     transpiledFiles: TranspiledFile[],
     emitHost: EmitHost = ts.sys
 ): OutputFile[] {
-    let { rootDir, outDir, outFile, luaLibImport } = options;
+    let { rootDir, outDir, luaLibImport } = options;
 
     const configFileName = options.configFilePath as string | undefined;
     // TODO: Use getCommonSourceDirectory
```
ark120202: Can't we use `getProjectRootDir` now? (for `rootDir = rootDir || baseDir`, can't leave comment on that line)

src/bundle.ts
```diff
@@ -0,0 +1,91 @@
...
+        moduleTableEntries.push(`["lualib_bundle"] = function() ${lualibBundle} end,\n`);
+    }
+
+    // Create ____modules table containing all entries from moduleTableEntries
+    const moduleTable = createModuleTableNode(moduleTableEntries);
+
+    // Override `require` to read from ____modules table.
+    const requireOverride =
+        `local ____originalRequire = require\n` +
+        `function require(file) if ____modules[file] then return ____modules[file]() ` +
```
ark120202: It also should cache imported modules

src/bundle.ts
```diff
@@ -0,0 +1,91 @@
...
+        `else print("Could not find module '"..file.."' to require."); return ____originalRequire(file) end end\n`;
```
ark120202:
```suggestion
        `else return ____originalRequire(file) end end\n`;
```

src/cli/parse.ts
```diff
@@ -21,9 +21,23 @@ interface CommandLineOptionOfBoolean extends CommandLineOptionBase {
     type: "boolean";
 }
 
-type CommandLineOption = CommandLineOptionOfEnum | CommandLineOptionOfBoolean;
+interface CommandLineOptionOfString extends CommandLineOptionBase {
+    type: "string";
```
ark120202: https://github.com/TypeScriptToLua/TypeScriptToLua/pull/648#discussion_r335312271

src/CompilerOptions.ts
Perryvw: We use it on the playground

src/Transpile.ts
Perryvw: Can we let prettier do this?

test/transpile/basic.spec.ts
Perryvw: I want to test this since this is the main use-case of the tool, even if parts of this are already tested, I don't think there is any other test that actually transpiles a project from disk (except bundle is separate since it has a different project setup).

src/bundle.ts
Perryvw: Not sure if this should be removed. Imagine an environment where require is nil (so we need bundling), you will get an 'attempted to call nil value' error, which is really unclear.

src/Transpile.ts
ark120202: Unfortunately now there's no good way to enforce that.
Prettier never does changes that may change something in runtime.
ESLint currently doesn't have a rule that follows the same order.
TSLint's [ordered-imports](https://palantir.github.io/tslint/rules/ordered-imports/) is close, but it only sorts them, while `Organize Imports` also removes unneeded ones, removes formatting and handles `export from`.
There's also an option to utilize original command with a [CLI](https://www.npmjs.com/package/organize-imports-cli), but I'm not sure if we want to add a new tool.

src/bundle.ts
ark120202: Well, if that's a problem we just issue an error when `____originalRequire` is nil. Falling back to original require is important when there are environment-provided modules, and we shouldn't do any IO in these cases

src/resolve.ts
```diff
@@ -0,0 +1,15 @@
+import * as path from "path";
+import * as ts from "typescript";
+import { normalizeSlashes } from "./utils";
+
+export function getProjectRootDir(program: ts.Program): string {
+    const options = program.getCompilerOptions();
+    const projectDir = options.configFilePath || program.getCommonSourceDirectory();
+
+    return normalizeSlashes(options.rootDir ? path.resolve(projectDir, options.rootDir) : projectDir);
```
ark120202: When `rootDir` is specified it always should be resolved from config file, common source directory should be used as a fallback only when it's not defined at all

test/cli/parse.spec.ts
```diff
@@ -115,10 +115,13 @@ describe("command line", () => {
             ["luaTarget", "5.2", { luaTarget: tstl.LuaTarget.Lua52 }],
             ["luaTarget", "5.3", { luaTarget: tstl.LuaTarget.Lua53 }],
             ["luaTarget", "jit", { luaTarget: tstl.LuaTarget.LuaJIT }],
-        ])("--%s %s", (optionName, value, expected) => {
+
+            ["luaBundle", "foo", { luaBundle: "foo" }],
+            ["luaBundleEntry", "bar", { luaBundleEntry: "bar" }],
+        ])("{ %p: %p }", (optionName, value, expected) => {
```
ark120202: In that case I think it looks better because of similarity to actual command line
```suggestion
        ])("--%s %s", (optionName, value, expected) => {
```

src/Emit.ts
```diff
@@ -11,18 +12,15 @@ export interface OutputFile {
 
 let lualibContent: string;
 export function emitTranspiledFiles(
+    program: ts.Program,
     options: CompilerOptions,
```
ark120202: Since we pass program now, we can just get options from it

src/Emit.ts
```diff
@@ -11,18 +12,15 @@ export interface OutputFile {
...
+    let { outDir, luaLibImport } = options;
 
-    const configFileName = options.configFilePath as string | undefined;
-    // TODO: Use getCommonSourceDirectory
-    const baseDir = configFileName ? path.dirname(configFileName) : process.cwd();
-
-    rootDir = rootDir || baseDir;
-    outDir = outDir ? path.resolve(baseDir, outDir) : rootDir;
+    const rootDir = getProjectRootDir(program);
+    outDir = outDir ? path.resolve(rootDir, outDir) : rootDir;
```
ark120202: `outDir` should be resolved from config (though we probably don't need to resolve it at all, TypeScript resolves path during parsing)

src/resolve.ts
```diff
@@ -0,0 +1,8 @@
+import * as path from "path";
+import * as ts from "typescript";
+import { normalizeSlashes } from "./utils";
+
+export const resolveFromRootDir = (program: ts.Program, pathToResolve: string) =>
```
ark120202: Do we really need a new file for that? Looks like it's used only in bundle

src/bundle.ts
```diff
@@ -0,0 +1,87 @@
...
+    if (!transpiledFiles.some(f => resolveFromRootDir(program, f.fileName) === resolvedEntryModule)) {
+        return [[couldNotFindBundleEntryPoint(entryModule)], { fileName: bundleFile }];
+    }
+
+    // For each file: ["<file name>"] = function() <lua content> end,
+    const projectRootDir = program.getCommonSourceDirectory();
+    const moduleTableEntries: SourceChunk[] = transpiledFiles.map(f => moduleSourceNode(f, projectRootDir));
+
+    // If any of the modules contains a require for lualib_bundle, add it to the module table.
+    const lualibRequired = transpiledFiles.some(f => f.lua && f.lua.match(/require\("lualib_bundle"\)/));
```
ark120202:
```suggestion
    const lualibRequired = transpiledFiles.some(f => f.lua && f.lua.includes('require("lualib_bundle")');
```

src/bundle.ts
```diff
@@ -0,0 +1,87 @@
...
+        const lualibBundle = emitHost.readFile(path.resolve(__dirname, "../dist/lualib/lualib_bundle.lua"));
+        moduleTableEntries.push(`["lualib_bundle"] = function() ${lualibBundle} end,\n`);
+    }
+
+    // Create ____modules table containing all entries from moduleTableEntries
+    const moduleTable = createModuleTableNode(moduleTableEntries);
+
+    // Override `require` to read from ____modules table.
+    const requireOverride =
+        `local ____moduleCache = {}\n` +
```
ark120202: I think a multiline literal would look better there, even with incorrect indention

src/bundle.ts
```diff
@@ -0,0 +1,87 @@
...
+        `local ____originalRequire = require\n` +
+        `function require(file) if ____moduleCache[file] then return ____moduleCache[file] end\n` +
+        `if ____modules[file] then ____moduleCache[file] = ____modules[file](); return ____moduleCache[file] ` +
+        `else print("Could not find module '"..file.."' to require."); return ____originalRequire(file) end end\n`;
+    const entryPoint = `return require("${formatPath(entryModule)}")\n`;
```
ark120202:
```suggestion
    const entryPoint = `return require("${escapeString(formatPath(entryModule))}")\n`;
```

src/bundle.ts
```diff
@@ -0,0 +1,87 @@
...
+export function bundleTranspiledFiles(
+    bundleFile: string,
+    entryModule: string,
+    transpiledFiles: TranspiledFile[],
+    program: ts.Program,
+    emitHost: EmitHost
+): [ts.Diagnostic[], TranspiledFile] {
+    const diagnostics: ts.Diagnostic[] = [];
+
+    const resolvedEntryModule = resolveFromRootDir(program, entryModule);
```
ark120202: How about
```suggestion
    const entryModulePath = resolveFromRootDir(program, entryModule);
```

src/bundle.ts
ark120202: Also, it should use the same logic as one used in `moduleSourceNode`, with `resolvedEntryModule`

src/bundle.ts
ark120202: Also, all paths specified in options should be resolved from project directory, not rootDir

src/bundle.ts
```diff
@@ -0,0 +1,87 @@
...
+    if (!transpiledFiles.some(f => resolveFromRootDir(program, f.fileName) === resolvedEntryModule)) {
```
ark120202: Is `resolveFromRootDir` required? Isn't `fileName` always resolved and normalized?

Perryvw: Don't think it is when you manually construct the program (like we do in our tests)

src/bundle.ts
```diff
@@ -0,0 +1,109 @@
+import * as path from "path";
+import { SourceNode } from "source-map";
+import * as ts from "typescript";
+import { couldNotFindBundleEntryPoint } from "./diagnostics";
+import { EmitHost, TranspiledFile } from "./Transpile";
+import { formatPathToLuaPath, trimExtension } from "./utils";
+import { escapeString } from "./TSHelper";
+import { CompilerOptions } from "./CompilerOptions";
+
+const formatPath = (path: string) => escapeString(formatPathToLuaPath(trimExtension(path)));
```
ark120202: It's used only in `modulePath` now, so I think it's better to just inline it

src/bundle.ts
```diff
@@ -0,0 +1,109 @@
...
+
+    // Create ____modules table containing all entries from moduleTableEntries
+    const moduleTable = createModuleTableNode(moduleTableEntries);
+
+    // Override `require` to read from ____modules table.
+    const requireOverride = `
+local ____moduleCache = {}
+local ____originalRequire = require
+function require(file)
+    if ____moduleCache[file] then return ____moduleCache[file] end
```
ark120202: Would be nice to follow printer's formatting
```suggestion
    if ____moduleCache[file] then
        return ____moduleCache[file]
    end
```

src/bundle.ts
```diff
@@ -0,0 +1,109 @@
...
+    if ____modules[file] then
+        ____moduleCache[file] = ____modules[file]()
+        return ____moduleCache[file]
+    else
+        print("Could not find module '"..file.."' to require.")
```
ark120202: Formatting:
```suggestion
        print("Could not find module '" .. file .. "' to require.")
```

src/bundle.ts
```diff
@@ -0,0 +1,109 @@
...
+
+    const bundleNode = joinSourceChunks([moduleTable, requireOverride, entryPoint]);
+    const { code, map } = bundleNode.toStringWithSourceMap();
+
+    return [
+        diagnostics,
+        {
+            fileName: resolvedBundleFile,
+            lua: code,
+            sourceMap: map.toString(),
```
ark120202: Just to note, currently it handles inline source maps and traceback override incorrectly, but we'll have to change it for module support anyway, so I'm okay with that for now

src/bundle.ts
```diff
@@ -0,0 +1,109 @@
...
+            sourceMapNode: moduleTable,
+            declaration: undefined,
+            declarationMap: undefined,
```
ark120202: Maybe just remove these properties?

src/bundle.ts
```diff
@@ -0,0 +1,109 @@
...
+import { SourceNode } from "source-map";
+import * as ts from "typescript";
+import { couldNotFindBundleEntryPoint } from "./diagnostics";
+import { EmitHost, TranspiledFile } from "./Transpile";
+import { formatPathToLuaPath, trimExtension } from "./utils";
+import { escapeString } from "./TSHelper";
+import { CompilerOptions } from "./CompilerOptions";
+
+const formatPath = (path: string) => escapeString(formatPathToLuaPath(trimExtension(path)));
+const modulePath = (baseDir: string, pathToResolve: string) => formatPath(path.relative(baseDir, pathToResolve));
```
ark120202: I think that name sounds too variabl-y (`moduleSourceNode` already has a parameter with the same name). Also, conceptually it's close to `getExportPath` from tshelper, which is used only in tests now. Maybe we can just update it's implementation and use it instead?

test/transpile/bundle.spec.ts
```diff
@@ -0,0 +1,19 @@
...
+
+test("should transpile into one file", () => {
+    const transpileResult = transpileProject(inputProject);
+
+    expect(transpileResult.diagnostics).not.toHaveErrorDiagnostics();
+    expect(transpileResult.emitResult.length).toBe(1);
+
+    const { name, text } = transpileResult.emitResult[0];
+    // Verify the name is as specified in tsconfig
+    expect(path.join(name)).toBe(path.join(projectDir, "bundle.lua"));
```
ark120202: That `path.join(name)` looks a bit hacky, maybe instead we may have a helper in `run.ts`, like `buildVirtualProject`, but with a config file? It also would make paths relative, so we'll be able to simplify `basic.spec.ts` as well.

test/transpile/basic.spec.ts
ark120202: We have few tests with `tsconfig.json` in CLI, but yeah, we don't have any tests using `transpileProject` API. Maybe at least we may reduce scope of that test? I.e. rename it to `project` and remove that imaginary host API (which we already test)?

test/transpile/bundle.spec.ts
Perryvw: That should not have been there, I removed it.

src/bundle.ts
```diff
@@ -0,0 +1,112 @@
...
+        return ____moduleCache[file]
+    end
+    if ____modules[file] then
+        ____moduleCache[file] = ____modules[file]()
+        return ____moduleCache[file]
+    else
+        if ____originalRequire then
+            return ____originalRequire(file)
+        else
+            print("Could not find module '" .. file .. "' to require.")
```
ark120202:
```suggestion
            error("module '" .. file .. "' not found")
```

test/util.ts
```diff
@@ -330,24 +340,44 @@ export abstract class TestBuilder {
...
+        expect(this.getLuaDiagnostics().find(matcher)).toBeDefined();
+        return this;
+    }
+
+    public expectToHaveExactDiagnostic(diagnostic: ts.Diagnostic): this {
+        expect(this.getLuaDiagnostics()).toContainEqual(diagnostic);
+        return this;
+    }
+
+    public expectToHaveErrorDiagnostics(): this {
```
ark120202: Also should be reverted

test/util.ts
```diff
@@ -330,24 +340,44 @@ export abstract class TestBuilder {
...
         return this;
     }
 
-    public expectToHaveNoDiagnostics(): this {
+    public expectToHaveNoErrorDiagnostics(): this {
         expect(this.getLuaDiagnostics()).not.toHaveDiagnostics();
         return this;
     }
 
+    /*public expectExecutionError(message: string): this {
```
ark120202: I guess that should be removed

src/bundle.ts
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    // Create ____modules table containing all entries from moduleTableEntries
+    const moduleTable = createModuleTableNode(moduleTableEntries);
+
+    // Override `require` to read from ____modules table.
+    const requireOverride = `
+local ____moduleCache = {}
+local ____originalRequire = require
+function require(file)
```
wyozi: Just curious, is there a specific reason for this not being local? It would still equivalently be used by the whole file, but with the benefit of not leaking outside this file

Perryvw: No specific reason, we did not consider this at all. I think it is indeed a good idea to make this adjustment.

Perryvw: Addressed in #757

PR #752, opened 2019-11-25
Implement Number.prototype.toString(radix)

### LINKED ISSUE #723: Number.toString() doesn't support radix
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toString

```
numObj.toString([radix])
```

ark120202:
Resolves #723.

### DIFF
--- a/src/lualib/NumberToString.ts
+++ b/src/lualib/NumberToString.ts
@@ -1,0 +1,48 @@
+// tslint:disable-next-line: variable-name
+const ____radixChars = "0123456789abcdefghijklmnopqrstuvwxyz";
+
+function __TS__NumberToString(this: number, radix?: number): string {
+    if (radix === undefined || radix === 10 || this === Infinity || this === -Infinity || this !== this) {
+        return this.toString();
+    }
+
+    radix = Math.floor(radix);
+    if (radix < 2 || radix > 36) {
+        // tslint:disable-next-line: no-string-throw
+        throw "toString() radix argument must be between 2 and 36";
+    }
+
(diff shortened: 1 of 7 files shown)


test/unit/builtins/numbers.spec.ts
```diff
@@ -48,6 +48,18 @@ describe("Number", () => {
     });
 });
 
+const toStringRadixes = [undefined, 10, 2, 8, 9, 16, 17, 36, 36.9];
+const toStringValues = [...numberCases, 1024, 1.2, NaN];
+// TODO: flatMap
+const toStringPairs = toStringValues.reduce<Array<readonly [number, number | undefined]>>(
```
Perryvw: This is very hard to read, maybe we could make some helper function `combinations` or `testCombinations`. Also do we really need to check `NaN` for every radix value? Might be nicer as a separate test.

src/lualib/NumberToString.ts
```diff
@@ -0,0 +1,48 @@
+// tslint:disable-next-line: variable-name
+const ____radixChars = "0123456789abcdefghijklmnopqrstuvwxyz";
+
+function __TS__NumberToString(this: number, radix?: number): string {
+    if (radix === undefined || radix === 10 || this === Infinity || this === -Infinity || this !== this) {
+        return this.toString();
+    }
+
+    radix = Math.floor(radix);
```
Perryvw: Did you use some kind of reference for this implementation? If so, could you add a comment with a link.

PR #762, opened 2019-12-07
Drop `var` declarations support

### LINKED ISSUE #703: Drop `var` declarations
- We [don't handle it correctly now](https://typescripttolua.github.io/play.html#src=function%20foo()%20%7B%0A%20%20%20%20%7B%20var%20x%20%3D%201%3B%20%7D%0A%20%20%20%20console.log(x)%3B%0A%7D)
- It's pretty much deprecated in modern JS and especially TS
- `var` scoping may confuse Lua developers that are new to JS/TS
- Apparently, we'll be able to remove a lot of code:
	- Special `var` handling in `hoistVariableDeclarations`
	- ...which is the only place that uses `replaceStatementInParent`
	- ...which is the only place (except printer) that uses tstl.Node.parent
	- ...which is the main reason why nodes are mutated and the reason to use `cloneNode`

### LINKED ISSUE #761: Incorrect nested namespace hoisting
[Playground](https://typescripttolua.github.io/play.html#src=let%20Inner%20%3D%201%3B%0A%0Aexport%20namespace%20Outer%20%7B%0A%20%20%20%20export%20function%20bar()%20%7B%0A%20%20%20%20%20%20%20%20return%20Inner.foo%3B%0A%20%20%20%20%7D%0A%20%20%20%20namespace%20Inner%20%7B%0A%20%20%20%20%20%20%20%20export%20let%20foo%20%3D%20%22foo%22%3B%0A%20%20%20%20%7D%0A%7D%0A%0Aconsole.log(Inner)%3B%0A)

`local Inner` is hoisted to the file level scope, but is expected to be hoisted to namespace scope.

ark120202:
Resolves #703, fixes #761.


test/unit/hoisting.spec.ts
```diff
@@ -147,17 +110,20 @@ test("Hoisting due to reference from hoisted function", () => {
...
+            }
         }
-        export const foo = bar();
-    `;
-    const result = util.transpileExecuteAndReturnExport(code, "foo");
-    expect(result).toBe("foo");
+
+        export const foo = Outer.bar();
+        export { Inner };
+    `.expectToMatchJsResult();
```
Perryvw: This test seems slightly different from before, maybe we should keep both?

ark120202: Oh, I think I actually wanted to change a different test (`Nested Namespace Hoisting`)

PR #764, opened 2019-12-08
fix back assignment

### LINKED ISSUE #430: Support back assignment for variables exported with ExportDeclaration
```ts
let x = false;
export { x };
x = true;
```

Exported value should be `true`.

Since ExportDeclaration's ~may appear only on the top level~ symbols do exist in the `scopeSymbol.exports` that would be easy to implement.

hazzard993:
Closes #430

```ts
let x = false;
export { x };
x = true;
```

_x_ should now be `true`.


src/transformation/visitors/binary-expression/assignments.ts
```diff
@@ -1,15 +1,27 @@
...
+export function transformAssignmentLeftHandSideExpression(
+    context: TransformationContext,
+    node: ts.Expression
+): lua.AssignmentLeftHandSideExpression {
+    const symbol = context.checker.getSymbolAtLocation(node);
+    const left = context.transformExpression(node);
+
+    return lua.isIdentifier(left) && symbol && isSymbolExported(context, symbol)
+        ? createExportedIdentifier(context, left)
+        : (left as lua.AssignmentLeftHandSideExpression);
```
ark120202:
```suggestion
        : cast(left, lua.isAssignmentLeftHandSideExpression);
```
We should leave a runtime check there, because it's not enforced in visitor types

src/transformation/utils/export.ts
```diff
@@ -98,9 +98,7 @@ export function isSymbolExportedFromScope(
         return false;
     }
 
-    // ts.Iterator is not a ES6-compatible iterator, because TypeScript targets ES5
-    const it: Iterable<ts.Symbol> = { [Symbol.iterator]: () => scopeSymbol!.exports!.values() };
-    return [...it].includes(symbol);
+    return scopeSymbol.exports.has(symbol.escapedName);
```
ark120202: Would it work in such case?
```ts
export let a = 1;
(() => {
    let a = 2;
    a = 3;
})();
```

hazzard993: nope! +1 new test case

src/transformation/utils/export.ts
```diff
@@ -98,9 +98,11 @@ export function isSymbolExportedFromScope(
         return false;
     }
 
-    // ts.Iterator is not a ES6-compatible iterator, because TypeScript targets ES5
-    const it: Iterable<ts.Symbol> = { [Symbol.iterator]: () => scopeSymbol!.exports!.values() };
-    return [...it].includes(symbol);
+    if (symbol.declarations.find(d => findFirstNodeAbove(d, ts.isBlock))) {
```
ark120202: Having a block above isn't enough to match symbol identities there

```ts
export let a = 1;
(() => {
    a = 2;
})();
```

```ts
export let a = 1;
switch (0 as any) {
    case 1:
        let a = 2;
    case 2:
        a = 3
}
```

ark120202:
Actually, that's a much more complex problem than I've expected.

Initially I've thought that exported symbol would have the same reference as local one, and so it would be possible to just use the old symbol matching logic there, but looking at it more, it actually makes sense that they are different. For example, it should be possible to give exported symbol a different name with `{ foo as bar }`. Or even have multiple exported symbols for a single local one (`{ foo as foo1, foo as foo2 }`). To support that correctly we'd have to make quite huge changes to our export logic.

I'm not sure what would be the best way to go with this issue for now. If we're okay to go with incomplete (but better than nothing) support, I think the simplest thing would be to move export statements to the bottom of generated file. At least it would be enough to support the example from OP.

src/transformation/utils/export.ts
hazzard993: No longer using `ts.isBlock`

But I am curious what is expected to happen in a switch statement like this

hazzard993:
I think I managed to get around it without changing the existing export logic too much

It should now be possible to export a variable under numerous different names with export specifiers

When reassigning the variable synchronisation should still occur

```ts
let x = false;
export { x as a };
export { x as b };
x = true; // a and b now become true
```

```ts
x = true;
// becomes `exports.b = exports.a = x = true;` in TypeScript -> JavaScript
// becomes `____exports.b = <iife chain>` in TypeScript -> Lua
```

test/unit/modules/modules.spec.ts
```diff
@@ -209,3 +209,56 @@ test("Export Equals", () => {
...
+test.each(reassignmentTestCases)("export specifier fork (%p)", reassignment => {
+    util.testModule`
+        let x: any = false;
+        export { x as a };
+        export { x as b };
+        ${reassignment};
+    `.expectToMatchJsResult();
+});
+
+test.each(["(() => { let a = 2; a = 3 })()", "{ let a = 2; a = 3 }"])(
```
ark120202: Do we really need to test both of these, given how similar syntactically they are?

src/transformation/visitors/binary-expression/assignments.ts
```diff
@@ -113,6 +151,32 @@ export function transformAssignmentExpression(
     }
 }
 
+const canBeTransformedToLuaAssignmentStatement = (
+    context: TransformationContext,
+    expression: ts.DestructuringAssignment
+): expression is ts.ArrayDestructuringAssignment => {
+    return (
+        ts.isArrayLiteralExpression(expression.left) &&
+        expression.left.elements.every(expression => {
```
ark120202: The `expression` name is a bit overloaded on that line.

src/transformation/visitors/binary-expression/assignments.ts
```diff
@@ -1,15 +1,44 @@
...
+    context: TransformationContext,
+    rootAssignment: lua.AssignmentStatement,
+    symbols: ts.Symbol[]
+): lua.AssignmentStatement {
+    return symbols.reduce<lua.AssignmentStatement>((statement, symbol) => {
+        const [left] = statement.left;
+        const assignmentFunction = createImmediatelyInvokedFunctionExpression([statement], left);
+        const identifierToAssign = createExportedIdentifier(
+            context,
+            lua.createIdentifier(symbol.escapedName.toString())
```
ark120202: Escaped string shouldn't be used as an unescaped one, there should be an unescaped one under `name` or `text` name

src/transformation/visitors/binary-expression/destructuring-assignments.ts
```diff
@@ -169,14 +174,23 @@ function transformShorthandPropertyAssignment(
...
         assignmentVariableName,
         lua.createTableIndexExpression(root, extractionIndex)
     );
 
-    result.push(variableExtractionAssignmentStatement);
+    const symbol = context.checker.getShorthandAssignmentValueSymbol(node);
+    const dependentSymbols = symbol ? getDependenciesOfSymbol(context, symbol) : [];
+
+    result.push(
+        createNestedImmediatelyInvokedFunctionExpressionsForEachSymbolSideEffect(
```
ark120202: Is there a reason why `transformAssignment` can't be used there?

test/unit/modules/modules.spec.ts
```diff
@@ -209,3 +209,56 @@ test("Export Equals", () => {
...
+        let a = 5;
+        a = 6;
+    `.expectToMatchJsResult();
+});
+
+test("special export specifier side effects", () => {
+    util.testModule`
+        let x = [0, 1, 2];
+        export { x as a };
+        [x.length] = [0];
```
ark120202: Is that test really required? There's no any special behavior there and it works just because `x` is a reference.

src/transformation/visitors/binary-expression/compound.ts
```diff
@@ -177,6 +185,11 @@ export function transformCompoundAssignmentStatement(
         // Simple statements
         // ${left} = ${left} ${replacementOperator} ${right}
         const operatorExpression = transformBinaryOperation(context, left, right, replacementOperator, node);
-        return transformAssignment(context, lhs, operatorExpression);
+        const assignmentStatements = transformAssignment(context, lhs, operatorExpression);
+        if (assignmentStatements.length === 1) {
+            return assignmentStatements[0];
+        } else {
+            return lua.createDoStatement(assignmentStatements, lhs);
```
ark120202: Can't it also be changed to `lua.Statement[]`

src/transformation/visitors/binary-expression/assignments.ts
```diff
@@ -1,24 +1,53 @@
...
+        ? createExportedIdentifier(context, left)
+        : cast(left, lua.isAssignmentLeftHandSideExpression);
+}
+
+export function createAssignmentStatementsForEachSymbolSideEffect(
+    context: TransformationContext,
+    rootAssignment: lua.AssignmentStatement,
+    symbols: ts.Symbol[]
+): lua.AssignmentStatement[] {
+    return symbols.reduce<lua.AssignmentStatement[]>(
```
ark120202:
```ts
    return [
        rootAssignment,
        ...symbols.map(symbol => {
            const [left] = rootAssignment.left;
            const identifierToAssign = createExportedIdentifier(context, lua.createIdentifier(symbol.name));
            return lua.createAssignmentStatement(identifierToAssign, left);
        })
    ];
```

Also, it doesn't seem to be used anywhere else, does it have to be exported? Or maybe it can be just inlined to `transformAssignment`?

src/transformation/utils/export.ts
```diff
@@ -95,12 +86,43 @@ export function isSymbolExportedFromScope(
...
+            symbol.declarations
+                .filter(ts.isExportSpecifier)
+                .map(context.checker.getExportSpecifierLocalTargetSymbol)
+                .some((symbol): symbol is ts.Symbol => (symbol ? symbol === originalSymbol : false))
+        ) {
+            relatedSymbols.push(symbol);
+        }
+
+        return relatedSymbols;
+    }, []);
```
ark120202: Should be equivalent to 

```ts
    return getExportedSymbolsFromScope(context, context.sourceFile).filter(exportSymbol =>
        exportSymbol.declarations
            .filter(ts.isExportSpecifier)
            .map(context.checker.getExportSpecifierLocalTargetSymbol)
            .includes(originalSymbol)
    );
```

PR #765, opened 2019-12-08
Use lua's math.atan2 for Math.atan2

### LINKED ISSUE #746: Transpiling Math.atan2 to math.atan2
Currently `Math.atan2(y, x)` is transpiled to `math.atan(y / x)`.

`math.atan2` is available in [Lua 5](https://www.lua.org/manual/5.1/manual.html#pdf-math.atan2).

> math.atan2 (y, x)
Returns the arc tangent of y/x (in radians), but uses the signs of both parameters to find the quadrant of the result. (It also handles correctly the case of x being zero.)

Therefore `Math.atan2(y, x)` can be transpiled to `math.atan2(y, x)` for the extra functionality of quadrant checking and handling the case of x being zero.

hazzard993:
Closes #746

`Math.atan2` -> `math.atan2` for all Lua targets

Edit: Except for Lua 5.3 which uses `math.atan(y, x)`


Perryvw:
In case the target is Lua 5.3, we should use `math.atan(y, x)` since atan2 is deprecated, see https://www.lua.org/manual/5.3/manual.html#8.2

test/unit/builtins/math.spec.ts
```diff
@@ -24,3 +24,11 @@ test.each(["E", "LN10", "LN2", "LOG10E", "LOG2E", "SQRT1_2", "SQRT2"])("Math.%s"
         expect(builder.getLuaExecutionResult()).toBeCloseTo(builder.getJsExecutionResult());
     });
 });
+
+test.each([LuaTarget.Lua51, LuaTarget.Lua52, LuaTarget.Lua53, LuaTarget.LuaJIT])("Math.atan2 (%p)", luaTarget => {
+    util.testExpression`
+        Math.atan2(4, 5);
+    `
```
ark120202:
```ts
util.testExpression`Math.atan2(4, 5)`
```

test/unit/builtins/math.spec.ts
```diff
@@ -24,3 +24,11 @@ test.each(["E", "LN10", "LN2", "LOG10E", "LOG2E", "SQRT1_2", "SQRT2"])("Math.%s"
...
+        .setOptions({ luaTarget })
+        .expectLuaToMatchSnapshot();
+});
```
ark120202: We actually have a `testEachVersion` function for such tests:

```ts
util.testEachVersion("Math.atan2", () => util.testExpression`Math.atan2(4, 5)`.expectLuaToMatchSnapshot());
```

src/transformation/builtins/math.ts
```diff
@@ -32,6 +33,13 @@ export function transformMathCall(context: TransformationContext, node: Property
     const params = transformArguments(context, node.arguments, signature);
 
     const expressionName = expression.name.text;
+
+    if (context.options.luaTarget === LuaTarget.Lua53 && expressionName === "atan2") {
```
Perryvw: I'd prefer if you handle all atan2 cases in the same place (probably in its switch case)

src/transformation/builtins/math.ts
```diff
@@ -74,7 +68,13 @@ export function transformMathCall(context: TransformationContext, node: Property
         case "acos":
         case "asin":
         case "atan":
-        case "atan2":
+        case "atan2": {
+            if (expressionName === "atan2" && context.options.luaTarget === LuaTarget.Lua53) {
```
Perryvw: `expressionName === "atan2"` is already checked by the switch, it's not required here

Perryvw: Oh I see why it's here now, it's still trange to me, you can move atan2 up to be the first case or something

test/unit/builtins/math.spec.ts
```diff
@@ -24,3 +24,13 @@ test.each(["E", "LN10", "LN2", "LOG10E", "LOG2E", "SQRT1_2", "SQRT2"])("Math.%s"
         expect(builder.getLuaExecutionResult()).toBeCloseTo(builder.getJsExecutionResult());
     });
 });
+
+const expectMathAtan2: util.TapCallback = builder => expect(builder.getMainLuaCodeChunk()).toMatch("math.atan2");
+const expectMathAtan: util.TapCallback = builder => expect(builder.getMainLuaCodeChunk()).toContain("math.atan");
```
ark120202: Is there a reason to use `toMatch` in the first check and `toContain` in the second?

Also, wouldn't `.toContain("math.atan")` match `atan2` as well? If so matched value should include `(`.

src/transformation/builtins/math.ts
```diff
@@ -33,12 +34,12 @@ export function transformMathCall(context: TransformationContext, node: Property
...
     const expressionName = expression.name.text;
     switch (expressionName) {
-        // math.tan(x / y)
+        // math.atan(y, x) Lua 5.3 only
         case "atan2": {
             const math = lua.createIdentifier("math");
-            const atan = lua.createStringLiteral("atan");
-            const div = lua.createBinaryExpression(params[0], params[1], lua.SyntaxKind.DivisionOperator);
-            return lua.createCallExpression(lua.createTableIndexExpression(math, atan), [div], node);
+            const functionChoice = context.options.luaTarget === LuaTarget.Lua53 ? "atan" : expressionName;
```
ark120202:
```suggestion
            const methodName = context.options.luaTarget === LuaTarget.Lua53 ? "atan" : "atan2";
```

src/transformation/builtins/math.ts
```diff
@@ -33,12 +34,12 @@ export function transformMathCall(context: TransformationContext, node: Property
 
     const expressionName = expression.name.text;
     switch (expressionName) {
-        // math.tan(x / y)
+        // math.atan(y, x) Lua 5.3 only
```
ark120202: "Lua 5.3 **only**" sounds like it would throw when it's not 5.3
```suggestion
        // math.atan2(y, x)
        // Lua 5.3: math.atan(y, x)
```

test/unit/builtins/math.spec.ts
```diff
@@ -24,3 +24,13 @@ test.each(["E", "LN10", "LN2", "LOG10E", "LOG2E", "SQRT1_2", "SQRT2"])("Math.%s"
...
+
+util.testEachVersion("Math.atan2", () => util.testExpression`Math.atan2(4, 5)`, {
```
ark120202: Maybe we can also have some result comparison tests (only for 5.3)?

src/transformation/builtins/math.ts
Perryvw:
```suggestion
        // Lua 5.3: math.atan(y, x)
        // Otherwise: math.atan2(y, x)
```

test/unit/builtins/math.spec.ts
```diff
@@ -24,3 +24,17 @@ test.each(["E", "LN10", "LN2", "LOG10E", "LOG2E", "SQRT1_2", "SQRT2"])("Math.%s"
...
+    [tstl.LuaTarget.LuaJIT]: builder => builder.tap(expectMathAtan2),
+    [tstl.LuaTarget.Lua51]: builder => builder.tap(expectMathAtan2),
+    [tstl.LuaTarget.Lua52]: builder => builder.tap(expectMathAtan2),
+    [tstl.LuaTarget.Lua53]: builder => builder.tap(expectMathAtan),
+});
+
+test("Lua 5.3 Math.atan2 comparison", () => {
+    util.testExpression`Math.atan2(4, 5)`.setOptions({ luaTarget: tstl.LuaTarget.Lua53 }).expectToMatchJsResult();
```
ark120202: 5.3 is a default target

```suggestion
    util.testExpression`Math.atan2(4, 5)`.expectToMatchJsResult();
```

test/unit/builtins/math.spec.ts
```diff
@@ -24,3 +24,17 @@ test.each(["E", "LN10", "LN2", "LOG10E", "LOG2E", "SQRT1_2", "SQRT2"])("Math.%s"
...
+const expectMathAtan: util.TapCallback = builder => expect(builder.getMainLuaCodeChunk()).toContain("math.atan(");
+
+util.testEachVersion("Math.atan2", () => util.testExpression`Math.atan2(4, 5)`, {
+    [tstl.LuaTarget.LuaJIT]: builder => builder.tap(expectMathAtan2),
+    [tstl.LuaTarget.Lua51]: builder => builder.tap(expectMathAtan2),
+    [tstl.LuaTarget.Lua52]: builder => builder.tap(expectMathAtan2),
+    [tstl.LuaTarget.Lua53]: builder => builder.tap(expectMathAtan),
+});
+
+test("Lua 5.3 Math.atan2 comparison", () => {
```
ark120202: `comparison` is generally a default testing method and 5.3 is the only execution target we support anyway, so I'd rather call it consistently with all other comparison tests

```suggestion
test("Math.atan2(4, 5)", () => {
```

PR #766, opened 2019-12-08
Transpile console.error/info/warn to regular print like console.log

Perryvw:


src/transformation/builtins/console.ts
```diff
@@ -16,6 +16,7 @@ export function transformConsoleCall(
     const parameters = transformArguments(context, expression.arguments, signature);
 
     switch (methodName) {
+        case "error":
```
lolleko: Maybe also add the remaining console API functions like log, warn ...

https://console.spec.whatwg.org/#console-namespace

Perryvw: Good call, added .info and .warn too

PR #767, opened 2019-12-08
Added string.trim, string.trim{Left,Right,StartEnd}

Perryvw:
Also string.trimLeft, string.trimRight, string.trimStart and string.trimEnd


test/unit/builtins/string.spec.ts
```diff
@@ -265,3 +265,21 @@ test.each([
         `;
     expect(util.transpileAndExecute(code)).toBe(6);
 });
+
+const trimTestCases = [
```
ark120202: I haven't checked the spec, but from a quick look at MDN page it also has to handle `\r\n\t\f\v\uFEFF\xA0`

Perryvw: right, I'll add some more testcases

src/lualib/StringTrimStart.ts
```diff
@@ -0,0 +1,5 @@
+function __TS__StringTrimStart(this: string): string {
+    // http://lua-users.org/wiki/StringTrim
+    const [result] = string.gsub(this, "^[%s\xA0\u{FEFF}]*", "");
```
ark120202:
```suggestion
    const [result] = string.gsub(this, "^[%s\xA0\uFEFF]*", "");
```

test/unit/builtins/string.spec.ts
```diff
@@ -265,3 +265,28 @@ test.each([
...
+    "",
+    " ",
+    "   ",
+    "foo ",
+    "foo ",
```
ark120202: Are these 2 strings same?

Perryvw: One is a tab and one is a space, but alignment makes them look the same.

test/unit/builtins/string.spec.ts
```diff
@@ -265,3 +265,28 @@ test.each([
...
+    "    foo",
+    "   foo     ",
+    "    foo     ",
+    "  foo    bar     ",
+    "\r\nfoo\n\r\n",
+    "\r\nfoo\nbar\n\r\n",
+];
+describe.each(["trim", "trimEnd", "trimRight", "trimStart", "trimLeft"])("string.%s", trim => {
+    test.each(trimTestCases)("matches JS result (%p)", testString => {
+        util.testExpression`\`${testString}\`.${trim}()`.expectToMatchJsResult();
```
ark120202: Would it work with

```suggestion
        util.testExpression`${util.formatCode(testString)}.${trim}()`.expectToMatchJsResult();
```

PR #773, opened 2019-12-18
Load plugins from `tsconfig.json`

ark120202:


test/transpile/plugins/visitor-super.ts
```diff
@@ -0,0 +1,19 @@
+import * as ts from "typescript";
+import * as tstl from "../../../src";
+
+const plugin: tstl.Plugin = {
+    visitors: {
+        [ts.SyntaxKind.StringLiteral]: (node, context) => {
```
Perryvw: Could you add a comment just describing what these (also the other two) plugins do, something like `// Visit string literals, if original transformer returns a string literal, change the string to "bar" instead`? That way it is also easier to link to these as example for our users/documentation.

test/transpile/resolve-plugin/cjs.js
```diff
@@ -0,0 +1 @@
+module.exports = true;
```
Perryvw: Wondering if we should somehow separate these files from the test files, maybe with a data/ directory? What do you think?

PR #774, opened 2019-12-19
Generalize class accessors transform

### LINKED ISSUE #576: Class properties should override parent class accessors

### LINKED ISSUE #725: Inherited accessors break when base is type-erased
When extending type-erased bases with accessors, the accessors are called with incorrect `this`. 

## Example

```typescript
const A = class {
    get value(): this {
        return this;
    }
} as any;

class B extends A {
    
}

const b = new B();
console.assert(b.value === b); // boom
```

ark120202:
Fixes #725


test/unit/printer/sourcemaps.spec.ts
```diff
@@ -86,9 +86,8 @@ test.each([
         assertPatterns: [
             { luaPattern: "Bar = __TS__Class()", typeScriptPattern: "class Bar" },
             { luaPattern: "Bar.name =", typeScriptPattern: "class Bar" },
-            { luaPattern: "Bar.____super = Foo", typeScriptPattern: "Foo {" },
-            { luaPattern: "setmetatable(Bar,", typeScriptPattern: "Foo {" },
-            { luaPattern: "setmetatable(Bar.prototype,", typeScriptPattern: "Foo {" },
+            { luaPattern: "__TS__ClassExtends(", typeScriptPattern: "extends" },
+            { luaPattern: "Foo)", typeScriptPattern: "Foo {" },
```
Perryvw: This last pattern seems a bit strange to test. Could we instead test some other lua pattern we actually care about, otherwise just delete this. I doubt lua will ever throw an error on `Foo)`'s location anyway.

src/lualib/declarations/tstl.d.ts
```diff
@@ -12,11 +12,13 @@ interface LuaClass {
     ____getters?: { [key: string]: (self: LuaClass) => any };
     ____setters?: { [key: string]: (self: LuaClass, val: any) => void };
     __index?: any;
+    __newindex?: any;
 }
 
 interface LuaObject {
     constructor: LuaClass;
     ____getters?: { [key: string]: (self: LuaObject) => any };
```
Perryvw: Do we still need these _____getters and ____setters?

src/lualib/Descriptors.ts
```diff
@@ -0,0 +1,72 @@
...
+}
+
+function __TS__SetDescriptor(this: void, metatable: any, prop: string, descriptor: PropertyDescriptor): void {
+    if (descriptor.get) metatable.__index = ____descriptorIndex;
+    if (descriptor.set) metatable.__newindex = ____descriptorNewindex;
+    if (!metatable._descriptors) metatable._descriptors = {};
+    metatable._descriptors[prop] = descriptor;
+}
+
+function __TS__ObjectDefineProperty<T extends object>(
```
Perryvw: Some comments in this file might be good, starting with a link to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty for this function (including how this function differs from that in functionality)

test/unit/printer/sourcemaps.spec.ts
ark120202: I've changed it to a function call, which should be a more important mapping

Perryvw:
Can you add a 'Next release' section the changelog and add a point describing this PR?

test/unit/printer/sourcemaps.spec.ts
```diff
@@ -76,7 +76,7 @@ test.each([
     {
         code: `
             // @ts-ignore
-            class Bar extends Foo {
+            class Bar extends Foo() {
```
Perryvw: Why this change?

ark120202: https://github.com/TypeScriptToLua/TypeScriptToLua/pull/774#discussion_r368480625
As a function call it gets printed on a new line, which makes it important for mapping stack traces

Perryvw: I would prefer if we continue testing the most commonly used class pattern (like it was before). If you want to test this because of the new line, please just add another test.

ark120202: I'm fine with reverting it to just `Foo`, but I don't really want to add a new case just for a single mapping. So maybe just re-add `{ luaPattern: "Foo)", typeScriptPattern: "Foo {" }`? It is the thing we actually care about, even though it wouldn't be in the call stack, it's still useful for other sourcemap use cases

Perryvw: sure

PR #775, opened 2019-12-19
Share scope between all switch cases

### LINKED ISSUE #698: Switch cases should share scope
```ts
switch (null!) {
    case 1:
        let foo: number | undefined = 1;
    case 2:
        foo = 2;
    case 3:
        if (foo !== undefined) {
            console.log(foo + 1);
        }
}
```

Cases probably shouldn't be wrapped in `do` blocks?

ark120202:
Fixes #698.

### DIFF
--- a/test/unit/conditionals.spec.ts
+++ b/test/unit/conditionals.spec.ts
@@ -157,7 +157,32 @@ test.each([0, 1, 2, 3])("nestedSwitch (%p)", inp => {
     `.expectToMatchJsResult();
 });
 
-test.each([0, 1, 2])("switchLocalScope (%p)", inp => {
+test("switch cases scope", () => {
+    util.testFunction`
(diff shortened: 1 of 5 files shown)


test/unit/conditionals.spec.ts
```diff
@@ -157,7 +157,32 @@ test.each([0, 1, 2, 3])("nestedSwitch (%p)", inp => {
...
+    util.testFunction`
+        let foo: number = 0;
+        switch (foo) {
+            case 0:
+                let foo = true;
+            case 1:
+                return foo;
+        }
+    `.expectToMatchJsResult();
+});
```
Perryvw: What about adding a test case for 
```ts
let foo: number = 0;
switch (foo) {
    case 0: {
        let foo = true;
    }
    case 1:
        return foo;
}
```

PR #776, opened 2019-12-19
Assume files with imports to be modules

### LINKED ISSUE #564: Consider all files as modules
TSTL target environment is closer to node/bundler, where each module is isolated even if it doesn't contain any imports/exports. Even though in TypeScript the following code declares a global `onEvent` symbol (mostly because of legacy environments support) in runtime you usually won't expect `_G.onEvent` to be defined.

```ts
function onEvent() {}
listenToEvent(onEvent);
```

ark120202:
Closes #564.

Now it matches TypeScript's logic, making files with `imports` or `exports` modules. It still keeps files without any imports or exports global, so probably shouldn't break too much things, because TypeScript already issued an unused variable diagnostic in these cases.

### DIFF
--- a/src/transformation/context/context.ts
+++ b/src/transformation/context/context.ts
@@ -2,7 +2,6 @@ import * as ts from "typescript";
 import { CompilerOptions, LuaTarget } from "../../CompilerOptions";
 import * as lua from "../../LuaAST";
 import { unwrapVisitorResult } from "../utils/lua-ast";
-import { isFileModule } from "../utils/typescript";
 import { ExpressionLikeNode, ObjectVisitor, StatementLikeNode, VisitorMap } from "./visitors";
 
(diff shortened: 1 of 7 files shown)


src/transformation/context/context.ts
```diff
@@ -23,7 +22,7 @@ export class TransformationContext {
 
     public readonly options: CompilerOptions = this.program.getCompilerOptions();
     public readonly luaTarget = this.options.luaTarget ?? LuaTarget.LuaJIT;
-    public readonly isModule = isFileModule(this.sourceFile);
+    public readonly isModule = ts.isExternalModule(this.sourceFile);
```
lolleko: What if all imports are elided is a file still considered a module in that case?

ark120202: Yes

PR #783, opened 2019-12-30
Bugfix/loop destructuring

### LINKED ISSUE #770: Complex destructuring patterns don't work in loops
[Playground](https://typescripttolua.github.io/play.html#src=for%20(const%20%5B%5Bx%5D%5D%20of%20%5B%5B%5B1%5D%5D%5D)%20%7B%7D)

Currently value is just unpacked - https://github.com/TypeScriptToLua/TypeScriptToLua/blob/master/src/transformation/visitors/loops/for-of.ts#L33, but it should use the same logic as variable declaration does. For the same reason object destructuring is not supported currently.

### LINKED ISSUE #779: Parameter destructuring leaks variables
```ts
function func({a, b, c}) {}
```
results into 
```lua
    function(____bindingPattern0)
        a = ____bindingPattern0.a
        b = ____bindingPattern0.b
        c = ____bindingPattern0.c
```

**whereas**

```ts
function func(params) {
  const {a, b, c} = params;
}
```
results into
```lua
    function(params)
        local a = params.a
        local b = params.b
        local c = params.c
```

`local` should be there for parameter destructuring as well

Perryvw:
Fixed some issues regarding variable destructuring.

Fixes #770 
Fixes #779

### DIFF
--- a/src/transformation/utils/scope.ts
+++ b/src/transformation/utils/scope.ts
@@ -77,12 +77,14 @@ export function findScope(context: TransformationContext, scopeTypes: ScopeType)
 }
 
 const scopeIdCounters = new WeakMap<TransformationContext, number>();
-export function pushScope(context: TransformationContext, scopeType: ScopeType): void {
+export function pushScope(context: TransformationContext, scopeType: ScopeType): Scope {
     const nextScopeId = (scopeIdCounters.get(context) ?? 0) + 1;
     scopeIdCounters.set(context, nextScopeId);
 
     const scopeStack = getScopeStack(context);
-    scopeStack.push({ type: scopeType, id: nextScopeId });
+    const scope = { type: scopeType, id: nextScopeId };
+    scopeStack.push(scope);
+    return scope;
 }
 
 export function popScope(context: TransformationContext): Scope {
--- a/src/transformation/utils/scope.ts
+++ b/src/transformation/utils/scope.ts
@@ -82,7 +82,7 @@ export function pushScope(context: TransformationContext, scopeType: ScopeType):
     scopeIdCounters.set(context, nextScopeId);
 
     const scopeStack = getScopeStack(context);
-    const scope = { type: scopeType, id: nextScopeId };
+    const scope: Scope = { type: scopeType, id: nextScopeId };
     scopeStack.push(scope);
     return scope;
 }
--- a/src/transformation/visitors/variable-declaration.ts
+++ b/src/transformation/visitors/variable-declaration.ts
@@ -29,7 +29,7 @@ export function transformArrayBindingElement(
 export function transformBindingPattern(
     context: TransformationContext,
     pattern: ts.BindingPattern,
-    table: lua.Identifier,
+    table: lua.Expression,
     propertyAccessStack: ts.PropertyName[] = []
 ): lua.Statement[] {
     const result: lua.Statement[] = [];
--- a/src/transformation/visitors/variable-declaration.ts
+++ b/src/transformation/visitors/variable-declaration.ts
@@ -26,12 +26,16 @@ export function transformArrayBindingElement(
     }
 }
 
-export function transformBindingPattern(
+export function createDestructuringStatements(
     context: TransformationContext,
     pattern: ts.BindingPattern,
     table: lua.Expression,
     propertyAccessStack: ts.PropertyName[] = []
 ): lua.Statement[] {
+    if (pattern.elements.length === 0) {
+        return [];
+    }
+
     const result: lua.Statement[] = [];
     const isObjectBindingPattern = ts.isObjectBindingPattern(pattern);
 
@@ -48,7 +52,7 @@ export function transformBindingPattern(
                 propertyAccessStack.push(propertyName);
             }
 
-            result.push(...transformBindingPattern(context, element.name, table, propertyAccessStack));
+            result.push(...createDestructuringStatements(context, element.name, table, propertyAccessStack));
             continue;
         }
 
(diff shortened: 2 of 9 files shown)


src/transformation/utils/scope.ts
```diff
@@ -77,12 +77,14 @@ export function findScope(context: TransformationContext, scopeTypes: ScopeType)
...
 
 const scopeIdCounters = new WeakMap<TransformationContext, number>();
-export function pushScope(context: TransformationContext, scopeType: ScopeType): void {
+export function pushScope(context: TransformationContext, scopeType: ScopeType): Scope {
     const nextScopeId = (scopeIdCounters.get(context) ?? 0) + 1;
     scopeIdCounters.set(context, nextScopeId);
 
     const scopeStack = getScopeStack(context);
-    scopeStack.push({ type: scopeType, id: nextScopeId });
+    const scope = { type: scopeType, id: nextScopeId };
```
ark120202:
```suggestion
    const scope: Scope = { type: scopeType, id: nextScopeId };
```

src/transformation/utils/scope.ts
```diff
@@ -77,12 +77,14 @@ export function findScope(context: TransformationContext, scopeTypes: ScopeType)
 }
 
 const scopeIdCounters = new WeakMap<TransformationContext, number>();
-export function pushScope(context: TransformationContext, scopeType: ScopeType): void {
+export function pushScope(context: TransformationContext, scopeType: ScopeType): Scope {
```
ark120202: There should a bunch of usages of `peekScope` right after `pushScope`, would be nice to simplify them now

test/unit/destructuring.spec.ts
```diff
@@ -41,6 +41,19 @@ test.each([
...
+        expect(builder.getMainLuaCodeChunk()).toContain(lua);
+
+    util.testModule`
+        function test({a, b}) {}
+        test({ a: 4, b: 5});
+        `.tap(builder => {
+        expectLuaContains("local a =", builder);
+        expectLuaContains("local b =", builder);
+    });
+});
```
ark120202:
```ts
test("creates local variables for destructured function parameters", () => {
        function test({ a, b }: any) {}
    `;

    const code = builder.getMainLuaCodeChunk();
    expect(code).toContain("local a =");
    expect(code).toContain("local b =");
});
```

test/unit/destructuring.spec.ts
```diff
@@ -78,6 +91,16 @@ test.each(assignmentTestCases)("in assignment expression (%p)", ({ binding, valu
...
 
+test("destructuring in loop", () => {
+    util.testFunction`
+        let result = 0;
+        for (const [[x]] of [[[1]]]) {
+            result = x;
+        }
+        return result;
+        `.expectToMatchJsResult();
+});
```
ark120202: We already have some destructuring in `for...of` loop tests in `loops.spec.ts`.
Alternatively we might just run all of test cases there (filtering out currently unsupported things):

```ts
test.each(testCases)("in for...of array (%p)", ({ binding, value }) => {
    util.testFunction`
        let ${allBindings};
        {
            for (const ${binding} of [${value}]) {
                return { ${allBindings} };
            }
        }
    `.expectToMatchJsResult();
});
```

src/transformation/utils/scope.ts
Perryvw: Only found one in switch

src/transformation/visitors/variable-declaration.ts
```diff
@@ -29,7 +29,7 @@ export function transformArrayBindingElement(
 export function transformBindingPattern(
     context: TransformationContext,
     pattern: ts.BindingPattern,
-    table: lua.Identifier,
+    table: lua.Expression,
```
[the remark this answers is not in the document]
Perryvw: Seemed like no reason it was an identifier, also it makes sense this is an expression because you can also more complex expressions representing some destructible object.

test/unit/destructuring.spec.ts
Perryvw: moved

src/transformation/visitors/variable-declaration.ts
```diff
@@ -120,6 +124,93 @@ export function transformBindingPattern(
     return result;
 }
 
+export function transformBindingPattern(
```
ark120202: It transforms more than just binding pattern. Maybe `transform(Binding|Destructuring)VariableDeclaration`?

src/transformation/visitors/variable-declaration.ts
```diff
@@ -26,12 +26,16 @@ export function transformArrayBindingElement(
     }
 }
 
-export function transformBindingPattern(
+export function createDestructuringStatements(
```
ark120202: `pattern: ts.BindingPattern` is being transformed inside, so I think it should start with `transform`

test/unit/loops.spec.ts
```diff
@@ -7,174 +7,166 @@ import {
...
+test("for with expression", () => {
+    const inp = [0, 1, 2, 3];
+    util.testFunction`
+        let arrTest = ${util.formatCode(inp)};
+        let i: number;
+        for (i = 0 * 1; i < arrTest.length; ++i) {
+            arrTest[i] = arrTest[i] + 1;
+        }
+        return arrTest;
+        `.expectToMatchJsResult();
```
ark120202:
```suggestion
    `.expectToMatchJsResult();
```

Some other cases have the same thing

test/unit/loops.spec.ts
```diff
@@ -7,174 +7,166 @@ import {
...
-            for (i = 0 * 1; i < arrTest.length; ++i) {
-                arrTest[i] = arrTest[i] + 1;
-            }
-            return JSONStringify(arrTest);`
-    );
-    expect(result).toBe(JSON.stringify(expected));
+test("for with expression", () => {
+    const inp = [0, 1, 2, 3];
+    util.testFunction`
+        let arrTest = ${util.formatCode(inp)};
```
ark120202:
```suggestion
        let arrTest = [0, 1, 2, 3];
```

test/unit/loops.spec.ts
```diff
@@ -386,18 +368,17 @@ test("forof with iterator", () => {
...
-    `;
-    const compilerOptions = {
-        luaLibImport: tstl.LuaLibImportKind.Require,
-        luaTarget: tstl.LuaTarget.Lua53,
-        target: ts.ScriptTarget.ES2015,
-    };
-    const result = util.transpileAndExecute(code, compilerOptions);
-    expect(result).toBe("abc");
+    `
+        .setOptions({
```
ark120202: All of these options should be enabled by default now

PR #786, opened 2020-01-02
Fixed incorrect error on intersected number and string types

### LINKED ISSUE #782: Methods on types intersected with `number` cannot be used
Input:
```ts
type Vector = number & { normalize(): Vector };
(null! as Vector).normalize();
```

Current Result:
```
TSTL: Unsupported property on number: normalize
```

Expected Result:
No errors

Perryvw:
Fixes #782

### DIFF
--- a/src/LuaPrinter.ts
+++ b/src/LuaPrinter.ts
@@ -742,11 +742,15 @@ export class LuaPrinter {
     public printMethodCallExpression(expression: lua.MethodCallExpression): SourceNode {
         const chunks = [];
 
-        const prefix = this.printExpression(expression.prefixExpression);
-
         const name = this.printIdentifier(expression.name);
 
-        chunks.push(prefix, ":", name, "(");
+        const prefix = this.printExpression(expression.prefixExpression);
+
+        if (lua.isStringLiteral(expression.prefixExpression)) {
+            chunks.push("(", prefix, ")", ":", name, "(");
+        } else {
+            chunks.push(prefix, ":", name, "(");
+        }
 
         if (expression.params) {
             chunks.push(...this.printExpressionList(expression.params));
--- a/src/LuaPrinter.ts
+++ b/src/LuaPrinter.ts
@@ -742,15 +742,13 @@ export class LuaPrinter {
     public printMethodCallExpression(expression: lua.MethodCallExpression): SourceNode {
         const chunks = [];
 
+        const prefix = lua.isStringLiteral(expression.prefixExpression)
+            ? this.printExpression(lua.createParenthesizedExpression(expression.prefixExpression))
+            : this.printExpression(expression.prefixExpression);
+
         const name = this.printIdentifier(expression.name);
 
-        const prefix = this.printExpression(expression.prefixExpression);
-
-        if (lua.isStringLiteral(expression.prefixExpression)) {
-            chunks.push("(", prefix, ")", ":", name, "(");
-        } else {
-            chunks.push(prefix, ":", name, "(");
-        }
+        chunks.push(prefix, ":", name, "(");
 
         if (expression.params) {
             chunks.push(...this.printExpressionList(expression.params));
--- a/src/transformation/builtins/array.ts
+++ b/src/transformation/builtins/array.ts
@@ -8,7 +8,7 @@ import { PropertyCallExpression, transformArguments } from "../visitors/call";
 export function transformArrayPrototypeCall(
     context: TransformationContext,
     node: PropertyCallExpression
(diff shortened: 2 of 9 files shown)


test/unit/builtins/string.spec.ts
```diff
@@ -1,22 +1,7 @@
...
-    const tsHeader = `
-        declare global {
-            interface String {
-                upper(): string;
-            }
-        }
-    `;
-
-    util.testExpression`"test".upper()`.setTsHeader(tsHeader).expectToEqual("TEST");
+    util.testExpression`"test".toUpperCase()`.expectToEqual("TEST");
```
ark120202: Is that no longer supported?

test/unit/builtins/numbers.spec.ts
```diff
@@ -67,3 +67,10 @@ test.each(cases)("isNaN(%p)", value => {
 test.each(cases)("isFinite(%p)", value => {
     util.testExpressionTemplate`isFinite(${value} as any)`.expectToMatchJsResult();
 });
+
+test("number intersected method", () => {
+    util.testFunction`
+        type Vector = number & { normalize(): Vector };
+        return ({ normalize: () => 3 } as Vector).normalize();
+        `.expectToMatchJsResult();
```
ark120202:
```suggestion
    `.expectToMatchJsResult();
```

test/unit/builtins/string.spec.ts
```diff
@@ -287,3 +272,10 @@ describe.each(["trim", "trimEnd", "trimRight", "trimStart", "trimLeft"])("string
         util.testExpression`${util.formatCode(testString)}.${trim}()`.expectToMatchJsResult();
     });
 });
+
+test("string intersected method", () => {
+    util.testFunction`
+        type Vector = string & { abc(): Vector };
+        return ({ abc: () => "a" } as Vector).abc();
+        `.expectToMatchJsResult();
```
ark120202:
```suggestion
    `.expectToMatchJsResult();
```

(there are chances prettier would support it soon)

src/transformation/builtins/index.ts
```diff
@@ -85,11 +92,11 @@ export function transformBuiltinCallExpression(
         }
     }
 
-    if (isStringType(context, ownerType)) {
+    if (isStringType(context, ownerType) && hasStandardLibrarySignature(context, node)) {
         return transformStringPrototypeCall(context, node);
     }
 
-    if (isNumberType(context, ownerType)) {
+    if (isNumberType(context, ownerType) && hasStandardLibrarySignature(context, node)) {
```
ark120202: Could we use it for array (and maybe other types) as well?

src/LuaPrinter.ts
```diff
@@ -742,11 +742,15 @@ export class LuaPrinter {
     public printMethodCallExpression(expression: lua.MethodCallExpression): SourceNode {
         const chunks = [];
 
-        const prefix = this.printExpression(expression.prefixExpression);
-
         const name = this.printIdentifier(expression.name);
 
-        chunks.push(prefix, ":", name, "(");
+        const prefix = this.printExpression(expression.prefixExpression);
```
ark120202: How about something like that to remove parenthesizing duplication? 
```suggestion
        const prefix = this.printExpression(lua.isStringLiteral(expression.prefixExpression) ? expression.prefixExpression : lua.createParenthesizedExpression(expression.prefixExpression));;
```

Also, what about `"foo".bar`?

src/transformation/builtins/array.ts
```diff
@@ -3,15 +3,13 @@ import * as lua from "../../LuaAST";
 import { TransformationContext } from "../context";
 import { UnsupportedProperty } from "../utils/errors";
 import { LuaLibFeature, transformLuaLibFunction } from "../utils/lualib";
-import { isExplicitArrayType } from "../utils/typescript";
 import { PropertyCallExpression, transformArguments } from "../visitors/call";
 
 export function transformArrayPrototypeCall(
     context: TransformationContext,
     node: PropertyCallExpression
 ): lua.CallExpression | undefined {
```
ark120202:
```suggestion
): lua.CallExpression {
```

src/LuaPrinter.ts
Perryvw: Should not be affected by this PR, I'd prefer to keep that for the full parenthesis to printer change.

src/transformation/builtins/array.ts
Perryvw: That's not possible because the unsupported error is only thrown when isExplicitArrayType is true

PR #791, opened 2020-01-08
Switch case goto chain is now transpiled to an elseif chain

DoctorGester:
elseif instead of multiple if
Fixed 82 tests being skipped in conditionals.spec.ts

### DIFF
--- a/src/transformation/visitors/switch.ts
+++ b/src/transformation/visitors/switch.ts
@@ -19,9 +19,9 @@ export const transformSwitchStatement: FunctionVisitor<ts.SwitchStatement> = (st
     let statements: lua.Statement[] = [];
 
     const caseClauses = statement.caseBlock.clauses.filter(ts.isCaseClause);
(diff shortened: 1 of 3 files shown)


src/transformation/visitors/switch.ts
```diff
@@ -29,7 +31,20 @@ export const transformSwitchStatement: FunctionVisitor<ts.SwitchStatement> = (st
 
         const goto = lua.createGotoStatement(`${switchName}_case_${index}`);
         const conditionalGoto = lua.createIfStatement(condition, lua.createBlock([goto]));
-        statements.push(conditionalGoto);
+
+        resultIfStatements.push(conditionalGoto);
+    }
+
+    if (resultIfStatements.length > 0) {
```
Perryvw: Please use `statements.reduce((previousIf, currentIf) => (previousIf.elseBlock = currentIf, currentIf));` instead if this entire if statement if it works

DoctorGester: Gotem

PR #795, opened 2020-01-09
Fixed inconsistent scoping of top-level variable declarations

### LINKED ISSUE #720: Inconsistentcy in top-level variable handling in non-module files
[Playground](https://typescripttolua.github.io/play.html#src=const%20obj%20%3D%20%7B%20value1%3A%201%2C%20value2%3A%202%20%7D%3B%0Aconst%20value1%20%3D%20obj.value1%3B%0Aconst%20%7B%20value2%20%7D%20%3D%20obj%3B%0A%0Afunction%20fun1()%20%7B%7D%0Aconst%20fun2%20%3D%20()%20%3D>%20%7B%7D%3B%0A)

Input:

```ts
const obj = { value1: 1, value2: 2 };
const value1 = obj.value1;
const { value2 } = obj;

function fun1() {}
const fun2 = () => {};
```

Current Result:

```lua
local obj = {value1 = 1, value2 = 2}
local value1 = obj.value1
value2 = obj.value2
function fun1(self)
end
local fun2
fun2 = function()
end
```

Expected Result:

Either never use `local` for top-level variables in non-modules or always use `local` (ref #564).

### LINKED ISSUE #790: Destructuring leaks in blocks (do/end in Lua)
```ts
{
	const a = 1; // always works
	const [b] = [1]; // works on playground, breaks latest local version
	const { c } = { c: 1 }; // never works
}
```

```lua
-- Generated with latest TypeScriptToLua
do
    a = 1
    local b = 1
    local ____ = {c = 1}
    c = ____.c
end
```

https://typescripttolua.github.io/play#src=%7B%0A%09const%20a%20%3D%201%3B%20%2F%2F%20always%20works%0A%09const%20%5Bb%5D%20%3D%20%5B1%5D%3B%20%2F%2F%20works%20on%20playground%2C%20breaks%20latest%20local%20version%0A%09const%20%7B%20c%20%7D%20%3D%20%7B%20c%3A%201%20%7D%3B%20%2F%2F%20never%20works%0A%7D%0A

### LINKED ISSUE #792: Instantiation of userdata types declared as classes (0.30)
Hi,
after upgrading to 0.30, the instantiation of userdata types declared as classes does not work anymore.
Basically, I have bound a C userdata type `UserDataClass` and a corresponding factory function to Lua so that calling `UserDataClass.new()` would produce a userdata instance on which I would then call `instance:classMethod()` in Lua.
I have declared these types by simply writing
```typescript
declare class UserDataClass {
   classMethod();
}
```
and it would "just work" (which has been quite awesome).
How can I re-enable this functionality in 0.30?
Could this be considered a bug? 
I use this stuff a lot and it basically prevents me from upgrading right now.
Thank you!

Perryvw:
We have the convention of treating files without imports or exports as 'global' scripts where all top-level variables are considered global.

Variable declarations however did not adhere to this convention, that has been corrected.

Fixes #720, closes #790

### DIFF
--- a/src/transformation/utils/lua-ast.ts
+++ b/src/transformation/utils/lua-ast.ts
@@ -3,6 +3,7 @@ import * as ts from "typescript";
 import { LuaTarget } from "../../CompilerOptions";
 import * as lua from "../../LuaAST";
 import { TransformationContext } from "../context";
+import { getCurrentNamespace } from "../visitors/namespace";
 import { createExportedIdentifier, getIdentifierExportScope } from "./export";
 import { peekScope, ScopeType } from "./scope";
 import { isFunctionType } from "./typescript";
@@ -141,14 +142,10 @@ export function createLocalOrExportedOrGlobalDeclaration(
             );
         }
     } else {
-        const isTopLevelVariable = peekScope(context).type === ScopeType.File;
+        const scope = peekScope(context);
+        const isTopLevelVariable = scope.type === ScopeType.File;
 
-        if (rhs && isTopLevelVariable && !context.isModule) {
-            // top-level declarations in non-module files are considered global
-            assignment = lua.createAssignmentStatement(lhs, rhs, tsOriginal);
-        } else {
-            const scope = peekScope(context);
-
+        if (context.isModule || getCurrentNamespace(context) || !isTopLevelVariable) {
             const isPossibleWrappedFunction =
                 !isFunctionDeclaration &&
                 tsOriginal &&
@@ -176,6 +173,11 @@ export function createLocalOrExportedOrGlobalDeclaration(
                     declaration = undefined;
                 }
             }
+        } else if (rhs) {
+            // global
+            assignment = lua.createAssignmentStatement(lhs, rhs, tsOriginal);
+        } else {
+            return [];
         }
     }
 
--- a/src/transformation/utils/lua-ast.ts
+++ b/src/transformation/utils/lua-ast.ts
@@ -122,7 +122,6 @@ export function createLocalOrExportedOrGlobalDeclaration(
     let declaration: lua.VariableDeclarationStatement | undefined;
     let assignment: lua.AssignmentStatement | undefined;
 
-    const isVariableDeclaration = tsOriginal !== undefined && ts.isVariableDeclaration(tsOriginal);
     const isFunctionDeclaration = tsOriginal !== undefined && ts.isFunctionDeclaration(tsOriginal);
 
     const identifiers = Array.isArray(lhs) ? lhs : [lhs];
@@ -145,7 +144,7 @@ export function createLocalOrExportedOrGlobalDeclaration(
     } else {
         const insideFunction = findScope(context, ScopeType.Function) !== undefined;
 
-        if (context.isModule || getCurrentNamespace(context) || insideFunction || isVariableDeclaration) {
+        if (context.isModule || getCurrentNamespace(context) || insideFunction) {
             const scope = peekScope(context);
 
             const isPossibleWrappedFunction =
--- a/src/transformation/utils/lua-ast.ts
+++ b/src/transformation/utils/lua-ast.ts
@@ -143,8 +143,9 @@ export function createLocalOrExportedOrGlobalDeclaration(
         }
     } else {
         const insideFunction = findScope(context, ScopeType.Function) !== undefined;
+        const insideBlock = findScope(context, ScopeType.Block) !== undefined;
 
-        if (context.isModule || getCurrentNamespace(context) || insideFunction) {
+        if (context.isModule || getCurrentNamespace(context) || insideFunction || insideBlock) {
             const scope = peekScope(context);
 
             const isPossibleWrappedFunction =
--- a/src/transformation/utils/lua-ast.ts
+++ b/src/transformation/utils/lua-ast.ts
@@ -3,7 +3,6 @@ import * as ts from "typescript";
 import { LuaTarget } from "../../CompilerOptions";
 import * as lua from "../../LuaAST";
 import { TransformationContext } from "../context";
-import { getCurrentNamespace } from "../visitors/namespace";
 import { createExportedIdentifier, getIdentifierExportScope } from "./export";
 import { peekScope, ScopeType } from "./scope";
 import { isFunctionType } from "./typescript";
@@ -145,7 +144,7 @@ export function createLocalOrExportedOrGlobalDeclaration(
         const scope = peekScope(context);
         const isTopLevelVariable = scope.type === ScopeType.File;
 
-        if (context.isModule || getCurrentNamespace(context) || !isTopLevelVariable) {
+        if (context.isModule || !isTopLevelVariable) {
             const isPossibleWrappedFunction =
                 !isFunctionDeclaration &&
                 tsOriginal &&
--- a/src/transformation/utils/lua-ast.ts
+++ b/src/transformation/utils/lua-ast.ts
@@ -1,11 +1,10 @@
-import * as ts from "typescript";
 import * as assert from "assert";
+import * as ts from "typescript";
 import { LuaTarget } from "../../CompilerOptions";
 import * as lua from "../../LuaAST";
 import { TransformationContext } from "../context";
-import { getCurrentNamespace } from "../visitors/namespace";
 import { createExportedIdentifier, getIdentifierExportScope } from "./export";
-import { findScope, peekScope, ScopeType } from "./scope";
+import { peekScope, ScopeType } from "./scope";
 import { isFunctionType } from "./typescript";
 
 export type OneToManyVisitorResult<T extends lua.Node> = T | T[] | undefined;
@@ -142,10 +141,12 @@ export function createLocalOrExportedOrGlobalDeclaration(
             );
         }
     } else {
-        const insideFunction = findScope(context, ScopeType.Function) !== undefined;
-        const insideBlock = findScope(context, ScopeType.Block) !== undefined;
+        const isTopLevelVariable = peekScope(context).type === ScopeType.File;
 
-        if (context.isModule || getCurrentNamespace(context) || insideFunction || insideBlock) {
+        if (rhs && isTopLevelVariable && !context.isModule) {
+            // top-level declarations in non-module files are considered global
+            assignment = lua.createAssignmentStatement(lhs, rhs, tsOriginal);
+        } else {
             const scope = peekScope(context);
 
             const isPossibleWrappedFunction =
@@ -175,11 +176,6 @@ export function createLocalOrExportedOrGlobalDeclaration(
                     declaration = undefined;
                 }
             }
-        } else if (rhs) {
-            // global
-            assignment = lua.createAssignmentStatement(lhs, rhs, tsOriginal);
-        } else {
-            return [];
         }
     }
 
--- a/test/translation/transformation/topLevelVariables.ts
+++ b/test/translation/transformation/topLevelVariables.ts
@@ -2,6 +2,7 @@ const obj = { value1: 1, value2: 2 };
 const value1 = obj.value1;
 const { value2 } = obj;
 
+let noValueLet;
 let obj2 = { value3: 1, value4: 2 };
 let value3 = obj2.value3;
 let { value4 } = obj2;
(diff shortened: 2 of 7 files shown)


test/unit/objectLiteral.spec.ts
```diff
@@ -8,59 +8,53 @@ test.each([
...
+    `.expectToMatchJsResult();
 });
 
 test.each([`({x: "foobar"}.x)`, `({x: "foobar"}["x"])`, `({x: () => "foobar"}.x())`, `({x: () => "foobar"}["x"]())`])(
     "object literal property access (%p)",
     expression => {
-        const code = `return ${expression}`;
-        const expectResult = eval(expression);
-        expect(util.transpileAndExecute(code)).toBe(expectResult);
+        util.testFunction`return ${expression}`.expectToMatchJsResult();
```
ark120202:
```suggestion
        util.testExpression(expression).expectToMatchJsResult();
```

test/unit/objectLiteral.spec.ts
```diff
@@ -8,59 +8,53 @@ test.each([
...
-            `return ({ _G })._G.foobar;`,
-            undefined,
-            `foobar = "foobar"`,
-            "declare const _G: any;"
-        );
-
-        expect(result).toBe("foobar");
+        const luaResult = util.testFunction`
+            return ({ _G })._G.foobar;
+        `
```
ark120202:
```ts
util.testExpression`({ _G })._G.foobar`
```

test/unit/objectLiteral.spec.ts
```diff
@@ -8,59 +8,53 @@ test.each([
...
+            .setTsHeader(`declare const _G: any;`)
+            .setLuaHeader(`foobar = "foobar"`)
+            .getLuaExecutionResult();
```
ark120202:
```suggestion
            .expectToEqual("foobar");
```

test/unit/objectLiteral.spec.ts
```diff
@@ -8,59 +8,53 @@ test.each([
...
-
-        expect(result).toBe(1);
+        `.expectToMatchJsResult();
     });
 
     test.each([NaN, Infinity])("should support %p shorthand", identifier => {
-        const result = util.transpileAndExecute(`return ({ ${identifier} }).${identifier}`);
-
-        expect(result).toBe(identifier);
+        util.testFunction`return ({ ${identifier} }).${identifier}`.expectToMatchJsResult();
```
ark120202:
```suggestion
        util.testExpression`({ ${identifier} }).${identifier}`.expectToMatchJsResult();
```

test/unit/objectLiteral.spec.ts
```diff
@@ -8,59 +8,53 @@ test.each([
     { inp: `{[myFunc()]:3,b:"4"}`, out: '{\n    [myFunc(_G)] = 3,\n    b = "4"\n}' },
     { inp: `{x}`, out: `{x = x}` },
 ])("Object Literal (%p)", ({ inp, out }) => {
-    const lua = util.transpileString(`const myvar = ${inp};`);
-    expect(lua).toBe(`local myvar = ${out}`);
+    const lua = util.testModule`const myvar = ${inp};`.getMainLuaCodeChunk();
+    expect(lua).toBe(`myvar = ${out}`);
```
ark120202: Maybe let's use snapshots there?

test/unit/assignments.spec.ts
```diff
@@ -1,13 +1,23 @@
...
+
+test("const declaration top-level is global", () => {
+    const lua = util.testModule`const foo = true;`.getMainLuaCodeChunk();
+    expect(lua).toBe(`foo = true`);
+});
+
+test("let declaration top-level is global", () => {
+    const lua = util.testModule`let foo = true;`.getMainLuaCodeChunk();
+    expect(lua).toBe(`foo = true`);
 });
```
ark120202: A more functional test:
```ts
util.testModule`
	${kind} foo = true;
	// @ts-ignore
	return (globalThis as any).foo;
`.expectToEqual(true)
```

test/unit/assignments.spec.ts
```diff
@@ -1,13 +1,23 @@
...
+    const lua = util.testFunction`const foo = true;`.getMainLuaCodeChunk();
+    expect(lua).toContain(`local foo = true`);
 });
 
 test("let declaration", () => {
-    const lua = util.transpileString(`let foo = true;`);
-    expect(lua).toBe(`local foo = true`);
+    const lua = util.testFunction`let foo = true;`.getMainLuaCodeChunk();
+    expect(lua).toContain(`local foo = true`);
+});
```
ark120202:
```ts
util.testModule`
	${kind} foo = true;
	// @ts-ignore
	return (globalThis as any).foo;
`.expectToEqual(undefined)
```

test/unit/objectLiteral.spec.ts
Perryvw: I'd prefer to not use snapshots for this test.

src/transformation/utils/lua-ast.ts
```diff
@@ -175,11 +176,6 @@ export function createLocalOrExportedOrGlobalDeclaration(
                     declaration = undefined;
                 }
             }
-        } else if (rhs) {
-            // global
-            assignment = lua.createAssignmentStatement(lhs, rhs, tsOriginal);
-        } else {
-            return [];
```
ark120202: Can we really drop that branch? Wouldn't `let foo` in global context be transformed to `local foo` int that case?

src/transformation/utils/lua-ast.ts
```diff
@@ -143,9 +141,12 @@ export function createLocalOrExportedOrGlobalDeclaration(
             );
         }
     } else {
-        const insideFunction = findScope(context, ScopeType.Function) !== undefined;
+        const isTopLevelVariable = peekScope(context).type === ScopeType.File;
```
ark120202:
```suggestion
		const scope = peekScope(context);
        const isTopLevelVariable = scope.type === ScopeType.File;
```

Scope is used below anyway

test/unit/objectLiteral.spec.ts
```diff
@@ -8,59 +8,49 @@ test.each([
     { inp: `{[myFunc()]:3,b:"4"}`, out: '{\n    [myFunc(_G)] = 3,\n    b = "4"\n}' },
     { inp: `{x}`, out: `{x = x}` },
 ])("Object Literal (%p)", ({ inp, out }) => {
-    const lua = util.transpileString(`const myvar = ${inp};`);
-    expect(lua).toBe(`local myvar = ${out}`);
+    const lua = util.testModule`const myvar = ${inp};`.getMainLuaCodeChunk();
```
ark120202:
```suggestion
    const lua = util.testModule`
    	const myvar = ${inp};
    `.getMainLuaCodeChunk();
```

test/translation/transformation/topLevelVariables.ts
```diff
@@ -0,0 +1,10 @@
+const obj = { value1: 1, value2: 2 };
```
ark120202: So do we want to continue using translation tests?

Perryvw: I think in a small number of cases they are useful, especially now we use snapshot testing for them.

src/transformation/utils/lua-ast.ts
Perryvw: You are right, reverted

src/transformation/utils/lua-ast.ts
```diff
@@ -143,11 +142,10 @@ export function createLocalOrExportedOrGlobalDeclaration(
...
         }
     } else {
-        const insideFunction = findScope(context, ScopeType.Function) !== undefined;
-
-        if (context.isModule || getCurrentNamespace(context) || insideFunction || isVariableDeclaration) {
-            const scope = peekScope(context);
+        const scope = peekScope(context);
+        const isTopLevelVariable = scope.type === ScopeType.File;
 
+        if (context.isModule || getCurrentNamespace(context) || !isTopLevelVariable) {
```
ark120202: Have you checked if `getCurrentNamespace` is still required? I think it should push some scope?

test/unit/assignments.spec.ts
ark120202: In case you've missed it, it actually wasn't resolved

test/unit/objectLiteral.spec.ts
```diff
@@ -1,66 +1,62 @@
...
-    { inp: `{a:3,b:"4"}`, out: '{a = 3, b = "4"}' },
-    { inp: `{"a":3,b:"4"}`, out: '{a = 3, b = "4"}' },
-    { inp: `{["a"]:3,b:"4"}`, out: '{a = 3, b = "4"}' },
-    { inp: `{["a"+123]:3,b:"4"}`, out: '{["a" .. 123] = 3, b = "4"}' },
-    { inp: `{[myFunc()]:3,b:"4"}`, out: '{\n    [myFunc(_G)] = 3,\n    b = "4"\n}' },
-    { inp: `{x}`, out: `{x = x}` },
-])("Object Literal (%p)", ({ inp, out }) => {
-    const lua = util.transpileString(`const myvar = ${inp};`);
-    expect(lua).toBe(`local myvar = ${out}`);
+test.each([`{a:3,b:"4"}`, `{"a":3,b:"4"}`, `{["a"]:3,b:"4"}`, `{["a"+123]:3,b:"4"}`])("Object Literal (%p)", inp => {
```
ark120202: I think we should try to preserve prettier code style even in literals
```suggestion
test.each([`{ a: 3, b: "4" }`, `{ "a": 3, b: "4" }`, `{ ["a"]: 3, b: "4" }`, `{ ["a" + 123]: 3, b: "4" }`])("Object Literal (%p)", inp => {
```

test/unit/objectLiteral.spec.ts
```diff
@@ -1,66 +1,62 @@
...
+    util.testExpression(inp).expectToMatchJsResult();
+});
+
+test("object literal with function call to get key", () => {
+    util.testFunction`
+        const myFunc = () => "a";
+    	return {[myFunc() + "b"]: 3};
```
ark120202:
```suggestion
        return { [myFunc() + "b"]: 3 };
```

(there's a tab instead of 4 spaces)

test/translation/transformation/topLevelVariables.ts
```diff
@@ -0,0 +1,11 @@
...
+const value1 = obj.value1;
+const { value2 } = obj;
+
+let noValueLet;
```
ark120202: I wonder if it would be better to transform it to a nil assignment instead of just dropping it?

test/unit/assignments.spec.ts
Perryvw: I left these tests but also added that test below, as I feel they actually test different things.

src/transformation/utils/lua-ast.ts
Perryvw: It does indeed push a block scope, removed the check

test/unit/assignments.spec.ts
ark120202: Well, the test you have added below is indeed different, but the one I've written above should be functionally equivalent

src/transformation/utils/lua-ast.ts
```diff
@@ -1,11 +1,10 @@
-import * as ts from "typescript";
 import * as assert from "assert";
+import * as ts from "typescript";
 import { LuaTarget } from "../../CompilerOptions";
 import * as lua from "../../LuaAST";
 import { TransformationContext } from "../context";
-import { getCurrentNamespace } from "../visitors/namespace";
```
ark120202: Can we also remove `getCurrentNamespace`?

test/unit/assignments.spec.ts
Perryvw: They are not equivalent. The current test asserts a variable statement with local is created, the other test just asserts there is no variable with that name in the global scope. That test would also pass if we ignore the declaration completely, or if we export it under some other name

test/translation/transformation/topLevelVariables.ts
Perryvw: Maybe we should decide and do this in another PR

PR #796, opened 2020-01-10
Feature/array includes

### LINKED ISSUE #794: Missing method Array.includes()
`Unsupported property on array: includes`. 
`Array.includes()` polyfill is needed.

Perryvw:
Closes #794

### DIFF
--- a/src/lualib/ArrayIncludes.ts
+++ b/src/lualib/ArrayIncludes.ts
@@ -1,0 +1,22 @@
+// https://www.ecma-international.org/ecma-262/9.0/index.html#sec-array.prototype.includes
+function __TS__ArrayIncludes<T>(this: T[], searchElement: T, fromIndex = 0): boolean {
+    const len = this.length;
+    let k = fromIndex;
+
+    if (fromIndex < 0) {
+        k = len + fromIndex;
+    }
+
+    if (k < 0) {
+        k = 0;
+    }
+
+    while (k < len) {
+        if (this[k] === searchElement) {
+            return true;
+        }
+        k = k + 1;
+    }
+
+    return false;
+}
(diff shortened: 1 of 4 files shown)


test/unit/builtins/array.spec.ts
```diff
@@ -389,6 +389,38 @@ test.each([
...
 
+test.each([
+    { array: [1, 2, 3], includes: 2 },
+    { array: [1, 2, 3, 2, 2], includes: 2 },
+    { array: [1, 2, 3], includes: 4 },
+    { array: ["a", "b", "c"], includes: "d" },
+    { array: [[1], [2], [3]], includes: [2] },
+    { array: [1, [2], 3], includes: 2 },
+])("array.includes (%p)", ({ array, includes }) => {
+    util.testExpression`${util.formatCode(array)}.includes(${util.formatCode(includes)})`.expectToMatchJsResult();
```
ark120202:
```suggestion
    util.testExpressionTemplate`${array}.includes(${includes})`.expectToMatchJsResult();
```

src/lualib/ArrayIncludes.ts
```diff
@@ -0,0 +1,22 @@
...
+    if (fromIndex < 0) {
+        k = len + fromIndex;
+    }
+
+    if (k < 0) {
+        k = 0;
+    }
+
+    while (k < len) {
+        if (this[k] === searchElement) {
```
ark120202: FWIW `.includes` uses a different comparison algorithm

```js
[NaN].indexOf(NaN) // => -1
[NaN].includes(NaN) // => true
```

But I'm okay with leaving it as an edge case

src/lualib/ArrayIncludes.ts
```diff
@@ -0,0 +1,22 @@
...
+
+    if (fromIndex < 0) {
+        k = len + fromIndex;
+    }
+
+    if (k < 0) {
+        k = 0;
+    }
+
+    while (k < len) {
```
ark120202: Thoughts on using `forRange` there?

Perryvw: I like the `while` because it is closer to both the ecmascript standard, and to the lua that will be generated. But I think forRange is faster, so I changed it.

PR #801, opened 2020-01-17
Feature/move parenthesis handling to printer

### LINKED ISSUE #798: Assignment operators missing parenthesis
When using arithmetic assignment operators, parenthesis should be added to the right operand.

# [Example](https://typescripttolua.github.io/play.html#src=let%20x%20%3D%200%3B%0Ax%20*%3D%202%2B3)

## Input

```ts
let x = 0;
x *= 2+3 // = 0
```

## Compiled

```lua
local x = 0
x = x * 2 + 3 --> 3
```

## Expected

```lua
local x = 0
x = x * (2 + 3) --> 0
```

Perryvw:
Moved parenthesis handling from transformer to printer.

Fixes #798


ark120202:
Can `lua.ParenthesizedExpression` node and some printing logic (i.e. `canStripParenthesis`) be also removed now?

Perryvw:
> Can `lua.ParenthesizedExpression` node and some printing logic (i.e. `canStripParenthesis`) be also removed now?

Maybe yes, I'll play around with it. Right now we no longer introduce new parenthesized expressions, but we do still preserve ts parenthesized expressions. I don't know if there is still an argument for doing this, since we will add the parentheses if necessary anyway.

PR #802, opened 2020-01-17
Bugfix/constructor assignment breaks keyword properties

### LINKED ISSUE #789: Constructor assignment breaks keyword properties
As can be seen in the [playground](https://typescripttolua.github.io/play.html#src=%0Aclass%20A%20%7B%0A%20%20%20%20constructor(public%20error%3A%20string)%7B%7D%0A%7D%0Aconst%20a%20%3D%20new%20A(%2242%22)%0Aconsole.log(a.error)%0A), the property name `error` get's renamed to `____error`.

# Input

```ts
class A {
    constructor(public error: string){}
}

const a = new A("42");
console.log(a.error);
```

# Current output

```
nil
```

# Expected output

```
42
```

Perryvw:
Fixes #789

### DIFF
--- a/test/unit/classes/classes.spec.ts
+++ b/test/unit/classes/classes.spec.ts
@@ -1,89 +1,73 @@
 import * as util from "../../util";
 
 test("ClassFieldInitializer", () => {
-    const result = util.transpileAndExecute(
-        `class a {
+    util.testFunction`
+        class a {
             field: number = 4;
         }
-        return new a().field;`
-    );
-
-    expect(result).toBe(4);
+        return new a().field;
+    `.expectToMatchJsResult();
 });
 
 test("ClassNumericLiteralFieldInitializer", () => {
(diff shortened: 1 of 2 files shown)


test/unit/classes/classes.spec.ts
```diff
@@ -706,118 +625,109 @@ test("Class Method Runtime Override", () => {
...
-                super("bar");
-            }
-        }
-        export const baz = (new Bar()).prop;
-    `;
-    expect(util.transpileExecuteAndReturnExport(code, "baz")).toBe("bar");
+    util.testExpression`
+        (new Bar()).prop;
+    `
+        .setTsHeader(
```
ark120202: I'm probably going to remove `setTsHeader` in the future, so it's better to test it with `testModule`

Perryvw:
Fixed with #847

PR #804, opened 2020-01-18
Stop prefixing identifiers from global augs

### LINKED ISSUE #803: Members of declare global are prefixed with global
# Input

```ts
export const _ = 0;

declare global {
    export const globalVariable: number;
}

globalVariable;
```

# Current Output (0.30.1)

```diff
  local ____exports = {}
  ____exports._ = 0
- local ____ = global.globalVariable
  return ____exports
```

# Expected Output

```diff
  local ____exports = {}
  ____exports._ = 0
+ local ____ = globalVariable
  return ____exports
```

This is TypeScript's output for the same code:

```js
"use strict";
exports.__esModule = true;
exports._ = 0;
globalVariable;
```

hazzard993:
Closes #803

### DIFF
--- a/src/transformation/visitors/identifier.ts
+++ b/src/transformation/visitors/identifier.ts
@@ -26,12 +26,16 @@ export function transformIdentifier(context: TransformationContext, identifier: 
     return lua.createIdentifier(text, identifier, symbolId, identifier.text);
 }
 
+function isGlobalAugmentation(module: ts.ModuleDeclaration): boolean {
+    return (module.flags & ts.NodeFlags.GlobalAugmentation) !== 0;
+}
+
 export const transformIdentifierExpression: FunctionVisitor<ts.Identifier> = (node, context) => {
     // TODO: Move below to avoid extra transforms?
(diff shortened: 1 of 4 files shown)


src/transformation/visitors/identifier.ts
```diff
@@ -26,12 +26,16 @@ export function transformIdentifier(context: TransformationContext, identifier:
...
+    return (module.flags & ts.NodeFlags.GlobalAugmentation) !== 0;
+}
+
 export const transformIdentifierExpression: FunctionVisitor<ts.Identifier> = (node, context) => {
     // TODO: Move below to avoid extra transforms?
     const identifier = transformIdentifier(context, node);
 
     const exportScope = getIdentifierExportScope(context, identifier);
-    if (exportScope) {
+    if (exportScope && !(ts.isModuleDeclaration(exportScope) && isGlobalAugmentation(exportScope))) {
```
ark120202: I think it's better to handle this inside `getSymbolExportScope`

PR #808, opened 2020-01-19
Remove `noHoisting` option

### LINKED ISSUE #651: Remove noHoisting option
As was discussed in Discord a while back.

As an alternative we should recommend a linter rule. TSLint's one seemingly doesn't cover all cases. We should try ESLint's `no-use-before-define`, or provide a new rule.

ark120202:
Resolves #651.

### DIFF
--- a/test/unit/annotations/vararg.spec.ts
+++ b/test/unit/annotations/vararg.spec.ts
@@ -13,7 +13,7 @@ test("@vararg", () => {
         }
         return bar("A", "B", "C", "D");
     `
-        .tap(builder => expect(builder.getMainLuaCodeChunk()).not.toMatch("{...}"))
+        .tap(builder => expect(builder.getMainLuaCodeChunk()).not.toMatch("b = "))
         .tap(builder => expect(builder.getMainLuaCodeChunk()).not.toMatch("unpack"))
         .expectToMatchJsResult();
 });
--- a/test/unit/annotations/vararg.spec.ts
+++ b/test/unit/annotations/vararg.spec.ts
@@ -1,9 +1,13 @@
 import * as util from "../../util";
 
+const varargDeclaration = `
+    /** @vararg */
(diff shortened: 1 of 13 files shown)


test/unit/annotations/vararg.spec.ts
```diff
@@ -11,36 +12,34 @@ test.each([{}, { noHoisting: true }])("@vararg", compilerOptions => {
...
-        /** @vararg */ type LuaVarArg<A extends unknown[]> = A & { __luaVarArg?: never };
+        /** @vararg */
+        type LuaVarArg<A extends unknown[]> = A & { __luaVarArg?: never };
         declare const arg: LuaVarArg<string[]>;
         const arr = [...arg];
         const result = arr.join("");
     `;
 
-    const luaBody = util.transpileString(code, compilerOptions, false);
+    const luaBody = util.transpileString(code, undefined, false);
```
Perryvw: If you're touching this anyway could you rewrite it to the new TestBuilder

test/unit/annotations/vararg.spec.ts
```diff
@@ -11,45 +16,30 @@ test.each([{}, { noHoisting: true }])("@vararg", compilerOptions => {
...
-        end
-        return test("A", "B", "C", "D")
-    `;
-
-    expect(util.executeLua(lua)).toBe("ABCD");
+        export const result = [...arg].join("");
+    `
+        .setLuaFactory(code => `return (function(...) ${code} end)("A", "B", "C", "D")`)
+        .tap(builder => expect(builder.getMainLuaCodeChunk()).not.toMatch("unpack"))
+        .expectToEqual({ result: "ABCD" });
```
Perryvw: Can we not just expectToMatchJsResult()? I think this setLuaFactory pattern is confusing, and don't really see the need for it.

ark120202: This test tests module-level arguments, which is a lua-only concept

Issue #584, opened 2019-05-17
Constructor extension decorator clobbers constructor name


thatcosmonaut:
```ts
export function Numbers(...message_type_args: number[]) {
    return <T extends new(...args: any[]) => {}>(constructor: T) => {
        return class extends constructor {
            protected numbers = message_type_args;
        };
    };
}

@Numbers(10, 20)
class MyClass {}

const my_instance = new MyClass();
console.log(my_instance.constructor.name);
```

Running the transpiled Lua code will give "____" as the constructor name.

ehemsley:
I have confirmed that the constructor name is correct when compiling to JS, so this is unintended behavior.

ehemsley:
Hmm, that's not exactly correct actually - when targeting ES5 the result is "class_1" and when targeting ES2017 the result is "MyClass".

tomblind:
ES2017 seems to have some smart logic for naming unnamed classes:
```ts
const x = new (class {})();
console.log(x.constructor.name); // ""

class A {}

const B = class extends A {}
const b = new B();
console.log(b.constructor.name); // "B"

function makeC() { return class extends A {}; }
const C = makeC();
const c = new C();
console.log(c.constructor.name); // "A"
```

ark120202:
Looks like it shares logic with unnamed function expressions. It's actually even more complex: http://2ality.com/2015/09/function-names-es6.html (might be partially outdated).

Looks like type checker has logic to infer it: (try on https://ts-ast-viewer.com/)
```ts
let { foo: bar = /* select class expression */ class {} } = {};
```

```js
ts.unescapeLeadingUnderscores(typeChecker.symbolToEntityName(symbol, symbol.flags).escapedText)
// => "bar"
```

However it fails when expression is dynamic:

```ts
const obj = { ["foo" + "bar"]: class {} };
```

```js
ts.unescapeLeadingUnderscores(typeChecker.symbolToEntityName(symbol, symbol.flags).escapedText)
// => "["foo" + "bar"]", should be "foobar"
```

PR #810, opened 2020-01-19
Bugfix/anonymous class name

Perryvw:
As discovered in #802, the class name property on anonymous classes did not match the class names in JS.

Not entirely sure about the modules.spec test, since it depends on the logic of the unique variable generation staying the same. Maybe we can come up with a better way to test that.


src/transformation/visitors/class/index.ts
```diff
@@ -349,3 +346,15 @@ export const transformSuperExpression: FunctionVisitor<ts.SuperExpression> = (ex
...
 
 export const transformThisExpression: FunctionVisitor<ts.ThisExpression> = node => createSelfIdentifier(node);
+
+function getReflectionClassName(context: TransformationContext, declaration: ts.ClassLikeDeclaration): string {
+    if (declaration.name) {
+        return declaration.name.text;
+    } else if (ts.isVariableDeclaration(declaration.parent) && ts.isIdentifier(declaration.parent.name)) {
+        return declaration.parent.name.text;
+    } else if (hasDefaultExportModifier(declaration)) {
+        return context.createUniqueString("default");
```
ark120202: Isn't it just `"default"`?

Perryvw: In my test JS produced 'default_1', although I am still a bit hesitant on this because I'm pretty sure we will deviate for multiple anyway (since numberings will be different).

ark120202: You have probably tested some transpiled version, in native ESM on node it's just `"default"`

src/transformation/visitors/class/index.ts
```diff
@@ -45,14 +45,7 @@ export function transformClassAsExpression(
     context: TransformationContext,
     isDefaultExport = false
 ): lua.Expression {
-    let className: lua.Identifier;
```
Perryvw: Revert this? Test this.

src/transformation/visitors/class/index.ts
```diff
@@ -349,3 +353,15 @@ export const transformSuperExpression: FunctionVisitor<ts.SuperExpression> = (ex
 };
 
 export const transformThisExpression: FunctionVisitor<ts.ThisExpression> = node => createSelfIdentifier(node);
+
+function getReflectionClassName(declaration: ts.ClassLikeDeclaration): string {
+    if (declaration.name) {
+        return declaration.name.text;
+    } else if (ts.isVariableDeclaration(declaration.parent) && ts.isIdentifier(declaration.parent.name)) {
```
ark120202: Have you seen https://github.com/TypeScriptToLua/TypeScriptToLua/issues/584#issuecomment-495465046? I'm okay with leaving it as it is for now, but note that it wouldn't cover a lot of cases

src/transformation/visitors/class/index.ts
```diff
@@ -45,7 +45,14 @@ export function transformClassAsExpression(
     context: TransformationContext,
     isDefaultExport = false
 ): lua.Expression {
-    const className = isDefaultExport ? createDefaultExportIdentifier(expression) : lua.createAnonymousIdentifier();
+    let className: lua.Identifier;
```
ark120202: Have you tried removing that part? I don't really see why we need to infer local name twice

Perryvw: It's tricky because of line 61

ark120202: Can we do something like this?
```ts
export function transformClassAsExpression(
    node: ts.ClassLikeDeclaration,
    context: TransformationContext
): lua.Expression {
    const { statements, className } = transformClassLikeDeclaration(node, context);
    return createImmediatelyInvokedFunctionExpression(statements, className, node);
}

export const transformClassDeclaration: FunctionVisitor<ts.ClassDeclaration> = (node, context) => {
	const { statements } = transformClassLikeDeclaration(node, context);
	return statements;
}
```

src/transformation/visitors/class/index.ts
ark120202: Since we have decided not to use type information for that, here's an implementation that should specifically handle #584

```ts
function getReflectionClassName(
    declaration: ts.ClassLikeDeclaration,
    className: lua.Identifier
): lua.Expression {
    if (declaration.name) {
        return lua.createStringLiteral(declaration.name.text);
    } else if (ts.isVariableDeclaration(declaration.parent) && ts.isIdentifier(declaration.parent.name)) {
        return lua.createStringLiteral(declaration.parent.name.text);
    } else if (hasDefaultExportModifier(declaration)) {
        return lua.createStringLiteral("default");
    }

    if (getExtendedNode(context, declaration)) {
        return lua.createTableIndexExpression(
            lua.createTableIndexExpression(className, lua.createStringLiteral("____super")),
            lua.createStringLiteral("name")
        );
    }

    return lua.createStringLiteral("");
}
```

src/transformation/visitors/class/index.ts
```diff
@@ -65,28 +70,19 @@ interface ClassSuperInfo {
     extendedTypeNode?: ts.ExpressionWithTypeArguments;
 }
 
-export function transformClassDeclaration(
+function transformClassLikeDeclaration(
     classDeclaration: ts.ClassLikeDeclaration,
     context: TransformationContext,
     nameOverride?: lua.Identifier
-): OneToManyVisitorResult<lua.Statement> {
+): { statements: OneToManyVisitorResult<lua.Statement>; name: lua.Identifier } {
```
ark120202, line 77:
```suggestion
): { statements: lua.Statement[]; name: lua.Identifier } {
```

Or maybe just drop this annotation, since it's just an internal helper with a simple return type that would be used only for destructuring?

src/transformation/visitors/class/index.ts
```diff
@@ -40,23 +40,28 @@ import { getExtendedNode, getExtendedType, isStaticNode } from "./utils";
...
     popScope(context);
 
-    return createImmediatelyInvokedFunctionExpression(classDeclaration, className, expression);
+    return createImmediatelyInvokedFunctionExpression(unwrapVisitorResult(statements), name, expression);
+}
+
+export function transformClassDeclaration(
+    declaration: ts.ClassLikeDeclaration,
+    context: TransformationContext
+): OneToManyVisitorResult<lua.Statement> {
```
ark120202, line 55: Is there a reason why `FunctionVisitor` wouldn't work there?

src/transformation/visitors/class/index.ts
```diff
@@ -197,7 +193,14 @@ export function transformClassDeclaration(
 
     if (!isExtension && !isMetaExtension) {
         result.push(
-            ...createClassSetup(context, classDeclaration, className, localClassName, classNameText, extendedType)
+            ...createClassSetup(
+                context,
+                classDeclaration,
+                className,
+                localClassName,
+                getReflectionClassName(classDeclaration),
```
ark120202, line 201: Should we move `getReflectionClassName` implementation to setup instead of passing it as a parameter? Unrelated, buy maybe `extendedType` too?

src/transformation/visitors/class/setup.ts
```diff
@@ -56,7 +56,7 @@ export function createClassSetup(
     result.push(
         lua.createAssignmentStatement(
             lua.createTableIndexExpression(lua.cloneIdentifier(localClassName), lua.createStringLiteral("name")),
-            lua.createStringLiteral(classNameText),
+            reflectionClassName,
```
ark120202, line 59:
```suggestion
            getReflectionClassName(statement),
```

src/transformation/visitors/class/setup.ts
```diff
@@ -79,3 +79,23 @@ export function createClassSetup(
 
     return result;
 }
+
+export function getReflectionClassName(
+    declaration: ts.ClassLikeDeclaration,
+    className: lua.Identifier,
+    context: TransformationContext
+): lua.Expression {
```
ark120202, line 87:
```suggestion
export function getReflectionClassName(
    context: TransformationContext,
    declaration: ts.ClassLikeDeclaration,
    className: lua.Identifier
): lua.Expression {
```

test/unit/classes/classes.spec.ts
```diff
@@ -846,12 +758,30 @@ test("Class field override in subclass with constructors", () => {
...
+        function Numbers(...message_type_args: number[]) {
+            return <T extends new(...args: any[]) => {}>(constructor: T) => {
+                return class extends constructor {
+                    protected numbers = message_type_args;
+                };
+            };
+        }
+        
+        @Numbers(10, 20)
+        class MyClass {}
```
ark120202, line 783:
```suggestion
        const decorator = <T extends new (...args: any[]) => any>(constructor: T) => class extends constructor {};
        
        @decorator
        class MyClass {}
```

test/unit/modules/modules.spec.ts
```diff
@@ -175,22 +175,20 @@ test("Export Default Function", () => {
...
-        {
-            "main.ts": `
-                import defaultExport from "./module";
-                export const value = defaultExport.name;
-            `,
-            "module.ts": classDeclarationStatement,
-        },
-        "value"
-    );
+test("default exported name class has correct name property", () => {
```
ark120202, line 178: I think these would be better with other class tests

src/transformation/visitors/class/index.ts
```diff
@@ -65,28 +68,19 @@ interface ClassSuperInfo {
     extendedTypeNode?: ts.ExpressionWithTypeArguments;
 }
 
-export function transformClassDeclaration(
+function transformClassLikeDeclaration(
     classDeclaration: ts.ClassLikeDeclaration,
     context: TransformationContext,
```
ark120202, line 73: For consistency:
```suggestion
    context: TransformationContext,
    classDeclaration: ts.ClassLikeDeclaration,
```

Perryvw: Actually it is consistent right now with transformClassDeclaration and transformClassAsExpression

Issue #580, opened 2019-05-15
Explore using custom types instead of tuple and iterator decorators


ark120202:
```ts
type LuaIterable<T> = Iterable<T> & { __luaIterable: never };
type LuaTuple<T extends any[]> = T & { __luaTuple: never };
```
Solution for #579. Also could simplify implementation.

tomblind:
Other candidates for custom declarations:
```ts
declare class LuaTable<K, V> {
    public readonly length: number;
    public get(key: K): V | undefined;
    public set(key: K, value: V | undefined): void;
}
```
```ts
declare function luaForRange(i: number, j: number, k?: number): Iterable<number>;
```

ark120202:
Values that aren't just types would better be exported, so it would work with modules and we'll be able to implement it both for Lua and JS. I'll write details about new lualib structure that would allow it later.

hazzard993:
I've been thinking of this setup for a typing solution that could be used for #579 instead of directives for both ambient and implemented code.

```ts
/** To mock some exports typescript-to-lua could provide. */
declare module "typescript-to-lua" {
    export interface TupleReturn<T extends any> extends Array<T> { __luaTuple: never }
    /**
     * Specially used to transform provided values into a Lua tuple when used in certain
     * positions.
     *
     * @example
     * const [a, b] = tuple(1, 2, 3);   // --> local a, b = 1, 2, 3
     * return tuple(1, 2, 3);           // --> return 1, 2, 3
     */
    export function tuple<T extends any>(...values: T[]): TupleReturn<T>;
}
```

For this example the transpiler treats the function `tuple` differently due to its origin or another distinguishing factor.

Its intended use is outlined in its TSDoc comment.

```ts
import { tuple, TupleReturn } from "typescript-to-lua";  // Elide these imports

class Base {
    tupleReturn() {                     // TupleReturn<number>
        return tuple(1, 2, 3);          // return 1, 2, 3
    }
}

class Extension1 extends Base {
    tupleReturn() {                     // number[] ❌ Prevented by TS
        return [1, 2, 3];               // return { 1, 2, 3 }
    }
}

class Extension2 extends Base {
    tupleReturn() {                     // TupleReturn<number>
        const [a, b] = tuple(1, 2, 3);  // local a, b = 1, 2, 3
        const values = tuple(1, 2, 3);  // local values = { 1, 2, 3 }
        const [c, d] = values;          // local c, d = unpack(values)
[...]

Issue #653, opened 2019-07-08
Lua Helpers Refactor


ark120202:
## Transformation changes

The first part is making builtins transformation more generic. Currently all helpers are used imperatively. A declarative solution should make code cleaner and allow some features that currently can't be implemented:

- Non-call references (`[1.5, 2.5].map(Math.floor)`)
- Expose static information about supported builtins (for language service plugin)
- Some sort of plugin system (since we can't really support polyfilling). It probably should use the same API as #615.

## Importing changes

The second part is helper files location.

First thing, I think we should completely drop inlining. TSTL helpers have a completely different use case compared to TS ones - they are a lot bigger and used more commonly (in TS they are used only for syntactic transforms, not for builtins). They are actually more similar to polyfills, which aren't inlined by any popular transpiler. In addition they can cause some huge issues due to having different references. The only use case for inlining is environments without module support, which should be covered by bundling.

Requirements:

1. Using one helper shouldn't load all of them
2. Helpers should be shared between modules when possible
3. Exposing global types (#580)
4. Exposing importable helpers (as a replacement for directives like `forRange` and `luaTable`)
5. Modules (#432) compiled with older version should continue working when used in application using newer one (even across major releases)

#### Option 1

Keep helpers where they are, import them with absolute requires.

- 4 becomes harder: the only way would be rewriting these imports, which also would cause issues with versioning.
- To do 5 we should never change helper signatures or remove them, and instead always create new helpers with different names.

#### Option 2

Distribute helpers as a separate package, import them as modules.

- Depends on #432, which, in turn, depends on this issue.
- Repository structure would have to become a bit more complex (monorepo).
- We'll be able to depend on modules in helpers, which could be useful for some huge features (promises, console-like formatting).

andrei-markeev:
how about

### Option 3

When creating `lualib_bundle.lua`, only include helpers that were really used in the code.

Perryvw:
> ### Option 3
> When creating `lualib_bundle.lua`, only include helpers that were really used in the code.

This should be possible to do now, either by generating one lualib_bundle per file or generating one lualib_bundle for the union of all lualib functions used across files (I'd probably go for the second).

Although I think this is probably the nicest way of doing the bundle, and should probably be the default, I consider this only a minor improvement over the current bundle. So pretty low on my priorities list, but PRs welcome!

YoRyan:
[...]

PR #821, opened 2020-02-06
Multi Helper

hazzard993:
For #580

Giving a ~helper~ language-extension implementation a try.

This PR makes TypeScriptToLua declare the `$multi` function and `MultiReturn` type. These can be used as a type safe alternatives to `@tupleReturn`.

You'll be able to use something like the `tsconfig.json` below to get access to `$multi`.

```json
{
  "compilerOptions": {
    "types": [
      "typescript-to-lua/language-extensions"
    ]
  }
}
```

This function can only be used in a `ReturnStatement` or as a return expression in an `ArrowFunction`.

```ts
const x = () => $multi(1, 2, 3);

function f() {
  return $multi(1, 2, 3);
}
```


src/helpers/index.ts
```diff
@@ -0,0 +1 @@
+export * from "./tuple";
```
ark120202: The location of such helpers is a very important thing, see #653 for some context. The main problem is that every module using these helpers must have reference to it's source in dependencies (either regular or peer), otherwise it would depend on hoisting with is unstable and disallowed in Yarn 2.

src/helpers/tuple.ts
```diff
@@ -0,0 +1,5 @@
+export type Tuple<T extends any[]> = T & {
+    " __tupleBrand": never;
+};
+
+export declare function tuple<T extends any[]>(...values: T): Tuple<T>;
```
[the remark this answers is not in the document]
hazzard993: Would `multi()` and `MultiReturn` be better names?

tsconfig.json
```diff
@@ -13,5 +13,5 @@
         "stripInternal": true
     },
     "include": ["src"],
-    "exclude": ["src/lualib"]
+    "exclude": ["src/lualib", "helpers"]
```
ark120202:
```suggestion
    "exclude": ["src/lualib"]
```
It's out of `include`

test/unit/helpers/multi.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+        })();
+    `
+        .setReturnExport("a")
+        .expectToEqual(result);
+});
+
+test("multi helper call with destructuring assignment side effects", () => {
+    util.testModule`
+        import { multi } from "typescript-to-lua";
+        let a, b;
```
ark120202:
```suggestion
        let a;
```

src/transpilation/index.ts
```diff
@@ -55,6 +55,16 @@ export function createVirtualProgram(input: Record<string, string>, options: Com
...
+                const content = fs.readFileSync(filePath, "utf8");
+
+                libCache[filename] = ts.createSourceFile(filePath, content, ts.ScriptTarget.Latest, false);
+
+                return libCache[filename];
+            }
+
             if (filename.startsWith("lib.")) {
                 if (libCache[filename]) return libCache[filename];
                 const typeScriptDir = path.dirname(require.resolve("typescript"));
```
ark120202:
```ts
            let filePath: string | undefined;

            if (filename.startsWith("lib.")) {
                const typeScriptDir = path.dirname(require.resolve("typescript"));
                filePath = path.join(typeScriptDir, filename);
            }

            if (filename.includes("typescript-to-lua")) {
                filePath = path.resolve(__dirname, "../../helpers/multi.ts");
            }

            if (filePath !== undefined) {
                if (libCache[filename]) return libCache[filename];
                const content = fs.readFileSync(filePath, "utf8");
                libCache[filename] = ts.createSourceFile(filename, content, ts.ScriptTarget.Latest, false);
                return libCache[filename];
            }
```

Also, `filename.includes("typescript-to-lua")` sounds a bit unsafe, is `filename` always relative (if it could be absolute it might fail if someone cloned the repo somewhere in the `typescript-to-lua` directory)

src/transformation/helpers/multi.ts
```diff
@@ -0,0 +1,171 @@
+import * as ts from "typescript";
```
ark120202: I think it could be moved to `transformation/visitors/helpers`

src/transformation/helpers/multi.ts
```diff
@@ -0,0 +1,171 @@
...
+import { getDependenciesOfSymbol, createExportedIdentifier } from "../utils/export";
+import { createLocalOrExportedOrGlobalDeclaration } from "../utils/lua-ast";
+
+function isMultiHelperDeclaration(declaration: ts.Declaration): boolean {
+    return utils.getHelperFileKind(declaration.getSourceFile()) === utils.HelperKind.Multi;
+}
+
+function isMultiHelperCallSignature(context: TransformationContext, expression: ts.CallExpression): boolean {
+    const type = context.checker.getTypeAtLocation(expression.expression);
+    return Boolean(type.symbol?.declarations?.some(isMultiHelperDeclaration));
```
ark120202: Actually, I think I like `type.symbol?.declarations?.some(isMultiHelperDeclaration) ?? false` or `(type.symbol?.declarations ?? []).some(isMultiHelperDeclaration)` a bit more

helpers/multi.ts
```diff
@@ -0,0 +1,5 @@
+export type MultiReturn<T extends any[]> = T & {
+    " __multiBrand": never;
+};
+
+export declare function multi<T extends any[]>(...values: T): MultiReturn<T>;
```
ark120202: Would it be possible to provide `.js` files with some js implementation?

hazzard993: Attempted an implementation, would this be for unit testing?

helpers/multi.d.ts
```diff
@@ -2,4 +2,6 @@ export type MultiReturn<T extends any[]> = T & {
     readonly " __multiBrand": unique symbol;
 };
 
+export { MultiReturn as MR };
```
ark120202: What's the reason for adding an alias? If length is a real issue, I'd rather rename `MultiReturn` to `Multi` than see 2 different names in declarations

helpers/multi.js
```diff
@@ -0,0 +1,8 @@
+module.exports.symbol = Symbol("Multi Brand Symbol");
+
+module.exports.multi = (...values) => {
+    return {
+        " __multiBrand": module.exports.symbol,
+        ...values,
+    };
+};
```
ark120202: `__multiBrand` is just an implementation detail, we might switch to `unique type` once typescript would add it, so I think just that should be enough:

```ts
module.exports.multi = (...values) => values;
```

helpers/multi.ts
ark120202: That's one use case, yes. It also should be useful for universal Lua/JS modules. `multi` is probably not that useful, but something like `forRange` might be

helpers/multi.d.ts
hazzard993: If devs were to use this instead of `/** @tupleReturn */` this type would be used quite a lot especially in declarations and LuaJIT code.

If removing `/** @tupleReturn */` this type would have to be added to the end of call signatures which are usually already quite long.

I ran into a similar scenario when I was using `React.FunctionComponent` and then discovered the `React.FC` alias. Granted `FunctionComponent` is longer than `TupleReturn`.

ark120202: > If devs were to use this instead of /** @tupleReturn */ this type would be used quite a lot especially in declarations and LuaJIT code.

Yes, but it's not like `MultiReturn` is a long name. That's just 11 chars, and unlike annotations you have autocomplete there

> If removing /** @tupleReturn */ this type would have to be added to the end of call signatures which are usually already quite long.

That's only a question of formatting, i.e. you can just write it like:

```ts
declare function myFunction(
    someParameter1: Parameter1,
    someParameter2: Parameter2,
    someParameter3: Parameter3,
    someParameter4: Parameter4,
    someParameter5: Parameter5,
    someParameter6: Parameter6,
): MultiReturn<[ReturnValue1, ReturnValue2, ...OtherReturnValues[]]>
```

> I ran into a similar scenario when I was using React.FunctionComponent and then discovered the React.FC alias. Granted FunctionComponent is longer than TupleReturn.

With `React.FC` at least you have a namespace as a hint (well, in the currently recommended import style), but `MR` is pretty much impossible to understand without context

hazzard993: Good points, removing this alias as it may lead to unnecessary confusion

src/helpers/index.ts
hazzard993: Can a new repo in the TypeScriptToLua organisation be created for helper types? I'll change this PR to work with declarations coming from that repo

ark120202: If we'd decide to store it in a new package I'd rather do it in a monorepo. But I'm not really sure if that's the best option. For helpers there's an alternative - make all of these declarations global (which also would make it easier to use in declaration files), and for js compatibility just make a `lua-compat` module, which would reexport these types with JS implementation.

hazzard993: Liking those ideas, what should happen with `multi.js`?

ark120202: It wouldn't be a part of TSTL, but a `lua-compat` module would reexport tstl's global helpers with a JS implementation

src/transformation/visitors/helpers/multi.ts
```diff
@@ -0,0 +1,170 @@
...
+        return undefined;
+    }
+
+    if (!ts.isCallExpression(statement.expression.right)) {
+        return undefined;
+    }
+
+    if (!isMultiReturningCallExpression(context, statement.expression.right)) {
+        return undefined;
+    }
```
ark120202: IMO single-statement condition is more readable in these cases
```suggestion
    if (!ts.isBinaryExpression(statement.expression)) return;
    if (statement.expression.operatorToken.kind !== ts.SyntaxKind.EqualsToken) return;
    if (!ts.isCallExpression(statement.expression.right)) return;
    if (!isMultiReturningCallExpression(context, statement.expression.right)) return;
```

helpers/multi.d.ts
```diff
@@ -0,0 +1,9 @@
+declare global {
+    export type MultiReturn<T extends any[]> = T & {
+        readonly " __multiBrand": unique symbol;
+    };
+
+    export function multi<T extends any[]>(...values: T): MultiReturn<T>;
+}
+
+export {};
```
ark120202: There's a little reason to force file to be a module and use `declare global`
```suggestion
function multi<T extends any[]>(...values: T): MultiReturn<T>;

type MultiReturn<T extends any[]> = T & { readonly " __multiBrand": unique symbol; };
```

src/transformation/visitors/helpers/multi.ts
```diff
@@ -0,0 +1,170 @@
...
+    if (!ts.isCallExpression(statement.expression)) {
+        return undefined;
+    }
+
+    if (!isMultiHelperCallSignature(context, statement.expression)) {
+        return undefined;
+    }
+
+    const expressions = transformArguments(context, statement.expression.arguments);
+    return lua.createReturnStatement(expressions, statement);
```
ark120202: How would this work with spread (especially after #832)?

test/unit/helpers/multi.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+        .expectToEqual(result);
+});
+
+test.each([
+    "multi",
+    "multi()",
+    "({ multi });",
+    "[] = multi()",
+    "const [] = multi();",
+    "const a = multi();",
```
ark120202: So, it doesn't support `tupleReturn`'s auto-packing? That's not necessarily a bad thing, but then it shouldn't be an array and `[...multi(...)]` should work

helpers/multi.d.ts
```diff
@@ -0,0 +1,9 @@
+declare global {
+    export type MultiReturn<T extends any[]> = T & {
```
ark120202: Looks like intersecting tuple type with anything makes it work incorrectly with destructuring 
```ts
const [_1, _2, _3, _4] = [1, '2', 3];
//                 ^^ Error

declare function $multi<T extends any[]>(...values: T): MultiReturn<T>;

type MultiReturn<T extends any[]> = T;
const [_1, _2, _3, _4] = $multi(1, '2', 3);
//                 ^^ Error

type MultiReturn<T extends any[]> = T & { foo: any };
const [_1, _2, _3, _4] = $multi(1, '2', 3);
//                 ^^ string | number

// Workaround: add an empty tuple
type MultiReturn<T extends any[]> = T & [] & { foo: any };
const [_1, _2, _3, _4] = $multi(1, '2', 3);
//                 ^^ never
```

hazzard993: Nice find, thanks for the info!

src/transformation/visitors/helpers/multi.ts
hazzard993: Current plan:

You can spread the result of a function that calls $multi.

```ts
[...callMulti(1, 2), 3]; // [1, 2, 3]
```

But you can't spread a $multi call.

```ts
[...$multi(1, 2), 3]; // error, use [1, 2, 3] instead
```

helpers/multi.d.ts
```diff
@@ -0,0 +1,9 @@
...
+        readonly " __multiBrand": unique symbol;
```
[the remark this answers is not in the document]
hazzard993: Yep, to suppress the completion result

test/unit/helpers/multi.spec.ts
hazzard993: I'd say `$multi` is meant more for clarifying function signatures in Lua, not as an alternate way to spread elements in an array

```ts
[...$multi(1, 2), 3] // [1, 2, 3]
// or write
[1, 2, 3]
```

Can you think of any use cases where `[...multi(...)]` should ideally work?

hazzard993: > So, it doesn't support `tupleReturn`'s auto-packing?

In regards to avoiding auto-packing, this is so no value will be of type `MultiReturn`

ark120202: > Can you think of any use cases where `[...multi(...)]` should ideally work?

No, I'm perfectly fine with disallowing it. But in that case I also don't see much use in `const [a, b, c] = $multi(1, 2, 3)`. Should we disallow `$multi` everywhere except return statements?

src/transformation/visitors/helpers/multi.ts
```diff
@@ -0,0 +1,182 @@
...
+        if (ts.isOmittedExpression(element)) {
+            expression = lua.createAnonymousIdentifier(element);
+        }
+
+        if (!expression) {
+            context.diagnostics.push(unsupportedMultiFunctionAssignment(element));
+            continue;
+        }
+
+        leftIdentifiers.push(expression);
```
ark120202: Since `ts.ArrayBindingElement` is a closed set, the only way where this condition is true is `ts.isBindingElement(element) && !ts.isIdentifier(element.name)`, which I think is cleaner to be handled explicitly.

```suggestion
        if (ts.isBindingElement(element)) {
            if (element.initializer) {
                context.diagnostics.push(invalidMultiReturnArrayBindingPatternElementInitializer(element));
            } else if (ts.isIdentifier(element.name)) {
                leftIdentifiers.push(transformIdentifier(context, element.name));
            } else {
                context.diagnostics.push(unsupportedMultiFunctionAssignment(element));
            }
        } else if (ts.isOmittedExpression(element)) {
            leftIdentifiers.push(lua.createAnonymousIdentifier(element));
        }
```

src/transformation/visitors/helpers/multi.ts
```diff
@@ -0,0 +1,182 @@
+import * as ts from "typescript";
+import * as lua from "../../../LuaAST";
+import * as utils from "../../utils/helpers";
```
ark120202: `utils` is a bit too broad given how many utils we do have. Maybe `helpers` or `helperUtils`?

src/transformation/visitors/literal.ts
```diff
@@ -62,6 +63,14 @@ const transformNumericLiteralExpression: FunctionVisitor<ts.NumericLiteral> = ex
 };
 
 const transformObjectLiteralExpression: FunctionVisitor<ts.ObjectLiteralExpression> = (expression, context) => {
+    const violations = findMultiHelperAssignmentViolations(context, expression);
+    if (violations.length > 0) {
+        violations.forEach(element => {
+            context.diagnostics.push(unsupportedMultiHelperFunctionPosition(element));
+        });
```
ark120202:
```suggestion
        context.diagnostics.push(...violations.map(e => unsupportedMultiHelperFunctionPosition(e)));
```

test/tsconfig.json
```diff
@@ -13,6 +13,7 @@
         "cli/watch",
         "transpile/directories",
         "transpile/outFile",
-        "../src/lualib"
+        "../src/lualib",
+        "../helpers"
```
ark120202: `helpers` is meant to be included only in scripts running in the Lua context, while our tests run in node. If we do need to include it in some `tsconfig.json` (which we would need for ESLint), `lualib` is probably the best one.

src/transformation/visitors/helpers/multi.ts
```diff
@@ -0,0 +1,182 @@
...
+
+function transformMultiHelperCallArguments(
+    context: TransformationContext,
+    expression: ts.CallExpression
+): lua.Expression[] | lua.Expression {
+    if (!isMultiHelperCallSignature(context, expression)) {
+        return context.transformExpression(expression);
+    }
+
+    if (expression.arguments.length < 1) {
```
ark120202:
```suggestion
    if (expression.arguments.length === 0) {
```

src/transformation/visitors/helpers/multi.ts
```diff
@@ -0,0 +1,182 @@
...
+        return [];
+    }
+
+    const transformLeft = (expression: ts.Expression): lua.AssignmentLeftHandSideExpression => {
+        if (ts.isOmittedExpression(expression)) {
+            return lua.createAnonymousIdentifier(expression);
+        } else {
+            return transformAssignmentLeftHandSideExpression(context, expression);
+        }
+    };
```
ark120202:
```suggestion
    const transformLeft = (expression: ts.Expression): lua.AssignmentLeftHandSideExpression =>
        ts.isOmittedExpression(expression)
            ? lua.createAnonymousIdentifier(expression)
            : transformAssignmentLeftHandSideExpression(context, expression);
```

test/unit/helpers/multi.spec.ts
hazzard993: I think that's a good idea, pushed a change to allow `$multi` to only be used in ReturnStatements

helpers/index.d.ts
```diff
@@ -0,0 +1 @@
+import "./multi";
```
ark120202, line 1: Would it maybe be simpler to just put all helpers into one file (and maybe rename it to `helpers.d.ts`), and check it in code based on declaration name?

helpers/multi.js
```diff
@@ -0,0 +1 @@
+module.exports.multi = (...values) => values;
```
ark120202, line 1: Since we're going with globals this file should be removed

src/transformation/utils/helpers.ts
```diff
@@ -0,0 +1,26 @@
...
+}
+
+export function getHelperFileKind(context: TransformationContext, sourceFile: ts.SourceFile): HelperKind | undefined {
+    if (isSourceFileFromHelpers(sourceFile)) {
+        const baseFileName = path.basename(sourceFile.fileName).replace(/(\.d)?\.ts$/g, "");
+        switch (baseFileName) {
+            case "multi":
+                return HelperKind.Multi;
+            default:
+                context.diagnostics.push(unknownHelperKind(sourceFile, baseFileName));
```
ark120202, line 23: I can't think of any valid cases where this would happen, so it should be enough to throw an error or a make failed assertion. 

> Would it maybe be simpler to just put all helpers into one file, and check it in code based on declaration name?

(if this would be resolved, same would apply to unknown declaration names)

test/unit/helpers/multi.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+import { unsupportedMultiHelperFunctionPosition } from "../../../src/transformation/utils/diagnostics";
+
+const multiProjectOptions: tstl.CompilerOptions = {
+    types: [path.resolve(__dirname, "../../../helpers")],
+};
+
+test.each<[string, any]>([
+    ["return $multi();", undefined],
+    ["return $multi(1);", 1],
+    ["const ar = [1]; return $multi(...ar);", 1],
```
ark120202, line 13: What should it do if there are multiple values being spread, i.e. `const ar = [1, 2, 3]; return $multi(...ar, ...ar);`?

Also would be nice to test spread over function returning multiple values (`$multi(...unpack([1, 2, 3]), ...unpack([4, 5, 6]))`)

src/transformation/visitors/helpers/multi.ts
```diff
@@ -52,102 +40,6 @@ export function transformMultiHelperReturnStatement(
...
-    }
-
-    if (expression.arguments.length === 0) {
-        return lua.createNilLiteral(expression);
-    }
-
-    return expression.arguments.map(e => context.transformExpression(e));
-}
-
-export function transformMultiHelperVariableDeclaration(
```
ark120202: Unless I'm missing something, in addition to just removing `const [a, b] = $multi(1, 2)` (which was discussed in https://github.com/TypeScriptToLua/TypeScriptToLua/pull/821#discussion_r400049568) it also drops support for `const [a, b] = unpack([1, 2])` (and other functions returning multiple values)

helpers/index.d.ts
hazzard993: I've merged helpers into `helpers/index.d.ts` for now since I can't quite figure out how to adjust the test builder factory to just use `helpers.d.ts` it seems to insist on using `helpers/index.d.ts` and I've yet to find a workaround so far

test/unit/helpers/multi.spec.ts
hazzard993: Currently it works like this which I think is pretty reasonable

Source:
```ts
function x(...args: any[]) {
  return $multi(...args, ...args);
}

function y() {
  return $multi(...x("a", "b", "c"));
}

console.log(x(1, 2, 3));
console.log(y());
```

Lua:

```lua
require("lualib_bundle");
function x(self, ...)
    local args = {...}
    return __TS__Unpack(
        __TS__ArrayConcat(
            { __TS__Unpack(args) },
            { __TS__Unpack(args) }
        )
    )
end
function y(self)
    return x(nil, "a", "b", "c")
end
print(x(nil, 1, 2, 3))
print(y(nil))
```

Output:
```
1       2       3       1       2       3
a       b       c       a       b       c
```

ark120202: What do you think about limiting it to just regular values and spread of `MultiReturn` as the last element? I see `$multi` as a helper that allows to express Lua syntax that isn't expressible in TS, so doing more stuff might be quite a surprising behavior

hazzard993: Interesting, I don't see much of a problem limiting the spread expressions used in `$multi`, could you provide some code examples on how this would work?

ark120202:
```ts
return $multi(1, 2, 3);                                 // return 1, 2, 3
return $multi(1, 2, 3, ...string.byte("abc", 1, 3));    // return 1, 2, 3, string.byte("abc", 1, 3)
return $multi(1, 2, 3, ...unpack([1, 2, 3]));           // return 1, 2, 3, unpack({1, 2, 3})
return $multi(1, 2, 3, ...[1, 2, 3]);                   // error
return $multi(1, ...string.byte("abc", 1, 3), 2, 3);    // error
((...args: any[]) => $multi(1, 2, 3, ...args));         // error
((...args: Vararg<any[]>) => $multi(1, 2, 3, ...args)); // return 1, 2, 3, ...
```

ark120202:
I think `helpers` name might need some bikeshedding. Thoughts on `macro`/`macros`?

lolleko:
> I think `helpers` name might need some bikeshedding. Thoughts on `macro`/`macros`?

I agree "helper" is too generic. but I don't think "macros" is the correct term.

Personally I would prefer "extension". It is still a very generic term, but that's the term used for c++ compiler/language extensions: (https://gcc.gnu.org/onlinedocs/gcc-4.8.5/gcc/C_002b_002b-Extensions.html , https://docs.microsoft.com/en-us/cpp/build/reference/microsoft-extensions-to-c-and-cpp?view=msvc-160) 
For me these helpers seem to be exactly that, an extension to TS/Lua syntax...

test/unit/helpers/multi.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+test.each<[string, any]>(createCasesThatCall("$multi"))("invalid direct $multi function use (%s)", statement => {
+    util.testModule`
+        ${multiFunction}
+        ${statement}
+        export { a };
+    `
+        .setOptions(multiProjectOptions)
+        .setReturnExport("a")
+        .expectDiagnosticsToMatchSnapshot([invalidMultiFunctionUse.code]);
+});
```
lolleko, line 51: Personally I think these tests are really hard to read/understand. But that might be because I haven't done a lot of js/ts lately.

Perryvw:
> > I think `helpers` name might need some bikeshedding. Thoughts on `macro`/`macros`?
> 
> I agree "helper" is too generic. but I don't think "macros" is the correct term.
> 
> Personally I would prefer "extension". It is still a very generic term, but that's the term used for c++ compiler/language extensions: (https://gcc.gnu.org/onlinedocs/gcc-4.8.5/gcc/C_002b_002b-Extensions.html , https://docs.microsoft.com/en-us/cpp/build/reference/microsoft-extensions-to-c-and-cpp?view=msvc-160)
> For me these helpers seem to be exactly that, an extension to TS/Lua syntax...

As discussed on discord, let's go with language-extensions if @hazzard993 agrees:
`import { multi } from "typescript-to-lua/language-extensions";`

hazzard993:
In regards to #947, `pcall` can best be defined like this using `MultiReturn`

```ts
type ToArray<T> = T extends any[] ? T : [T];

declare function pcall<F extends (...args: any[]) => any>(
  this: void,
  cb: F,
  ...args: Parameters<F>
): MultiReturn<
  | [false, string, ...undefined[]]
  | [true, ...ToArray<ReturnType<F>>]
>;
```

It looks like TypeScript needs a bit more work to assign the typings correctly

```ts
const [result, a, b, c] = pcall((a: number) => {
  return $multi(1, 2, 3);
}, 0);

result; // boolean
a; // string | number | true (expected string | number)
b; // string | number | boolean | undefined (expected number | undefined)
c; // string | number | boolean | undefined (expected number | undefined)
```

src/transformation/visitors/expression-statement.ts
```diff
@@ -4,8 +4,14 @@ import { FunctionVisitor } from "../context";
 import { transformBinaryExpressionStatement } from "./binary-expression";
 import { transformLuaTableExpressionStatement } from "./lua-table";
 import { transformUnaryExpressionStatement } from "./unary-expression";
+import { transformMultiDestructuringAssignmentStatement } from "./language-extensions/multi";
 
 export const transformExpressionStatement: FunctionVisitor<ts.ExpressionStatement> = (node, context) => {
+    const multiResult = transformMultiDestructuringAssignmentStatement(context, node);
```
Perryvw, line 10: Would prefer if you only call this `transformMultiDestructuringAssignment` function after checking `isMultiFunctionNode` (same for function.ts)

src/transformation/visitors/return.ts
```diff
@@ -59,6 +60,11 @@ export const transformReturnStatement: FunctionVisitor<ts.ReturnStatement> = (st
         insideTryCatch = insideTryCatch || scope.type === ScopeType.Try || scope.type === ScopeType.Catch;
     }
 
+    const result = transformMultiReturnStatement(context, statement);
```
Perryvw, line 63: Also here would be nice if you could check if this is a multi return before transforming. This is to prevent getting huge complexities randomly (happened before), because we are doing more work than we really need to.

src/transformation/visitors/variable-declaration.ts
```diff
@@ -229,6 +230,11 @@ export function transformVariableDeclaration(
     context: TransformationContext,
     statement: ts.VariableDeclaration
 ): lua.Statement[] {
+    const result = transformMultiVariableDeclaration(context, statement);
```
Perryvw, line 233: Also here please check first.

src/utils.ts
```diff
@@ -5,6 +5,7 @@ import * as path from "path";
 export function castArray<T>(value: T | T[]): T[];
 export function castArray<T>(value: T | readonly T[]): readonly T[];
 export function castArray<T>(value: T | readonly T[]): readonly T[] {
+    // @ts-ignore
```
Perryvw, line 8: Why was this added now? This function already exists in the current codebase and is fine.

src/transformation/visitors/expression-statement.ts
Perryvw, line 10: Could you check if this is a multireturn first before calling any transformation functions? We had a big performance issue a while ago where we did a lot more work than we needed because we were calling transformation functions too early. is
```ts
if (isMultiDestructuringAssignment(context, node)) {
    return transformMultiDestructuringAssignmentStatement(context, node);
}
```

src/transformation/visitors/function.ts
```diff
@@ -55,6 +56,11 @@ function isRestParameterReferenced(context: TransformationContext, identifier: l
 
 export function transformFunctionBodyContent(context: TransformationContext, body: ts.ConciseBody): lua.Statement[] {
     if (!ts.isBlock(body)) {
+        const result = transformMultiCallExpressionToReturnStatement(context, body);
```
Perryvw, line 59: Also here please check first if this is needed.

src/transformation/visitors/return.ts
Perryvw, line 63: Please check if this is MultiReturn here too first please.

src/transformation/visitors/variable-declaration.ts
Perryvw, line 233: Also check first here

Perryvw (commented):
Would it be possible to

src/utils.ts
hazzard993: Accidentally included, some issue with my local build

test/unit/language-extensions/multi.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+import {
+    invalidMultiFunctionUse,
+    invalidMultiTypeToNonArrayBindingPattern,
+    invalidMultiTypeArrayBindingPatternElementInitializer,
+} from "../../../src/transformation/utils/diagnostics";
+
+const multiProjectOptions: tstl.CompilerOptions = {
+    types: [path.resolve(__dirname, "../../../language-extensions")],
+};
+
```
Perryvw, line 13: Could you add one simple test that tests the main use case and also serves as example on how to use this. If I understand correctly it should be something like:
```ts
util.testModule`
    function multiReturn(): MultiReturn<[string, number]> {
        return $multi("foo", 5);
    }

    const [a, b] = multiReturn();
    export { a, b };
`.expectToEqual({a: "foo", b: 5});
```

Perryvw (commented):
Changes look good, would like you to add one or a couple extra tests (see comments)

test/unit/language-extensions/multi.spec.ts
hazzard993: Added this example use case

LoganDark:
Is `MultiReturn<[true]> | MultiReturn<[false, string]>` or `MultiReturn<[true] | [false, string]>` preferred? I've been using the former, but I don't know if that confuses anything or not.

Perryvw:
@LoganDark From the few experiments I did it does not really matter, I was hoping one would provide proper type narrowing, but seems like for now neither do that. Maybe we can come up with a way to update the type later. For now I would probably go with the second option personally, but you can do whatever you prefer.

LoganDark:
> For now I would probably go with the second option personally, but you can do whatever you prefer.

The second one does provide the benefit of being definitely wrapped in exactly one outer `MultiReturn`. I'm using the first one right now, maybe I'll switch if there turn out being benefits to the other.

LoganDark:
There are issues:

- `const bytes = string.byte(...)` errors ("expected a destructuring pattern") because `string.byte` returns a `MultiReturn`. I now have to do `const bytes = [...string.byte(...)]` which is annoying
- `string.byte(...)[1]` generates `string.byte("0")[1]` in Lua, which is incorrect. It should either do `(string.byte("0"))`, or `({string.byte("0")})[1]` like it did before

Perryvw:
Instead of wrapping the right side you can also do it on the left instead: `const [...bytes] = string.byte(...);` to capture all, or `const [byte] = string.byte(...)` to capture the first one, or even a combination of the two: `const [byte1, ...otherBytes] = string.byte(...)`.

LoganDark:
> Instead of wrapping the right side you can also do it on the left instead: `const [...bytes] = string.byte(...);` to capture all, or `const [byte] = string.byte(...)` to capture the first one, or even a combination of the two: `const [byte1, ...otherBytes] = string.byte(...)`.

Yeah, I suppose that's reasonable. It's still annoying that I have to spread it, though. I expected it to act just like the tuples did before.

PR #829, opened 2020-02-24
Resolve exported identifiers in for-in loops

### LINKED ISSUE #781: `for...in` loop doesn't resolve exported identifiers
Input:
```ts
export let foo = '';
for (foo in { x: true }) {
    print(foo);
}
```

Current Result:
```lua
____exports.foo = ""
for ____key in pairs({x = true}) do
    foo = ____key
    print(____exports.foo)
end
```

Expected Result:
```lua
____exports.foo = ""
for ____key in pairs({x = true}) do
    ____exports.foo = ____key
    print(____exports.foo)
end
```

hazzard993:
Closes #781

Not entirely sure what

```ts
let foo = '';
export { foo as bar };
for (foo in { x: true }) {}
```

is meant to change `bar` to.

TypeScript doesn't change `bar` specifically in for-in loops.

### DIFF
--- a/test/unit/modules/modules.spec.ts
+++ b/test/unit/modules/modules.spec.ts
@@ -253,3 +253,22 @@ test("export as specifier shouldn't effect local vars", () => {
         a = 6;
     `.expectToMatchJsResult();
 });
+
+test("export modified in for in loop", () => {
+    util.testModule`
+        export let foo = '';
+        for (foo in { x: true }) {}
+    `
+        .setReturnExport("x")
+        .expectToMatchJsResult();
+});
+
+test("export dependency modified in for in loop", () => {
+    util.testModule`
+        let foo = '';
+        export { foo as bar };
+        for (foo in { x: true }) {}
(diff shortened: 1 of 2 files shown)


test/unit/modules/modules.spec.ts
```diff
@@ -253,3 +253,22 @@ test("export as specifier shouldn't effect local vars", () => {
         a = 6;
     `.expectToMatchJsResult();
 });
+
+test("export modified in for in loop", () => {
+    util.testModule`
+        export let foo = '';
+        for (foo in { x: true }) {}
+    `
+        .setReturnExport("x")
```
ark120202: Isn't it supposed to check `foo`?

Perryvw: Good catch

Perryvw: Fixed 770d5ffdc5cd1021fab3a248f0e391fd88307a27

PR #830, opened 2020-02-25
Fix default exports reference bug

### LINKED ISSUE #787: References to named default exports do not correctly reference the export
When exporting a class as default, later references to the class are not correctly translated. The actual export is defined as `____exports.default`, but later references attempt to use `____exports.<className>`. [Here is an MCVE.](https://typescripttolua.github.io/play.html#src=export%20default%20class%20Test%20%7B%0A%0A%09static%20builder()%3A%20Test%20%7B%0A%0A%09%09return%20new%20Test()%3B%0A%0A%09%7D%0A%0A%7D)

This can likely be fixed by using either `____exports.default` or just the class name.

### LINKED ISSUE #813: Mistranspilation when calling a static method on a default export class
```ts
export default class Bug {
    static init() {}
}

Bug.init()
```
Transpiles to:
```lua
require("lualib_bundle");
local ____exports = {}
____exports.default = __TS__Class("Bug")
local Bug = ____exports.default
Bug.name = "Bug"
function Bug.prototype.____constructor(self)
end
function Bug.init(self)
end
____exports.Bug:init()
return ____exports
```
The mistranspilation is at line `____exports.Bug:init()`. `____exports.Bug` is never set, resulting in a runtime error when trying to run the code.

Should probably be `____exports.default:init()` instead.

hazzard993:
Fixes #787
Also Fixes #813

### DIFF
--- a/src/transformation/visitors/identifier.ts
+++ b/src/transformation/visitors/identifier.ts
@@ -20,10 +20,12 @@ export function transformIdentifier(context: TransformationContext, identifier: 
         }
     }
 
-    const text = hasUnsafeIdentifierName(context, identifier) ? createSafeName(identifier.text) : identifier.text;
+    const symbol = context.checker.getSymbolAtLocation(identifier);
+    const name = symbol ? ts.unescapeLeadingUnderscores(symbol.escapedName) : identifier.text;
+    const text = hasUnsafeIdentifierName(context, identifier) ? createSafeName(name) : name;
 
     const symbolId = getIdentifierSymbolId(context, identifier);
-    return lua.createIdentifier(text, identifier, symbolId, identifier.text);
+    return lua.createIdentifier(text, identifier, symbolId, name);
 }
 
 export const transformIdentifierExpression: FunctionVisitor<ts.Identifier> = (node, context) => {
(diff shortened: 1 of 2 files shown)


src/transformation/visitors/identifier.ts
```diff
@@ -20,10 +20,12 @@ export function transformIdentifier(context: TransformationContext, identifier:
         }
     }
 
-    const text = hasUnsafeIdentifierName(context, identifier) ? createSafeName(identifier.text) : identifier.text;
+    const symbol = context.checker.getSymbolAtLocation(identifier);
+    const name = symbol ? ts.unescapeLeadingUnderscores(symbol.escapedName) : identifier.text;
```
ark120202:
```suggestion
    const name = symbol?.name ?? identifier.text;
```

src/transformation/visitors/identifier.ts
```diff
@@ -20,10 +20,12 @@ export function transformIdentifier(context: TransformationContext, identifier:
...
     }
 
-    const text = hasUnsafeIdentifierName(context, identifier) ? createSafeName(identifier.text) : identifier.text;
+    const symbol = context.checker.getSymbolAtLocation(identifier);
+    const name = symbol?.name ?? identifier.text;
+    const text = hasUnsafeIdentifierName(context, identifier) ? createSafeName(name) : name;
 
     const symbolId = getIdentifierSymbolId(context, identifier);
-    return lua.createIdentifier(text, identifier, symbolId, identifier.text);
+    return lua.createIdentifier(text, identifier, symbolId, name);
```
ark120202: I wonder if it fits `transformIdentifierExpression` better?

PR #832, opened 2020-02-27
Intermediate SpreadElements fix

### LINKED ISSUE #696: Array spread shouldn't use unpack
Tracking unresolved part of https://github.com/TypeScriptToLua/TypeScriptToLua/issues/121#issuecomment-396013177

hazzard993:
Fixes #696 

```diff
- before
+ after

- [1, ...[2, 3], 4] // --> { 1, 2, 4 }
+ [1, ...[2, 3], 4] // --> { 1, 2, 3, 4 }

- ((...values) => values)(1, ...[2, 3], 4) // --> { 1, 2, 4 }
+ ((...values) => values)(1, ...[2, 3], 4) // --> { 1, 2, 3, 4 }
```

Uses ArrayConcat only when necessary to replicate TS behavior

```
[1, ...[2, 3], 4, ...[5, 6]]
    ~~~~~~~~~~~~~~~~~~~~~~~
Using Array.concat here

[1, [2, 3], 4]

Array.concat not needed

[1, ...[2, 3]]

Array.concat not needed
```


src/transformation/visitors/call.ts
```diff
@@ -16,13 +16,75 @@ import { transformLuaTableCallExpression } from "./lua-table";
...
+            const lastChunk = chunks[chunks.length - 1];
+            lastChunk.push(argument);
+        }
+        return chunks;
+    }, []);
+
+    return chunks.map(chunk => wrapInTable(...chunk.map(expression => context.transformExpression(expression))));
+}
+
+export function flattenExpressions(
```
ark120202:
```suggestion
export function flattenSpreadExpressions(
```

src/transformation/visitors/call.ts
```diff
@@ -16,13 +16,75 @@ import { transformLuaTableCallExpression } from "./lua-table";
...
+    return [before, after];
+}
+
+function transformSpreadableExpressionsIntoArrayConcatArguments(
+    context: TransformationContext,
+    expressions: readonly ts.Expression[] | ts.NodeArray<ts.Expression>
+): lua.Expression[] {
+    // [...array, a, b, ...tuple()] --> [ [...array], [a, b], [...tuple()] ]
+    // chunk non-spread arguments together so they don't concat
+    const chunks = expressions.reduce<ts.Expression[][]>((chunks, argument, index, expressions) => {
```
ark120202: I don't really see how `reduce` makes code more readable there, when the reduced value is always the same. IMO it would be better as a regular for...of loop

src/transformation/visitors/literal.ts
```diff
@@ -137,13 +138,10 @@ const transformObjectLiteralExpression: FunctionVisitor<ts.ObjectLiteralExpressi
...
 
 const transformArrayLiteralExpression: FunctionVisitor<ts.ArrayLiteralExpression> = (expression, context) => {
-    const values = expression.elements.map(element =>
-        lua.createTableFieldExpression(
-            ts.isOmittedExpression(element) ? lua.createNilLiteral(element) : context.transformExpression(element),
-            undefined,
-            element
-        )
+    const filteredElements = expression.elements.map(e =>
+        ts.isOmittedExpression(e) ? ts.createIdentifier("undefined") : e
```
ark120202: Have you checked that `ts.createIdentifier("undefined")`  is actually transformed to `nil`?

test/unit/spread.spec.ts
```diff
@@ -87,3 +87,22 @@ describe("in object literal", () => {
...
     });
 });
+
+const spreadCases = [
+    "1, 2, ...[3, 4, 5]",
+    "...[1, 2], 3, 4, 5",
+    "1, 2, ...'spread', 4, 5",
+    "1, [2], ...[3, 4, 5]",
+    "...[1, 2, 3], 4, ...[5, 6]",
+];
```
ark120202: It would be nice to add a test case with a nested array value, to test that `.concat` doesn't flatten args

test/unit/spread.spec.ts
```diff
@@ -87,3 +87,22 @@ describe("in object literal", () => {
...
+
+test.each(spreadCases)("spread equivalence for array elements (%p)", expression => {
+    util.testExpression`[${expression}]`.expectToMatchJsResult();
+});
+
+test.each(spreadCases)("spread equivalence for call arguments (%p)", expression => {
```
ark120202: Like with `in object literal/of object literal`, I think that should be tested as a part of `in function|array call/of array literal`

src/transformation/visitors/literal.ts
hazzard993: That is tested over at array.spec.ts

Perryvw: Why not just use actual `nil` here

src/transformation/visitors/call.ts
```diff
@@ -16,13 +16,75 @@ import { transformLuaTableCallExpression } from "./lua-table";
...
+    const transformedPreSpreadExpressions = preSpreadExpressions.map(a => context.transformExpression(a));
+
+    // Nothing special required
+    if (postSpreadExpressions.length === 0) {
+        return transformedPreSpreadExpressions;
+    }
+
+    // Only one spread element at the end? Will work as expected
+    const [firstExpression] = postSpreadExpressions;
+    if (postSpreadExpressions.length === 1 && ts.isSpreadElement(firstExpression)) {
```
Perryvw: isn't `ts.isSpreadElement(firstExpression)` always true by definition of `getExpressionsBeforeAndAfterFirstSpread`?

src/transformation/visitors/call.ts
```diff
@@ -16,13 +16,75 @@ import { transformLuaTableCallExpression } from "./lua-table";
...
+    for (const expression of expressions) {
+        if (ts.isSpreadElement(expression)) {
+            chunks.push([expression]);
+            const next = expressions[expressions.indexOf(expression) + 1];
+            if (next && !ts.isSpreadElement(next)) {
+                chunks.push([]);
+            }
+        } else {
+            const lastChunk = chunks[chunks.length - 1];
+            lastChunk.push(expression);
```
Perryvw: Seems like `[a, b, ...array, ...c]` would cause this function to throw an error because this line will reference an undefined, could you make it so it also works for this case?

test/unit/spread.spec.ts
```diff
@@ -92,7 +92,7 @@ const spreadCases = [
     "1, 2, ...[3, 4, 5]",
     "...[1, 2], 3, 4, 5",
     "1, 2, ...'spread', 4, 5",
-    "1, [2], ...[3, 4, 5]",
+    "1, 2, ...[3, 4, 5], [2]",
```
ark120202: I have meant something like `[1, ...[[2]], 3]`, since both `[1].concat(2)` and `[1].concat([2])` are `[1, 2]`

test/unit/spread.spec.ts
```diff
@@ -5,6 +5,14 @@ import * as util from "../util";
 const expectUnpack: util.TapCallback = builder => expect(builder.getMainLuaCodeChunk()).toMatch(/[^.]unpack\(/);
 const expectTableUnpack: util.TapCallback = builder => expect(builder.getMainLuaCodeChunk()).toContain("table.unpack");
 
+const spreadCases = [
```
ark120202: It's actually also the case with `tupleReturn` call spread, not sure how it is handled by the current implementation: https://typescripttolua.github.io/play#code/A4JwlgdgLgFAUAAiQgdGqBDARgGwKYoCuEwGAxgNYwDaAjADQIBMjAzALoCU9iyaKmXAWKlKNACyMArIwBsXHpyA

src/transformation/visitors/call.ts
```diff
@@ -16,13 +16,75 @@ import { transformLuaTableCallExpression } from "./lua-table";
...
 export type PropertyCallExpression = ts.CallExpression & { expression: ts.PropertyAccessExpression };
 
+function getExpressionsBeforeAndAfterFirstSpread(
+    expressions: readonly ts.Expression[]
+): [readonly ts.Expression[], readonly ts.Expression[]] {
+    // [a, b, ...c, d, ...e] --> [a, b] and [...c, d, ...e]
+    const index = expressions.findIndex(ts.isSpreadElement);
+    const hasSpreadElement = index !== -1;
+    const before = hasSpreadElement ? expressions.slice(0, index) : expressions;
+    const after = hasSpreadElement ? expressions.slice(index, expressions.length) : [];
```
ark120202:
```suggestion
    const after = hasSpreadElement ? expressions.slice(index) : [];
```

src/transformation/visitors/call.ts
```diff
@@ -16,13 +16,75 @@ import { transformLuaTableCallExpression } from "./lua-table";
...
+    context: TransformationContext,
+    expressions: readonly ts.Expression[] | ts.NodeArray<ts.Expression>
+): lua.Expression[] {
+    // [...array, a, b, ...tuple()] --> [ [...array], [a, b], [...tuple()] ]
+    // chunk non-spread arguments together so they don't concat
+    const chunks: ts.Expression[][] = [];
+    for (const expression of expressions) {
+        if (ts.isSpreadElement(expression)) {
+            chunks.push([expression]);
+            const next = expressions[expressions.indexOf(expression) + 1];
```
ark120202:
```suggestion
    for (const [index, expression] of expressions.entries()) {
        if (ts.isSpreadElement(expression)) {
            chunks.push([expression]);
            const next = expressions[index + 1];
```

test/unit/spread.spec.ts
```diff
@@ -5,6 +5,14 @@ import * as util from "../util";
...
+    "1, 2, ...[3, 4, 5]",
+    "...[1, 2], 3, 4, 5",
+    "1, 2, ...'spread', 4, 5",
```
ark120202: Should string spread be there? I'd rather test it as a part of `of string literal` (it also could be added to `in function call`)

src/transformation/visitors/literal.ts
ark120202: > Why not just use actual nil here

It's TS AST, not Lua

> That is tested over at array.spec.ts

Have you actually checked it's output? Since `undefined` is not defined this test might pass even if it wouldn't be correctly transformed

hazzard993: Added c8ebe4e to use spread and check for the expected undefined value

hazzard993: > Why not just use actual `nil` here

I intended `flattenSpreadExpressions` to transform TS expressions without caring where those expressions are

`OmittedExpression` has no transformer so it will break if one gets to `flattenSpreadExpressions`

src/transformation/visitors/call.ts
hazzard993: true, that can be refactored

test/unit/spread.spec.ts
```diff
@@ -33,6 +48,32 @@ describe("in array literal", () => {
...
+    });
+
+    test("of array literal /w OmittedExpression", () => {
+        util.testFunction`
+            const array = [1, 2, ...[3], , 5];
+            return { a: array[0], b: array[1], c: array[2], d: array[3] };
+        `.expectToMatchJsResult();
+    });
+
+    test("of string literal mixed", () => {
```
Perryvw: What's mixed about this?

test/unit/spread.spec.ts
```diff
@@ -33,6 +48,32 @@ describe("in array literal", () => {
         [tstl.LuaTarget.Lua53]: builder => builder.tap(expectTableUnpack).expectToMatchJsResult(),
     });
 
+    test("of tuple return call", () => {
```
ark120202: Would be nice to test it in function call as well.
Maybe the duplication there could be reduced? With something like
```ts
describe.each(["function call",	"array literal"])("in %s", kind => {
	const factory = code => kind === "function call"
		? `((...args: any[]) => args)(${code})`
		: `[${code}]`;

	test.each(spreadCases)("of array literal (%p)", expression => {
        util.testExpression(factory(expression)).expectToMatchJsResult();
    });
});
```

Perryvw: wew, I would rather have some duplication if it keeps our tests clear

test/unit/spread.spec.ts
hazzard993: Changed to just `of string literal`

test/unit/spread.spec.ts
ark120202: That's not just some duplication: since both cases follow a single code path pretty much all cases would be the same, except for `OmittedExpression` since it's valid only in arrays and target-specific tests. I don't really see how a simple factory function makes these cases harder to read, but it makes it easier to keep them consistent (for example even now `of tuple return` test has different name and function call doesn't have a `of iterable` case). That's how all of the current tests would look:

```ts
describe.each(["function call",	"array literal"])("in %s", kind => {
    const factory = code => kind === "function call"
        ? `((...args: any[]) => args)(${code})`
        : `[${code}]`;

    test.each(arrayLiteralCases)("of array literal (%p)", expression => {
        util.testExpression(factory(expression)).expectToMatchJsResult();
    });

    test.each(arrayLiteralCases)("of tuple return call", expression => {
        util.testFunction`
            ${tupleReturnDefinition}
            return ${factory(`...tuple(${expression})`)};
        `.expectToMatchJsResult();
    });

    test("of string literal", () => {
        util.testExpression(factory('..."spread", ..."string"')).expectToMatchJsResult();
    });

    // TODO:
    test.each(["", "string", "string with spaces", "string 1 2 3"])("of string literal (%p)", str => {
        util.testExpression(factory(`...${str}`)).expectToMatchJsResult();
    });

    test("of iterable", () => {
        util.testFunction`
            const it = {
                i: -1,
                [Symbol.iterator]() {
                    return this;
                },
                next() {
                    ++this.i;
                    return {
                        value: 2 ** this.i,
                        done: this.i == 9,
                    }
                }
            };

            return ${factory('...it')};
        `.expectToMatchJsResult();
    });
});

describe("in function call", () => {
    util.testEachVersion(
        undefined,
        () => util.testFunction`
            function foo(a: number, b: number, ...rest: number[]) {
                return { a, b, rest }
            }
            const array = [0, 1, 2, 3] as const;
            return foo(...array);
        `,
        {
            [tstl.LuaTarget.LuaJIT]: builder => builder.tap(expectUnpack),
            [tstl.LuaTarget.Lua51]: builder => builder.tap(expectUnpack),
            [tstl.LuaTarget.Lua52]: builder => builder.tap(expectTableUnpack),
            [tstl.LuaTarget.Lua53]: builder => builder.tap(expectTableUnpack).expectToMatchJsResult(),
        }
    );

});

describe("in array literal", () => {
    util.testEachVersion(undefined, () => util.testExpression`[...[0, 1, 2]]`, {
        [tstl.LuaTarget.LuaJIT]: builder => builder.tap(expectUnpack),
        [tstl.LuaTarget.Lua51]: builder => builder.tap(expectUnpack),
        [tstl.LuaTarget.Lua52]: builder => builder.tap(expectTableUnpack),
        [tstl.LuaTarget.Lua53]: builder => builder.tap(expectTableUnpack).expectToMatchJsResult(),
    });

    test("of array literal /w OmittedExpression", () => {
        util.testFunction`
            const array = [1, 2, ...[3], , 5];
            return { a: array[0], b: array[1], c: array[2], d: array[3] };
        `.expectToMatchJsResult();
    });
});
```

PR #837, opened 2020-03-13
handle non-indexable types for lefthand value of instanceof

### LINKED ISSUE #835: __TS__InstanceOf errors when left-hand value compiles to userdata, function, boolean or number
All of the following if-statements will have the same error with their respective types.
```typescript
class SomeClass {
}

const co: any = coroutine.create(() => {});
if (co instanceof SomeClass) {}

const fn: any = () => {};
if (fn instanceof SomeClass) {}

const bool: any = true;
if (bool instanceof SomeClass) {}

const num: any = 5;
if (num instanceof SomeClass) {}
```

```text
/usr/local/opt/lua/bin/lua5.3: dist/bundle.lua:716: attempt to index a thread value (local 'obj')
stack traceback:
	dist/bundle.lua:716: in function '__TS__InstanceOf'
```
Some are a bit contrived, and casting as any was need to get TS to compile for some of them. However, a more real-life case happens when you have a type union:

```typescript
class SomeClass {
    constructor(public someFn: (this: void) => void) {}
}

function takesClassOrFunction(clsOrFn: SomeClass | ((this:void) => void)): void {

    if (clsOrFn instanceof SomeClass) {
        return clsOrFn.someFn()
    } else {
        return clsOrFn();
    }
}

const noop = () => {};
const cls = new SomeClass(noop);
takesClassOrFunction(cls);
takesClassOrFunction(noop);
```

cdbajorin:
Fixes #835

### DIFF
--- a/src/lualib/InstanceOf.ts
+++ b/src/lualib/InstanceOf.ts
@@ -8,7 +8,7 @@ function __TS__InstanceOf(this: void, obj: LuaClassInstance, classTbl: LuaClass)
         return !!classTbl[Symbol.hasInstance](obj);
     }
 
-    if (type(obj) === "table") {
+    if (typeof obj === "object") {
         let luaClass = obj.constructor;
         while (luaClass !== undefined) {
             if (luaClass === classTbl) {
--- a/src/lualib/InstanceOf.ts
+++ b/src/lualib/InstanceOf.ts
@@ -8,7 +8,7 @@ function __TS__InstanceOf(this: void, obj: LuaClassInstance, classTbl: LuaClass)
         return !!classTbl[Symbol.hasInstance](obj);
     }
 
-    if (obj !== undefined) {
+    if (type(obj) === "table") {
         let luaClass = obj.constructor;
         while (luaClass !== undefined) {
             if (luaClass === classTbl) {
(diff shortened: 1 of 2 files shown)


src/lualib/InstanceOf.ts
```diff
@@ -8,7 +8,7 @@ function __TS__InstanceOf(this: void, obj: LuaClassInstance, classTbl: LuaClass)
         return !!classTbl[Symbol.hasInstance](obj);
     }
 
-    if (obj !== undefined) {
+    if (type(obj) === "table") {
```
ark120202:
```suggestion
    if (typeof obj === "object") {
```

test/unit/classes/instanceof.spec.ts
```diff
@@ -49,6 +49,28 @@ test("null instanceof Class", () => {
     `.expectToMatchJsResult();
 });
 
+test("function instance of Class", () => {
```
ark120202:
```suggestion
test("function instanceof Class", () => {
```

Same applies to other new tests

test/unit/classes/instanceof.spec.ts
```diff
@@ -49,6 +49,28 @@ test("null instanceof Class", () => {
...
+    util.testFunction`
+        class myClass {}
+        const noop = () => {}
```
ark120202:
```suggestion
        const noop = () => {};
```

PR #839, opened 2020-03-14
Switch to ESLint

ark120202:


src/LuaPrinter.ts
```diff
@@ -31,7 +31,7 @@ export const escapeString = (value: string) => `"${value.replace(escapeStringReg
  * `foo.bar` => passes (`function foo.bar()` is valid)
  * `getFoo().bar` => fails (`function getFoo().bar()` would be illegal)
  */
-const isValidLuaFunctionDeclarationName = (str: string) => /^[a-zA-Z0-9_\.]+$/.test(str);
+const isValidLuaFunctionDeclarationName = (str: string) => /^[a-zA-Z0-9_.]+$/.test(str);
```
Perryvw: Is this change correct? Seems unintended

src/LuaPrinter.ts
```diff
@@ -200,7 +200,7 @@ export class LuaPrinter {
             (luaLibImport === LuaLibImportKind.Require && luaLibFeatures.size > 0)
         ) {
             // Require lualib bundle
-            header += `require("lualib_bundle");\n`;
+            header += 'require("lualib_bundle");\n';
```
Perryvw: If we are disallowing `` ` `` strings, we should at least be consistent, this string was changed to ' while the one on 194 was changed to "?

src/lualib/ArrayReduceRight.ts
```diff
@@ -3,7 +3,7 @@ function __TS__ArrayReduceRight<T>(
     this: void,
     arr: T[],
     callbackFn: (accumulator: T, currentValue: T, index: number, array: T[]) => T,
-    ...initial: Vararg<T>
+    ...initial: Vararg<T[]>
```
Perryvw: This does not seem correct, VarArg<T> extends Array<T>, so seems like initial is T[][] here?

src/lualib/ArraySplice.ts
```diff
@@ -1,5 +1,5 @@
 // https://www.ecma-international.org/ecma-262/9.0/index.html#sec-array.prototype.splice
-function __TS__ArraySplice<T>(this: void, list: T[], ...args: Vararg<unknown>): T[] {
+function __TS__ArraySplice<T>(this: void, list: T[], ...args: Vararg<unknown[]>): T[] {
```
Perryvw: Same as comment on reduce right

src/lualib/InstanceOf.ts
```diff
@@ -1,10 +1,10 @@
 function __TS__InstanceOf(this: void, obj: LuaClassInstance, classTbl: LuaClass): boolean {
     if (typeof classTbl !== "object") {
-        // tslint:disable-next-line: no-string-throw
         throw "Right-hand side of 'instanceof' is not an object";
     }
 
     if (classTbl[Symbol.hasInstance] !== undefined) {
+        // eslint-disable-next-line no-implicit-coercion
         return !!classTbl[Symbol.hasInstance](obj);
```
Perryvw: wouldnt it be nicer to just change this to an actual comparison

src/lualib/Map.ts
```diff
@@ -130,7 +128,7 @@ Map = class Map<K, V> {
     }
 
     public keys(): IterableIterator<K> {
-        const nextKey = this.nextKey;
+        const { nextKey } = this;
```
Perryvw: Does eslint force this? I am not sure about this rule actually.

src/lualib/Symbol.ts
```diff
@@ -1,11 +1,6 @@
-// tslint:disable-next-line: variable-name
 const ____symbolMetatable = {
-    __tostring(): string {
-        if (this.description === undefined) {
-            return "Symbol()";
-        } else {
-            return "Symbol(" + this.description + ")";
-        }
+    __tostring(this: symbol): string {
+        return `Symbol(${this.description || ""})`;
```
Perryvw: Wouldn't `??` be more correct here?

src/LuaPrinter.ts
ark120202: It was reported by [no-useless-escape](https://eslint.org/docs/rules/no-useless-escape) - the escape there is useless, because in square brackets all special characters (except closing bracket) don't have to be escaped

src/LuaPrinter.ts
ark120202: Prettier uses non-default quotes type when the string contains default one, to avoid using escapes

> If we are disallowing ` strings

I'm not really sure about it myself, but I think being more consistent is better. If anything, I wouldn't mind switching to template strings everywhere ([quotes](https://eslint.org/docs/rules/quotes) has an option for that)

src/lualib/ArrayReduceRight.ts
ark120202: I have changed `Vararg` declaration (https://github.com/TypeScriptToLua/TypeScriptToLua/pull/839/files#diff-a325139454ae742b51dd67eea598d304R5), since it was triggering [no-empty-interface](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-empty-interface.md). This definition of `Vararg` is probably going to be added to builtin types anyway

src/lualib/InstanceOf.ts
ark120202: A comparison with `true`? I'm fine with doing that for now, but in the future (when we'll implement boolean coercion algorithm) I think it should use `Boolean(...)`, since TypeScript doesn't check implementations of well-known symbols.

src/lualib/Symbol.ts
ark120202: We don't support nullish coalescing yet, but in general it doesn't matter since the only possible falsy value there is an empty string

src/lualib/Map.ts
ark120202: That's a [`prefer-destructuring`](https://eslint.org/docs/rules/prefer-destructuring) rule. I don't really mind disabling it.

src/transformation/visitors/loops/for-of.ts
```diff
@@ -88,6 +88,7 @@ function transformForOfLuaIteratorStatement(
             // Variables NOT declared in for loop - catch iterator values in temps and assign
             // for ____value0 in ${iterable} do
             //     ${initializer} = ____value0
+            // eslint-disable-next-line no-lonely-if
```
Perryvw, line 91: This rule seems a little awkward. I think this is very much a case-by-case basis and would prefer not to have a few eslint-disables as possible.

src/typescript-internal.ts
```diff
@@ -1,3 +1,5 @@
+// TODO:
+/* eslint-disable @typescript-eslint/no-unnecessary-qualifier */
```
Perryvw, line 2: What is the action TODO here?

src/transformation/visitors/loops/for-of.ts
ark120202: I guess that's true, though I don't think I had any issues with it in years of being in my config. I have disabled this rule and also fixed this case, since it actually turned out to look neater with the logic flow here

Issue #846, opened 2020-03-29
Keyword escaping on class methods does not match escaping on method calls


Perryvw:
See [Playground](https://typescripttolua.github.io/play#code/MYGwhgzhAECyCeBhcVoG8BQ1vQA4FcAjEAS2GgFMAnKgeyoAoBKdaKigF3yoDtoAiAKI16AQn4BuaAF8MsjAtxUSPDgx4UA7nCQoIzAHTU6jJkwlA)

Either the method call should not be escaped, or the class member should be escaped, current mismatch causes a Lua runtime error.

PR #847, opened 2020-03-30
Fixed incorrect escaping of fields

### LINKED ISSUE #833: table properties using builtin names are incorrectly marked as unsafe and renamed with underscores
Using identifiers in the builtins list as property names is valid Lua:
```lua
local Vector = {}
Vector.__index = Vector
function Vector.new(x,y)
    return setmetatable({
        x = x,
        y = y
    }, Vector)
end

function Vector:unpack()
    return self.x, self.y
end

local v = Vector.new(2,3)
local x,y = v:unpack()
print(x,y)
```
If I were to write a declaration for this:
```typescript
interface Vector {
    new: (this: void, x: number, y: number) => Vector;
    /** @tupleReturn **/
    unpack(): [number, number];
}
let vec2 = {} as Vector;

let v = vec2.new(2,3);
let [x,y] = v.unpack();
let [a,b] = v["unpack"]();
```
which compiles to:
```lua
vec2 = {}
v = vec2.new(2, 3)
x, y = v:____unpack()
a, b = v.unpack(v)
```

___unpack() will cause an error. Accessing it via string index is currently a workaround.

I believe this is related to #789

Perryvw:
Fixes #833, #789, #846

### DIFF
--- a/src/transformation/visitors/call.ts
+++ b/src/transformation/visitors/call.ts
@@ -10,7 +10,6 @@ import { LuaLibFeature, transformLuaLibFunction } from "../utils/lualib";
 import { isValidLuaIdentifier } from "../utils/safe-names";
 import { isArrayType, isExpressionWithEvaluationEffect, isInDestructingAssignment } from "../utils/typescript";
 import { transformElementAccessArgument } from "./access";
-import { transformIdentifier } from "./identifier";
 import { transformLuaTableCallExpression } from "./lua-table";
 
 export type PropertyCallExpression = ts.CallExpression & { expression: ts.PropertyAccessExpression };
@@ -117,7 +116,7 @@ export function transformContextualCallExpression(
 
         return lua.createMethodCallExpression(
             table,
-            transformIdentifier(context, left.name),
+            lua.createIdentifier(left.name.text, left.name),
             transformedArguments,
             node
         );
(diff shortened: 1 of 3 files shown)


src/transformation/visitors/call.ts
```diff
@@ -117,7 +116,7 @@ export function transformContextualCallExpression(
 
         return lua.createMethodCallExpression(
             table,
-            transformIdentifier(context, left.name),
+            lua.createIdentifier(left.name.text, left.name),
```
ark120202: Would that work for syntactically invalid identifiers?

src/transformation/visitors/class/members/constructor.ts
```diff
@@ -66,13 +66,15 @@ export function transformConstructorDeclaration(
...
-        const assignment = lua.createAssignmentStatement(
-            lua.createTableIndexExpression(createSelfIdentifier(), lua.createStringLiteral(declarationName.text)),
-            declarationName
-        );
-        bodyWithFieldInitializers.push(assignment);
+        if (ts.isIdentifier(declaration.name)) {
+            // self.declarationName = declarationName
+            const assignment = lua.createAssignmentStatement(
+                lua.createTableIndexExpression(createSelfIdentifier(), lua.createStringLiteral(declaration.name.text)),
+                transformIdentifier(context, declaration.name as ts.Identifier)
```
ark120202:
```suggestion
                transformIdentifier(context, declaration.name)
```

PR #859, opened 2020-04-29
fixed paths in source maps

### LINKED ISSUE #855: Incorrect paths in source maps in subfolders
When a ts file in a subfolder is transpiled with a source map and output to another folder (using `outDir`), the source map does not contain the correct relative path information to find the original source.

Example:

*tsconfig.json*
```json
{
    "compilerOptions": {
        "sourceMap": true,
        "rootDir": "ts",
        "outDir": "lua"
    },
    "include": [
        "ts/sub/foobar.ts"
    ]
}
```

*lua/sub/foobar.lua.map*
```
{"version":3,"sources":["foobar.ts"],"names":[],"mappings":";MAAY","file":"foobar.lua","sourceRoot":"..\\ts"}
```

In this example, the map would point to `lua/ts/foobar.ts` instead of `ts/sub/foobar.ts`.

The obvious thing to do would be to fix sourceRoot in these situations:
```
{"version":3,"sources":["foobar.ts"],"names":[],"mappings":";MAAY","file":"foobar.lua","sourceRoot":"..\\..\\ts\\sub"}
```

But, looking at typescript's output for the same project, it seems they opt to include the path information in `sources`:
```
{"version":3,"file":"foobar.js","sourceRoot":"","sources":["../../ts/sub/foobar.ts"],"names":[],"mappings":"AAAA,OAAO,CAAC,GAAG,CAAC,QAAQ,CAAC,CAAC"}
```

TS only uses `sourceRoot` when it's explicitly declared in `tsconfig.json`:
```
{
    "compilerOptions": {
        ...
        "sourceRoot": "ts"
    },
    ...
}
```
```
{"version":3,"file":"foobar.js","sourceRoot":"ts/","sources":["sub/foobar.ts"],"names":[],"mappings":"AAAA,OAAO,CAAC,GAAG,CAAC,QAAQ,CAAC,CAAC"}
```
Perhaps emulating that would be the ideal solution.

tomblind:
- 'sources' now contain full relative paths to original ts file
- 'sourceRoot' is only set if specified explicitly from tsconfig.json

fixes #855

### DIFF
--- a/src/LuaPrinter.ts
+++ b/src/LuaPrinter.ts
@@ -129,7 +129,27 @@ export class LuaPrinter {
     private sourceFile: string;
 
     public constructor(private options: CompilerOptions, private emitHost: EmitHost, fileName: string) {
-        this.sourceFile = path.basename(fileName);
+        // Determine file name used in source map
+        const fileNameOnly = path.basename(fileName);
+        if (options.outDir) {
+            const fileDir = path.dirname(fileName);
+            const relativeFileDir = path.relative(this.options.rootDir || process.cwd(), fileDir);
+            if (options.sourceRoot) {
+                // When sourceRoot is specified, just use relative path inside rootDir
+                this.sourceFile = path.join(relativeFileDir, fileNameOnly);
+            } else {
+                // Calculate relative path from rootDir to outDir
+                const absoluteOutDir = path.resolve(options.outDir, relativeFileDir);
+                const relativeOutDir = path.relative(absoluteOutDir, fileDir);
+                this.sourceFile = path.join(relativeOutDir, fileNameOnly);
+            }
+            if (path.sep !== "/") {
+                // We want forward slashes, even in windows
+                this.sourceFile = this.sourceFile.replace(new RegExp(`\\${path.sep}`, "g"), "/");
+            }
+        } else {
+            this.sourceFile = fileNameOnly; // File will be in same dir as source
+        }
     }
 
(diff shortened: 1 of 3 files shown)


src/LuaPrinter.ts
```diff
@@ -129,7 +129,27 @@ export class LuaPrinter {
...
+            const relativeFileDir = path.relative(this.options.rootDir || process.cwd(), fileDir);
+            if (options.sourceRoot) {
+                // When sourceRoot is specified, just use relative path inside rootDir
+                this.sourceFile = path.join(relativeFileDir, fileNameOnly);
+            } else {
+                // Calculate relative path from rootDir to outDir
+                const absoluteOutDir = path.resolve(options.outDir, relativeFileDir);
+                const relativeOutDir = path.relative(absoluteOutDir, fileDir);
+                this.sourceFile = path.join(relativeOutDir, fileNameOnly);
+            }
```
ark120202, line 145: I haven't actually tested that code, but can't it be simplified into something like that:
```suggestion
            const relativeFileName = path.relative(program.getCommonSourceDirectory(), fileName);
            if (options.sourceRoot) {
                // When sourceRoot is specified, just use relative path inside rootDir
                this.sourceFile = relativeFileName;
            } else {
                // Calculate relative path from rootDir to outDir
                const outputPath = path.resolve(options.outDir, relativeFileName);
                this.sourceFile = path.relative(path.dirname(outputPath), fileName);
            }
```

PR #861, opened 2020-05-02
Replaced error throw with diagnostic

Perryvw:


src/transformation/context/context.ts
```diff
@@ -78,7 +84,12 @@ export class TransformationContext {
     }
 
     public transformExpression(node: ExpressionLikeNode): lua.Expression {
```
ark120202, line 86: Same thing should be done for `superTransformExpression`

PR #864, opened 2020-05-05
Added memory benchmarks to CI

lolleko:
For a preview of the results see: https://github.com/TypeScriptToLua/TypeScriptToLua/runs/653550386

**From benchmark/README.md:**
## TSTL Benchmarks

These benchmarks are written in typescript and transpiled to lua by using tstl.

### Currently only memory benchmarks are supported

To add a new benchmark add a new file to `memory_benchmarks`
and **default** export a function with the following type: `() => void`.

For example (memory_benchmarks/my_benchmark.ts):

```ts
export default myBenchmark() {
    cont n = 123;
    for (let i = 0; i < n; i++) {
        // Do something memory instensive
    }
}
```

**Goal**

The goal of memory benchmarks is to track how much (memory) `"garbage"` is created by tstl.
For that reason garabage collection is disabled in the benchmarks.

You can force the creation of `"garbage"` by creating a lot of anonymous functions or temporary tables (see [lua-users.org](http://lua-users.org/wiki/OptimisingGarbageCollection) for more information).

To avoid crashes in the CI your benchmark should not use more than 500MB of memory.

**Running locally**

1. Create a benchmark baseline called "benchmark_baseilne.json":  
   `tstl -p tsconfig.53.json && cd dist && lua -- run.lua benchmark_baseilne.json`
2. Make some changes to tstl.
3. Create an updated benchmark and compare with the baseline:  
   `tstl -p tsconfig.53.json && cd dist && lua -- run.lua benchmark_updated.json benchmark_baseilne.json`
4. The above command will output comparison data as json to stdout.
   If you provide a path as third argument the comparison data will be written to that path instead.  
   `tstl -p tsconfig.53.json && cd dist && lua -- run.lua benchmark_updated.json benchmark_baseilne.json result.md`


benchmark/src/util.ts
```diff
@@ -0,0 +1,85 @@
...
+        content = file.read("a");
+    } else {
+        // JIT
+        // @ts-ignore
+        content = file.read("*a");
+    }
+
+    if (content && content[0]) {
+        return content[0];
+    }
```
ark120202, line 39: Isn't content going to be always defined, since it's a tuple?
```suggestion
    return content[0];
```

benchmark/src/util.ts
```diff
@@ -0,0 +1,85 @@
...
+                // on windows we need to append the directory path
+                // on unix this is done by find automatically
+                files = files.map(f => `${dir}/${f}`);
+            }
+            return findResult.split("\n").filter(p => p !== "");
+        }
+    }
+}
+
+export function loadBenchmarksFromDirectory(dir = ""): BenchmarkFunction[] {
```
ark120202, line 70: IMO in this case (and some others) the default isn't very useful

benchmark/src/util.ts
```diff
@@ -0,0 +1,85 @@
...
+    // Memory tests
+    const benchmarkPaths = readDir(dir);
+
+    if (!benchmarkPaths) {
```
ark120202, line 74: Why would a `readDir` function ever return undefined? Wouldn't it be better to throw inside `readDir`, so it's clear that something is wrong, instead of just assuming that there are no benchmarks?

benchmark/tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
+    "compilerOptions": {
+        "target": "esnext",
+        "lib": ["esnext"],
+        // Dev types are JIT
+        "types": ["lua-types/jit"],
+        "moduleResolution": "node",
+        "strict": true,
+        "outDir": "dist",
+        "rootDir": "src"
```
ark120202, line 10: To make it closer to https://github.com/TypeScriptToLua/TypeScriptToLua/blob/master/tsconfig.json
```suggestion
        "target": "esnext",
        "lib": ["esnext"],
        // Dev types are JIT
        "types": ["lua-types/jit"],
        "moduleResolution": "node",

        "outDir": "dist",
        "rootDir": "src",

        "strict": true,
        "noUnusedLocals": true,
        "noUnusedParameters": true
```

benchmark/src/run.ts
```diff
@@ -0,0 +1,50 @@
...
+        comparisonInfo = { summary: memoryComparisonInfo[0], text: memoryComparisonInfo[1] };
+    }
+
+    // Output comparison info
+    if (!arg[2]) {
+        // Output to stdout as json by default, this is used by the CI to retrieve the info
+        print(json.encode(comparisonInfo));
+    } else {
+        // Output to file as markdown if arg[2] is set, this is useful for local development
+        const updatedResultsFile = io.open(arg[2], "w+")[0] as LuaFile;
```
ark120202, line 50:
```suggestion
        const updatedResultsFile = io.open(arg[2], "w+")[0]!;
```

benchmark/src/benchmark_types.ts
```diff
@@ -0,0 +1,22 @@
...
+    benchmarkName: string;
+    preExecMemoryUsage: number;
+    postExecMemoryUsage: number;
+    memoryUsedForExec: number;
+    memoryAfterGC: number;
+}
+
+export function isMemoryBenchmarkResult(result: BenchmarkResult): result is MemoryBenchmarkResult {
+    return result.kind == BenchmarkKind.Memory;
+}
```
ark120202, line 22: Sum types are generally easier to use than a common interface with manual type guards

```suggestion
export type BenchmarkFunction = () => void;

export enum BenchmarkKind {
    Memory = "memory",
}

type BenchmarkResult = MemoryBenchmarkResult;

export interface MemoryBenchmarkResult {
    kind: BenchmarkKind.Memory;
    benchmarkName: string;
    preExecMemoryUsage: number;
    postExecMemoryUsage: number;
    memoryUsedForExec: number;
    memoryAfterGC: number;
}
```

benchmark/src/util.ts
```diff
@@ -0,0 +1,85 @@
...
+        fileHandle.close();
+
+        if (fileContent) {
+            return fileContent;
+        }
+    }
+}
+
+export function readAll(file: LuaFile): string | undefined {
+    let content: (string | undefined)[];
```
ark120202, line 27: TSLint doesn't appear to be checking these files. We could merge #839 first, or you could change the linting script to include new project

```diff
-"lint:tslint": "tslint -p . && tslint -p test && tslint -p src/lualib",
+"lint:tslint": "tslint -p . && tslint -p test && tslint -p benchmark && tslint -p src/lualib",
```

benchmark/src/util.ts
```diff
@@ -0,0 +1,85 @@
...
+    let isWindows = false;
+    let [success, findHandle] = pcall(() => io.popen(`find '${dir}' -maxdepth 1 -type f | sed "s|^\./||"`));
+
+    if (!success) {
+        [success, findHandle] = pcall(() => io.popen(`dir /A-D /B ${dir}`));
+        isWindows = true;
+    }
+
+    if (success) {
+        // appereantly TS can't infer this on it's own
```
ark120202, line 52: I doubt TS would ever start tracking dependencies between unrelated variables. Instead of casting we might change code a bit to make it inferrable https://typescripttolua.github.io/play#code/KYDwDg9gTgLgBAMwK4DsDGMCWEVysAQwBMARTKACiPLgF44AiBgSgC44BnGKTFAcwDaAXTgAfOKiLAEvYETgBvAFBxVcADbB4mDgHVeRCAHcOdRAXUdgAbhVrN8SGGAoASsA5J18emDQX1CgpmOgA+OGwAOicXCgADGRR5AHIAEgVqKABfZLgAWgBbAhApMBgACzgARnyYAE9nRDFOOUYOUQA9AB1IgHpRUQY45mZbO1VMBDgKAEIYtw8vGAEABiEQ5TUtuHn3T28zPwCgkNpwqPn4zLhegEE8khuAITh0zKzh0fHtnX0k41M9G4SBs3yySm+k2mu0W3lW60U3y2vV6cAIYGc+AIKBg6jqcAAKgBlOD+FDJbQoBDAKBwCo6OA4CIwZKmYwoJFqNA4LiIAwACWxRE0hwgzgW+2WVSEY22XJ58ESRD2SzMWKIt3UgSVgqSmi+ctUOqFmkiaHUECswVlhqhFCVKu8G05cocfM0gL5SUdMEiHDA6kwMAoDC6KBYtkNPymFF+BgBzqjUZRjNwRnjJjgRmAcBQwFaMAgaIxLnkFRzmWAGGg+LABAqLqTKaZqEwIDp5QZDMMebgACN8Uq0UhC0UsP4tXVG1GZB6zLOPJEimB7WE4HE3uQsr10ggPgak6pwYfVPgYEgoLgHbDff7A8HQ+HmJFZzAaRQwGvPzNaPQmAe5WPNRwSyIA

benchmark/src/util.ts
```diff
@@ -0,0 +1,85 @@
...
+        findHandle = findHandle as LuaFile;
+
+        const findResult = readAll(findHandle);
+        findHandle.close();
+
+        if (findResult) {
+            let files = findResult.split("\n");
+            if (isWindows) {
```
ark120202, line 60: Maybe instead we could make `find` return just file names with `-printf "%f\n"`, so we can have a single code path for both platforms

benchmark/src/util.ts
```diff
@@ -0,0 +1,85 @@
...
+                // on windows we need to append the directory path
+                // on unix this is done by find automatically
+                files = files.map(f => `${dir}/${f}`);
+            }
+            return findResult.split("\n").filter(p => p !== "");
```
ark120202, line 65: `files` isn't used in returned expression

benchmark/src/util.ts
lolleko: Right. I Improved error handling, reader related errors are not correctly reported to console. Also fixed some popen issues.

benchmark/src/util.ts
lolleko: I ran tslint local for /benchmark and fixed all issues but I did not change the config to avoid conflicts with #839.

benchmark/src/util.ts
lolleko: `-printf` is a rather new unix extension to `find` and not part of the posix standard. So it won't work on BSD systems (macOS) and older unix systems.

benchmark/src/util.ts
```diff
@@ -0,0 +1,107 @@
+import { BenchmarkFunction } from "./benchmark_types";
+
+export type Result<T, E> = Ok<T, E> | Err<T, E>;
+
+export class Ok<T, E> {
```
Perryvw, line 5: This seems a little heavy and awkward, what I usually do is
```ts
type Result<T> = { success: true, value: T } | { success: false, error: string };
```
And then use TypeScript's type system to do the type narrowing
```ts
if (result.success) {
    // result.value available
} else {
    // result.error available
}

Comment (ark120202) [reply benchmark/src/util.ts:5]: IMO monads in non-functional languages look kind of weird in general. Why not just throw a regular error? It's effect of jumping through the stack is useful in the majority of cases, especially there, where the only thing we could do on error is die.

File: benchmark/README.md [hunk 12]
```diff
@@ -0,0 +1,39 @@
...
+
+### Currently only memory benchmarks are supported
+
+To add a new benchmark add a new file to `memory_benchmarks`
+and **default** export a function with the following type: `() => void`.
+
+For example (memory_benchmarks/my_benchmark.ts):
+
+```ts
+export default myBenchmark() {
```
ark120202, line 13:
```suggestion
export default function myBenchmark() {
```

benchmark/src/memory_benchmarks/array_concat.ts
```diff
@@ -0,0 +1,10 @@
+// tslint:disable-next-line:no-default-export
```
ark120202, line 1: We could just disable `no-default-export` for these directories

benchmark/src/util.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    if (!readBenchmarkDirResult.success) {
+        return { success: false, error: readBenchmarkDirResult.error };
+    }
+
+    return {
+        success: true,
+        value: readBenchmarkDirResult.value.map(f => {
+            // replace slashes with dots
+            let dotPath = string.gsub(f, "%/", ".")[0];
```
ark120202, line 78: Do we have to do this?

benchmark/src/util.ts
```diff
@@ -0,0 +1,84 @@
...
+    }
+    return { success: false, error: `Can't readAll for file ${file}` };
+}
+
+export function readDir(dir: string): Result<string[]> {
+    const findHandle = isWindows ? io.popen(`dir /A-D /B ${dir}`) : io.popen(`find '${dir}' -maxdepth 1 -type f`);
+    const findReadAllResult = readAll(findHandle);
+
+    if (!findHandle.close()) {
+        return { success: false, error: `readDir popen failed for dir ${dir} see stdout for more information.` };
```
ark120202, line 48: What's the point of using non-standard to the platform monads, when then only thing we can do is propagate the error down on the call stack and die in the end?

benchmark/src/util.ts
```diff
@@ -0,0 +1,71 @@
...
+    const fileOpenArray = io.open(path, "rb");
+
+    if (fileOpenArray?.[0]) {
+        const fileHandle = fileOpenArray[0];
+        const fileContent = readAll(fileHandle);
+        fileHandle.close();
+
+        return fileContent;
+    }
+    throw Error(`Can't open file ${path}`);
```
ark120202, line 29: The optional chaining in `fileOpenArray?.[0]` would just get ignored since we don't support it now. Also, `io.open` appears to return a tuple, which never can be null. Since code flow analysis isn't require there, we can just destructure it in-place
```suggestion
    const [fileHandle] = io.open(path, "rb");
    if (!fileHandle) {
        throw Error(`Can't open file ${path}`);
    }

    const fileContent = readAll(fileHandle);
    fileHandle.close();
    return fileContent;
```

benchmark/src/util.ts
```diff
@@ -0,0 +1,71 @@
...
+    const content = file.read(_VERSION === "Lua 5.3" ? "a" : ("*a" as any)) as [string | undefined];
+
+    if (content[0]) {
+        return content[0];
+    }
+    throw Error(`Can't readAll for file ${file}`);
+}
+
+export function readDir(dir: string): string[] {
+    const findHandle = isWindows ? io.popen(`dir /A-D /B ${dir}`) : io.popen(`find '${dir}' -maxdepth 1 -type f`);
```
ark120202, line 42:
```suggestion
    const findHandle = io.popen(isWindows ? `dir /A-D /B ${dir}` : `find '${dir}' -maxdepth 1 -type f`);
```

benchmark/src/util.ts
```diff
@@ -0,0 +1,71 @@
...
+export function toFixed(num: number, decimalPlaces = 0): string {
+    return string.format(`%.${decimalPlaces}f`, num);
+}
+
+export function calculatePercentageChange(oldValue: number, newValue: number): number {
+    return (newValue / oldValue) * 100 - 100;
+}
+
+// @ts-ignore
+export const isWindows = package.config.sub(1, 1) === "\\";
```
ark120202, line 12:
```suggestion
export const isWindows = package.config.startsWith("\\");
```

benchmark/src/util.ts
lolleko: The linter forced me :(

benchmark/README.md
```diff
@@ -0,0 +1,43 @@
...
+The goal of memory benchmarks is to track how much (memory) `"garbage"` is created by tstl.
+For that reason garabage collection is disabled in the benchmarks.
+
+You can force the creation of `"garbage"` by creating a lot of anonymous functions or temporary tables (see [lua-users.org](http://lua-users.org/wiki/OptimisingGarbageCollection) for more information).
+
+To avoid crashes in the CI your benchmark should not use more than 500MB of memory.
+
+**Running locally**
+
+1. Create a benchmark baseline called "benchmark_baseilne.json":
```
Perryvw, line 36:
```suggestion
1. Create a benchmark baseline called "benchmark_baseline.json":  
```

Perryvw: Actually there are a lot of these copy pasted here.

PR #868, opened 2020-05-09
Fix side effects in string index causing issues

### LINKED ISSUE #867: Indexing on string fails when index is an expression with side effects
Indexing on strings (e.g., `myString[7]`) duplicates the passed index expression, which results in incorrect code when the index expression has side effects, such as incrementing a value (`myString[i++]`).

Example, which would produce the string `"01"` instead of the expected `"0"`:
https://typescripttolua.github.io/play/#code/FAGwpgLgBAllC8UAMBuUkoGcICcFQCIkBGAJgLWGxwG0YBqegXTSA

Perryvw:
Fixes #867

### DIFF
--- a/src/transformation/visitors/access.ts
+++ b/src/transformation/visitors/access.ts
@@ -4,7 +4,7 @@ import { transformBuiltinPropertyAccessExpression } from "../builtins";
 import { FunctionVisitor, TransformationContext } from "../context";
 import { AnnotationKind, getTypeAnnotations } from "../utils/annotations";
 import { createExpressionPlusOne } from "../utils/lua-ast";
-import { isArrayType, isNumberType, isStringType } from "../utils/typescript";
+import { isArrayType, isNumberType, isStringType, isExpressionWithEvaluationEffect } from "../utils/typescript";
 import { tryGetConstEnumValue } from "./enum";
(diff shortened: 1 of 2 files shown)


src/transformation/visitors/access.ts
```diff
@@ -31,19 +31,13 @@ export const transformElementAccessExpression: FunctionVisitor<ts.ElementAccessE
...
-        return lua.createCallExpression(
-            lua.createTableIndexExpression(lua.createIdentifier("string"), lua.createStringLiteral("sub")),
-            [table, createExpressionPlusOne(index), createExpressionPlusOne(index)],
-            expression
-        );
+        return transformStringIndex(context, expression);
     }
 
+    const table = context.transformExpression(expression.expression);
     return lua.createTableIndexExpression(table, transformElementAccessArgument(context, expression), expression);
```
ark120202, line 41: That `table` variable name might be a bit confusing, since you don't necessarily access a property on a table. And in general the variable in this case doesn't seem to be very helpful here
```suggestion
    return lua.createTableIndexExpression(
        context.transformExpression(expression.expression),
        transformElementAccessArgument(context, expression),
        expression
     );
```

PR #873, opened 2020-05-16
Add `luaTarget: "universal"`

### LINKED ISSUE #749: Add `luaTarget: "universal"`
It would be useful for libraries supporting all lua versions (lualib is a good example). It should enable all possible limitations (i.e. no switch) and import all differently-named standard functions from lualib (`__TS__Unpack = table.unpack or unpack`).

We should also consider making it a default.

ark120202:
Resolves #749


src/transformation/visitors/binary-expression/bit.ts
```diff
@@ -105,8 +106,9 @@ export function transformUnaryBitOperation(
     operator: lua.UnaryBitwiseOperator
 ): lua.Expression {
     switch (context.luaTarget) {
+        case LuaTarget.Universal:
         case LuaTarget.Lua51:
-            context.diagnostics.push(unsupportedForTarget(node, "Bitwise operations", LuaTarget.Lua51));
+            context.diagnostics.push(unsupportedForTarget(node, "Bitwise operations", context.luaTarget));
```
Perryvw, line 111: For the other messages you left the version mentioned in the diagnostic to be 5.1 (which also makes sense). We should probably be consistent with this though.

test/unit/builtins/math.spec.ts
```diff
@@ -27,8 +27,11 @@ test.each(["E", "LN10", "LN2", "LOG10E", "LOG2E", "SQRT1_2", "SQRT2"])("Math.%s"
 
 const expectMathAtan2: util.TapCallback = builder => expect(builder.getMainLuaCodeChunk()).toContain("math.atan2(");
 const expectMathAtan: util.TapCallback = builder => expect(builder.getMainLuaCodeChunk()).toContain("math.atan(");
+const expectTsMathAtan2: util.TapCallback = builder =>
```
Perryvw, line 30: Maybe it would be nicer to call this `expectLualibMathAtan2`.

test/unit/spread.spec.ts
```diff
@@ -5,6 +5,7 @@ import { formatCode } from "../util";
 // TODO: Make some utils for testing other targets
 const expectUnpack: util.TapCallback = builder => expect(builder.getMainLuaCodeChunk()).toMatch(/[^.]unpack\(/);
 const expectTableUnpack: util.TapCallback = builder => expect(builder.getMainLuaCodeChunk()).toContain("table.unpack");
+const expectTsUnpack: util.TapCallback = builder => expect(builder.getMainLuaCodeChunk()).toContain("__TS__Unpack");
```
Perryvw, line 8: Since we changed `expectTsMathAtan2` to `expectLualibMathAtan2`, I would suggest also changing this one to `expectLualibUnpack`

PR #882, opened 2020-05-22
Fix class .toString inheritance

### LINKED ISSUE #728: Inheritance breaks toString()
## Code

```ts
class A {
    public toString() {
        return "A";
    }
}

class B extends A { }

console.log(new B());
```

## Expected output

```
A
```

## Current output

```
table: 0x2d
```

ark120202:
Fixes #728


test/unit/classes/classes.spec.ts
```diff
@@ -352,15 +352,29 @@ test("ClassComputedMethodCall", () => {
...
+
+test(".toString() with inheritance", () => {
+    util.testFunction`
+        class A {
+            public toString() {
+                return "A";
+            }
+        }
+
+        class B extends A {}
```
Perryvw, line 375: Do we have a test where B overrides in herited toString? Do we need one?

PR #887, opened 2020-05-23
Fix `array.join` accepting only string and number elements

### LINKED ISSUE #841: `Array.prototype.join` doesn't work on tables
[Playground](https://typescripttolua.github.io/play#code/A4JwlgdgLgFA2gbwL4F0B0ArA9pGBKPIA)

**Input:**
```ts
print([{}].join())
```

**Current Result:**
```lua
print(
    table.concat({{}}, ",") -- invalid value (table) at index 1 in table for 'concat'
)
```

**Expected Result:**
```lua
function __TS__ArrayJoin(self, separator)
	if separator == nil then separator = ", " end

	local result = ""
	for index, value in ipairs(self) do
		if index > 1 then
			result = result + separator
		end
		result = result + tostring(value)
	end
	return result
end

print(
    __TS__ArrayJoin({{}}, ",") -- "table: 0x25"
)
```

ark120202:
Fixes #841

### DIFF
--- a/src/lualib/Iterator.ts
+++ b/src/lualib/Iterator.ts
@@ -27,6 +27,6 @@ function __TS__Iterator<T>(
         const iterator = iterable[Symbol.iterator]();
         return [__TS__IteratorIteratorStep, iterator];
     } else {
-        return ipairs(iterable as readonly T[]);
+        return ipairs(iterable as readonly T[]) as any;
     }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,8 @@
 
 - Fixed iteration over generators stopping at first yielded `nil` value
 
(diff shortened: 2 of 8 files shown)


src/lualib/Iterator.ts
```diff
@@ -27,6 +27,6 @@ function __TS__Iterator<T>(
         const iterator = iterable[Symbol.iterator]();
         return [__TS__IteratorIteratorStep, iterator];
     } else {
-        return ipairs(iterable as readonly T[]);
+        return ipairs(iterable as readonly T[]) as any;
```
Perryvw, line 30: What is the reasoning behind this `as any?`

ark120202: `ipairs` returns a `LuaTupleIterator` which is incompatible with `__TS__Iterator` signature. Generally this should be resolved with 880

PR #891, opened 2020-06-01
Make strings iterable

### LINKED ISSUE #843: Strings cannot be iterated
[Playground](https://typescripttolua.github.io/play/#code/GYewTgBAFAxiB2BnALhGALAhpEwIHJQR8BKCAbwFgAoCOiABzAEt5lYswSBuGgXxpA)

**Input:**
```ts
for (const char of 'foo') {
    print(char);
}
```

**Current Result:**
```lua
function __TS__Iterator(iterable)
    if iterable[Symbol.iterator] then
        local iterator = iterable[Symbol.iterator](iterable)
        return function()
            local result = iterator:next()
            if not result.done then
                return result.value
            else
                return nil
            end
        end
    else
        local i = 0
        return function()
            i = i + 1
            return iterable[i]
        end
    end
end

for char in __TS__Iterator("foo") do
    print(char)
end
```

**Expected Result:**
```lua
function __TS__Iterator(iterable)
    if iterable[Symbol.iterator] then
        local iterator = iterable[Symbol.iterator](iterable)
        return function()
            local result = iterator:next()
            if not result.done then
                return result.value
            else
                return nil
            end
        end
	elseif type(iterable) == "string" then
        local i = 0
        return function()
            i = i + 1
            return string.sub(iterable, i, i)
        end
    else
        local i = 0
        return function()
            i = i + 1
            return iterable[i]
        end
    end
end

for char in __TS__Iterator("foo") do
    print(char)
end
```

ark120202:
Resolves #843.

**Changelog:**
- Strings now can be iterated over

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Strings now can be iterated over
+
 ## 0.34.0
 
 - Added new `"luaTarget"` option value - `"universal"`. Choosing this target makes TypeScriptToLua generate code compatible with all supported Lua targets.
--- a/src/lualib/Iterator.ts
+++ b/src/lualib/Iterator.ts
@@ -17,15 +17,24 @@ function __TS__IteratorIteratorStep<T>(this: Iterator<T>): [true, T] | [] {
 }
 
 /** @tupleReturn */
+function __TS__IteratorStringStep(this: string, index: number): [number, string] | [] {
+    index += 1;
+    if (index > this.length) return [];
+    return [index, string.sub(this, index, index)];
+}
+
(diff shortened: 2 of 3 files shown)


Validark:
In roblox-ts, here is how we compile string iteration:
```ts
for (const letter of "Iñtërnâtiônàlizætiøn☃💩") {
	print(letter);
}
```
```lua
for letter in string.gmatch("Iñtërnâtiônàlizætiøn☃💩", "[%z\1-\127\194-\244][\128-\191]*") do
	print(letter);
end;
```

ark120202:
Thanks for suggestion, but I don't see much point in trying to properly support unicode just in one place. Consistently not supporting it sounds better than doing it randomly

PR #892, opened 2020-06-01
Fix incorrectly escaped `string.replace` replace value

### LINKED ISSUE #865: String replace error when using slash in String.replace
PG: https://typescripttolua.github.io/play#code/EQFwpgziD05cA6ATmADgGwIYGMwAphpgAaAAkWAEoBuIA

ark120202:
Fixes #865.

### DIFF
--- a/test/unit/builtins/string.spec.ts
+++ b/test/unit/builtins/string.spec.ts
@@ -56,11 +56,11 @@ test.each([
     { inp: "hello test", searchValue: "test", replaceValue: "" },
     { inp: "hello test", searchValue: "test", replaceValue: "world" },
     { inp: "hello test", searchValue: "test", replaceValue: "%world" },
+    { inp: "hello test", searchValue: " test", replaceValue: "." },
     { inp: "hello %test", searchValue: "test", replaceValue: "world" },
     { inp: "hello %test", searchValue: "%test", replaceValue: "world" },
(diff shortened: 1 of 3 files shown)


test/unit/builtins/string.spec.ts
```diff
@@ -56,11 +56,11 @@ test.each([
     { inp: "hello test", searchValue: "test", replaceValue: "" },
     { inp: "hello test", searchValue: "test", replaceValue: "world" },
     { inp: "hello test", searchValue: "test", replaceValue: "%world" },
+    { inp: "hello test", searchValue: " test", replaceValue: "." },
```
Perryvw, line 59: Could you also add some tests containing some of these other non-alphanumerics (both as search and replace values)?

test/unit/builtins/string.spec.ts
```diff
@@ -56,11 +56,11 @@ test.each([
...
     { inp: "hello %test", searchValue: "test", replaceValue: "world" },
     { inp: "hello %test", searchValue: "%test", replaceValue: "world" },
-    { inp: "hello test", searchValue: "test", replaceValue: (): string => "a" },
-    { inp: "hello test", searchValue: "test", replaceValue: (): string => "%a" },
-    { inp: "aaa", searchValue: "a", replaceValue: "b" },
```
Perryvw, line 63: Why did you remove this? I don't think there is another test that replaces multiple references in the same string.

PR #895, opened 2020-06-09
Fix some string operations

### LINKED ISSUE #814: `string.slice` negative end index is not shifted
Input:
```ts
console.log('abcd'.slice(1, -1)); // => "bc"
```

Current Result:
```lua
print(
    string.sub("abcd", 2, -1) -- => "bcd"
)
```

Expected Result:
```lua
print(
    string.sub("abcd", 2, -2) -- => "bc"
)
```

### LINKED ISSUE #890: Indexing string out of bounds is an empty string instead of undefined
[Playground](https://typescripttolua.github.io/play#code/A4JwlgdgLgFFCewCmB7AZgAgERYNoAYBdASgG4g)

Input:
```ts
print(typeof ""[0]);
```

Current Result:
```lua
print(
    __TS__TypeOf(
        string.sub("", 1, 1)
    )
)
```

Expected Result:
```lua
function __TS__StringIndex(self, index)
	if index <= #self then
		return string.sub("", index, index)
	end
end

print(
    __TS__TypeOf(
        __TS__StringIndex("", 1)
    )
)
```

ark120202:
Fixes #890, fixes #814.

**Changelog:**
- Fixed incorrect `NaN` and negative index handling of string `charAt`, `charCodeAt`, `slice`, `substr`, `substring`
- Fixed `String.prototype.substr` evaluating length expression twice
- Fixed indexing string out of bounds being an empty string instead of `undefined`


src/transformation/builtins/string.ts
```diff
@@ -46,35 +52,37 @@ export function transformStringPrototypeCall(
...
-                return createStringCall("sub", node, caller, arg1);
-            } else {
-                const arg1 = createExpressionPlusOne(params[0]);
-                const arg2 = params[1];
-                return createStringCall("sub", node, caller, arg1, arg2);
+            return transformLuaLibFunction(context, LuaLibFeature.StringSubstring, node, caller, ...params);
+
+        case "slice": {
+            const literalArg1 = getNumberLiteralValue(params[0]);
+            if (params[0] && literalArg1 !== undefined) {
```
Perryvw, line 64: Could you add a comment here explaining what this condition is and why it is added (also for `charAt` and `charCodeAt`). 
Something like `When slicing a string literal, translate to sliced Lua literal at compile-time to avoid lualib function call.`

src/transformation/utils/lua-ast.ts
```diff
@@ -22,24 +22,41 @@ export function createExportsIdentifier(): lua.Identifier {
     return lua.createIdentifier("____exports");
 }
 
-export function createExpressionPlusOne(expression: lua.Expression): lua.Expression {
-    if (lua.isNumericLiteral(expression)) {
-        const newNode = lua.cloneNode(expression);
-        newNode.value += 1;
+export const createExpressionPlusOne = (expression: lua.Expression) => modifyNumericExpression(expression, 1);
+export function modifyNumericExpression(expression: lua.Expression, change: number): lua.Expression {
```
Perryvw, line 26: `modifyNumericExpression` is a little generic since this only adds/subtracts. Suggestions: `addToNumericExpression` or `addLiteralToNumericExpression`.

src/transformation/utils/lua-ast.ts
```diff
@@ -22,24 +22,41 @@ export function createExportsIdentifier(): lua.Identifier {
...
+    return change > 0
+        ? lua.createBinaryExpression(expression, lua.createNumericLiteral(change), lua.SyntaxKind.AdditionOperator)
+        : lua.createBinaryExpression(expression, lua.createNumericLiteral(-change), lua.SyntaxKind.SubtractionOperator);
+}
+
+export function getNumberLiteralValue(expression?: lua.Expression) {
+    if (!expression) return undefined;
+    return lua.isNumericLiteral(expression)
+        ? expression.value
+        : lua.isUnaryExpression(expression) &&
```
Perryvw, line 55: Not very readable, probably easier to read if you change the first conditional to a regular if/else?

src/transformation/visitors/access.ts
```diff
@@ -16,32 +17,31 @@ export function transformElementAccessArgument(
 
     const type = context.checker.getTypeAtLocation(node.expression);
     const argumentType = context.checker.getTypeAtLocation(node.argumentExpression);
-    if (isNumberType(context, argumentType) && isArrayType(context, type)) {
+    if (isArrayType(context, type) && isNumberType(context, argumentType)) {
```
Perryvw, line 20: What is the reason for this change? Just ordering them alphabetically?

test/unit/builtins/string.spec.ts
```diff
@@ -99,6 +103,8 @@ test.each([
     { inp: "hello test", searchValue: "t", offset: 6 },
     { inp: "hello test", searchValue: "t", offset: 7 },
     { inp: "hello test", searchValue: "h", offset: 4 },
+    { inp: "hello test", searchValue: "o", offset: -1 },
```
Perryvw, line 106: Maybe it's nicer to use a searchValue that actually appears in the string multiple times (to show it returns the right one)?

src/transformation/visitors/access.ts
ark120202: I have changed it to match general evaluation order (first LHS then RHS), but there is no any runtime difference there

PR #905, opened 2020-07-14
Method and Method Property decorators

hazzard993:
#537 was just for classes.

Added support for decorators to be used with methods and method properties

```diff
 class Foo {
+  @decorator
   method() {}

+  @decorator
   property;

+  @decorator
+  static method(@a p1, @b @c p2) {}

+  @decorator
   static property;
 }
```


src/transformation/visitors/class/decorators.ts
```diff
@@ -2,42 +2,38 @@ import * as ts from "typescript";
...
+            targetTableName,
+            targetFieldExpression
+        );
+    }
 
-    return lua.createAssignmentStatement(
-        className,
-        transformLuaLibFunction(context, LuaLibFeature.Decorate, undefined, decoratorTable, className)
-    );
+    return transformLuaLibFunction(context, LuaLibFeature.Decorate, undefined, decoratorTable, targetTableName);
```
ark120202, line 38: This looks a bit better to me than that `if (targetFieldExpression)` condition above
```suggestion
    const parameters = [decoratorTable, targetTableName];
    if (targetFieldExpression) parameters.push(targetFieldExpression);
    return transformLuaLibFunction(context, LuaLibFeature.Decorate, undefined, ...parameters);
```

src/transformation/visitors/class/index.ts
```diff
@@ -282,29 +285,61 @@ function transformClassLikeDeclaration(
         }
     }
 
+    const noPrototype = isExtension || isMetaExtension;
+
+    // Decorate accessors
```
ark120202, line 290: I think it would be better to refactor it into a single-pass loop over class members. Currently decorator evaluation order would be different from TypeScript ([Playground](https://www.typescriptlang.org/play/index.html?experimentalDecorators=true&target=1#code/MYGwhgzhAECC0G8BQ1XQAIApMEpoF4A+aYAewDsJSQBTAOhFIHNMByAWxoBcALUgEwCMrHDhRpOvAYNyJoAXyTjUTbtFVcuNAE4y8CBcuhGssoiQpVaDZmw1btIsWmgQuYLgEtg6tfZ0ATLIGikYm2HjmZJTU9IwsrAAO2qSJTkaJAK4ARiDe0MmpYS6mkcTRVnG2HNx8-AHpLpJ1QfqK8kA)). It also should make this code easier to read.

```ts
const decorationStatements = [];
for (const member of classDeclaration.members) {
	if (ts.isAccessor(member)) {
		result.push(transformAccessor(member));
		if (member.decorators) decorationStatements.push(...)
	} else if (ts.isMethodDeclaration(member)) {
		result.push(transformMethod(member));
	} else ...
}
result.push(...decorationStatements);
```

hazzard993:
Reverting to draft just to clean up and make some more tests

hazzard993:
Added some more descriptor features

```ts
Object.defineProperty(target, key, descriptor);
Object.getOwnPropertyDescriptor(target, key);
Object.getOwnPropertyDescriptors(target, key);
```

Descriptor objects can also be used in decorators in a mostly similar way to TypeScript

```ts
const decorator = (target, key, desc) => {
  desc.value = false;
  desc.writable = false;
  desc.configurable = false;
};

class Foo {
  @decorator
  static bar() {}
}

Foo.bar; // this value is now "false"
Foo.bar = 5; // throws runtime error because writable=false
delete Foo.bar; // throws runtime error because configurable=false
```

src/lualib/Decorate.ts
```diff
@@ -11,8 +11,17 @@ function __TS__Decorate(this: void, decorators: Function[], target: {}, key?: st
...
                 result = decorator(result);
-            } else if (desc !== undefined) {
-                result = decorator(target, key, result);
+            } else if (desc === true) {
+                const value = rawget(target, key);
+                let descriptor: PropertyDescriptor;
+                if (value === undefined) {
+                    descriptor = __TS__ObjectGetOwnPropertyDescriptor(target, key);
+                } else {
+                    descriptor = __TS__CloneDescriptor({ configurable: true, writable: true, value });
```
ark120202, line 20: Can't it just be an object, without using `__TS__CloneDescriptor` here?

src/lualib/Decorate.ts
```diff
@@ -11,8 +11,17 @@ function __TS__Decorate(this: void, decorators: Function[], target: {}, key?: st
...
+                }
+                __TS__SetDescriptor(target, key, { ...descriptor, ...decorator(target, key, descriptor) });
```
ark120202, line 22: `__TS__SetDescriptor` has a quite large performance impact. It would be nice to avoid using actual descriptors, unless it's necessary

```ts
const isSimpleValue = descriptor.configurable === true && descriptor.writable === true && !descriptor.get && !descriptor.set;
if (isSimpleValue) {
	target[key] = descriptor.value;
} else {
	__TS__SetDescriptor(target, key, descriptor);
}
```

hazzard993: Descriptors shouldn't exist until the user decides to put them in via decorators/getters/setters or `Object.defineProperty`

An issue here is if the user explicitly sets the descriptor they may not be able to see it via `Object.getOwnPropertyDescriptor` since it is optimised away

So should we implement and note this as a bug/quirk of descriptors in TSTL?

hazzard993: Or maybe an opt-in `--optimizeDescriptors` option that engages this behaviour for generated code so users can have a choice

ark120202: > An issue here is if the user explicitly sets the descriptor they may not be able to see it via Object.getOwnPropertyDescriptor since it is optimised away

That would be an issue anyway, because non-decorated members should have descriptors too. So I think it's fine to accept it as an another point of difference

test/unit/builtins/array.spec.ts
```diff
@@ -209,19 +209,11 @@ describe("delete", () => {
...
             const array = [1, 2, 3, 4];
-            const exists = delete array[2];
-            return { exists, a: array[0], b: array[1], c: array[2], d: array[3] };
-        `.expectToMatchJsResult();
-    });
-
-    test("returns false when element not exists", () => {
-        util.testFunction`
-            const array = [1, 2, 3, 4];
-            const exists = delete array[4];
```
Perryvw, line 223: Why was this test removed? It still seems a valid thing to test.

test/unit/descriptors.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+    });
+
+    test.each(["return { value: true }", "desc.value = true"])(
+        "Use decorator to override method value",
+        overrideStatement => {
+            util.testFunction`
+                const decorator = (target, key, desc): any => { ${overrideStatement} };
+                class Foo { @decorator static method() {} }
+                return Foo.method;
+            `.expectToMatchJsResult();
```
Perryvw, line 89: Maybe it would be nice to also have a `.expectToEqual(true)` for this one

test/unit/builtins/array.spec.ts
hazzard993: Both tests return `true` regardless if the element exists or not in actual JS

```
> array = [1, 2, 3, 4];
[ 1, 2, 3, 4 ]
> delete array[4];
true
```

I'm not sure of a way to delete an element from an array and expect a `false` result

test/unit/descriptors.spec.ts
hazzard993: I used this one thinking

- More test cases could be added, some which don't just change `Foo.method` to `true`
- To ensure no new transpiler specific behaviour is introduced, match what JS does

I think these justify the `expectToMatchJsResult`, what do you think?

test/unit/builtins/array.spec.ts
hazzard993: I looked into it a bit more and I've put the test back using `Object.defineProperty` to expect `success` to be `false`

src/lualib/ObjectGetOwnPropertyDescriptors.ts
```diff
@@ -0,0 +1,5 @@
+function __TS__ObjectGetOwnPropertyDescriptors(this: void, object: any): Record<any, PropertyDescriptor | undefined> {
+    const metatable = getmetatable(object);
+    if (!metatable) return;
```
ark120202, line 3:
```suggestion
    if (!metatable) return {};
```

src/transformation/visitors/class/members/method.ts
```diff
@@ -34,3 +44,40 @@ export function transformMethodDeclaration(
...
+    node: ts.MethodDeclaration,
+    className: lua.Identifier,
+    noPrototype: boolean
+): lua.Statement | undefined {
+    const methodTable = transformMemberExpressionOwnerName(node, className, noPrototype);
+    const methodName = transformMethodName(context, node);
+
+    const parameterDecorators = node.parameters
+        .flatMap((parameter, index) =>
+            parameter?.decorators?.map(decorator =>
```
ark120202, line 59: Is `parameter` nullable?

test/unit/builtins/array.spec.ts
```diff
@@ -209,19 +209,27 @@ describe("delete", () => {
...
             const array = [1, 2, 3, 4];
-            const exists = delete array[4];
-            return { exists, a: array[0], b: array[1], c: array[2], d: array[3] };
+            Object.defineProperty(array, 2, { configurable: false });
+
+            let success;
+            try {
+                success = delete array[2];
+            } catch {
+                success = false;
```
ark120202, line 229: Whether it thrown or not is something we want to compare too
```suggestion
                success = "error";
```

test/unit/classes/decorators.spec.ts
```diff
@@ -126,3 +126,60 @@ test("Exported class decorator", () => {
...
+            decoratorTargetKey = key;
+        };
+
+        class Foo {
+            @decorator
+            ${memberStatement}
+        }
+
+        const targetKind = decoratorTarget === Foo.prototype ? "prototype" : "class";
+        return [targetKind, decoratorTargetKey];
```
ark120202, line 159: Object diffs look a little better when test fails
```suggestion
        return { targetKind, decoratorTargetKey };
```

test/unit/descriptors.spec.ts
```diff
@@ -0,0 +1,92 @@
+import * as util from "../util";
+
+const trueFalseUndefTests = [[true], [false], [undefined]] as const;
+
+describe("Object.defineProperty", () => {
+    test.each(trueFalseUndefTests)("writable (%p)", value => {
+        util.testFunction`
+            const foo = { bar: true };
+            Object.defineProperty(foo, "bar", { writable: ${value} });
+            try { foo.bar = false } catch {}
```
ark120202, line 10: Is there any test that makes sure that it throws?

src/transformation/visitors/class/members/method.ts
hazzard993: Good find, nor is `node` on line 71

test/unit/descriptors.spec.ts
hazzard993: Modified the test to capture errors in the result

test/unit/builtins/array.spec.ts
Perryvw: Good find

Perryvw: Is this try/catch needed? It is not supposed to throw an error right?

hazzard993: Nope it is meant to, you can make a `DeleteExpression` throw an error if you assign an unconfigurable descriptor using `Object.defineProperty` to the index of the array for the element you're trying to delete

hazzard993:
```ts
array = [1, 2, 3, 4];
Object.defineProperty(array, "0", { configurable: false });
delete array[0]; // throws an error within the test JS environment
```

test/unit/descriptors.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+            } catch {
+                err = true;
+            }
+
+            return { prop: foo.bar, err };
+        `.expectToMatchJsResult();
+    });
+});
+
+describe("Decorators /w descriptors", () => {
```
ark120202, line 73: Thoughts on testing that along with other decorator tests?

test/unit/descriptors.spec.ts
```diff
@@ -0,0 +1,97 @@
+import * as util from "../util";
+
+const trueFalseUndefTests = [[true], [false], [undefined]] as const;
+
+describe("Object.defineProperty", () => {
```
ark120202, line 5: And that one in `builtins/object.spec.ts`

ark120202: Also would be nice to have some tests for `getOwnPropertyDescriptor`/`getOwnPropertyDescriptors`

test/unit/builtins/array.spec.ts
Perryvw: Could you split this into two tests then, one testing the error, the other testing the return value & functionality?

test/unit/builtins/object.spec.ts
```diff
@@ -121,3 +121,93 @@ describe(".hasOwnProperty()", () => {
         `.expectToMatchJsResult();
     });
 });
+
+const trueFalseTests = [[true], [false]] as const;
```
hazzard993, line 125: Used to be true, false and undefined

But these two cases aren't possible to replicate in Lua since you can't check is a property exists and is set to undefined

```ts
const foo = { bar: true };
Object.defineProperty(foo, "bar", { configurable: undefined }); // configurable -> false
Object.defineProperty(foo, "bar", {}); // configurable -> true
```

test/unit/builtins/object.spec.ts
```diff
@@ -121,3 +121,93 @@ describe(".hasOwnProperty()", () => {
...
+
+describe("Object.getOwnPropertyDescriptor", () => {
+    test("descriptor is exactly the same as the last one set", () => {
+        util.testFunction`
+            const foo = { bar: true };
+            Object.defineProperty(foo, "bar", {});
+            return Object.getOwnPropertyDescriptor(foo, "bar");
+        `.expectToMatchJsResult();
+    });
+});
```
ark120202, line 203: I think we might want to consider constructing a synthetic descriptor for a regular field in `Object.getOwnPropertyDescriptor`. Otherwise, we need a test negative behavior too (i.e. `Object.getOwnPropertyDescriptor({ foo: true }, 'foo')` to be `undefined`). Same goes for `Object.getOwnPropertyDescriptors`

test/unit/builtins/object.spec.ts
```diff
@@ -121,3 +121,93 @@ describe(".hasOwnProperty()", () => {
...
+
+            return { prop: foo.bar, err };
+        `.expectToMatchJsResult();
+    });
+});
+
+describe("Object.getOwnPropertyDescriptor", () => {
+    test("descriptor is exactly the same as the last one set", () => {
+        util.testFunction`
+            const foo = { bar: true };
```
ark120202, line 198: Do we need `bar: true` here?

test/unit/builtins/object.spec.ts
```diff
@@ -121,3 +121,93 @@ describe(".hasOwnProperty()", () => {
...
+            Object.defineProperty(foo, "bar", {});
+            return Object.getOwnPropertyDescriptor(foo, "bar");
+        `.expectToMatchJsResult();
+    });
```
ark120202, line 202: Not sure if it belongs there or to class accessor test, but I think we need to test that **own**PropertyDescriptor doesn't include inherited descriptors.
```ts
class A { get a() { return true } }
Object.getOwnPropertyDescriptor(new A(), "a") // should be undefined
```

ark120202: Also maybe
```ts
class A { get a() { return true } }
Object.getOwnPropertyDescriptor(A.prototype, "a") // not undefined
```

test/unit/classes/decorators.spec.ts
```diff
@@ -126,3 +126,86 @@ test("Exported class decorator", () => {
...
+            decoratorTargetKey = key;
+        };
+
+        class Foo {
+            @decorator
+            ${memberStatement}
+        }
+
+        const targetKind = decoratorTarget === Foo.prototype ? "prototype" : "class";
+        return { targetKind, decoratorTargetKey };
```
ark120202, line 159: I'd merge this and the next test into something like that:
```ts
		let decoratorParameters: any;
		const decorator = (target, key, index) => {
			const targetKind = target === Foo ? "Foo" : target === Foo.prototype ? "Foo.prototype" : "unknown";
			decoratorParameters = { targetKind, key, index };
        };

		class Foo {
			${classMember}
		}

		return decoratorParameters;
```

test/unit/builtins/array.spec.ts
hazzard993: f78a797 should fix this up

test/unit/classes/decorators.spec.ts
```diff
@@ -126,3 +126,60 @@ test("Exported class decorator", () => {
...
+            decoratorTarget = target;
+            decoratorTargetKey = key;
+        };
+
+        class Foo {
+            @decorator
+            ${memberStatement}
+        }
+
+        const targetKind = decoratorTarget === Foo.prototype ? "prototype" : "class";
```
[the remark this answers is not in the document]
hazzard993: Would `expectToMatchJsResult` reveal the discrepancy that `decoratorTarget` was targeting the wrong object?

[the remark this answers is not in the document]
hazzard993: Although I don't understand I do see an alternative in your example code that may fix this, added in d0b0b7e

test/unit/classes/decorators.spec.ts
hazzard993: Good idea, did something like this with 9f88c3f

test/unit/builtins/object.spec.ts
hazzard993: I can't quite replicate this behaviour, the metatable logic is getting the best of me. Would it be okay to bug this for a future fix?

ark120202: Sure, I was assuming it already worked like that

ark120202 (approved):

test/unit/classes/decorators.spec.ts
```diff
@@ -126,3 +126,61 @@ test("Exported class decorator", () => {
...
+        `.expectToMatchJsResult();
+    });
+
+    test.each(["return { value: true }", "desc.value = true"])(
+        "Use decorator to override method value",
+        overrideStatement => {
+            util.testFunction`
+                const decorator = (target, key, desc): any => { ${overrideStatement} };
+                class Foo { @decorator static method() {} }
+                return Foo.method;
```
Perryvw, line 182: Shouldn't this be `Foo.method()`?

Perryvw: nevermind, I see it's right.

PR #910, opened 2020-07-31
fixed declarations that reference their own identifiers

### LINKED ISSUE #894: Referencing return value in callback passed as argument
[playground](https://typescripttolua.github.io/play#code/PQKhAIAEDsHsGUCmAbAZgSWgMQJbMeCMAFAAmiAxsgIYBOBqArtBQC46zTgAOtO0rABQA6UXQDmAZwBc4ZgGs4Ad2gBtALoBKWQDdYOUgG5ixfK3CTGAI0kU+VxLVmDN4ALwA+cHoPHi-VkdUagoCAAlqaFJ8cABvVFhYKzpZSVY+aHEAXxNEAA9uWFpzJhZ2TgtrW3tEQQorZ1dPb31SbXAIqJjY4nA+yps7HAdad3B6437welZGWi54xOSncAAiBKS6VazjHOIKTjTwAAsxy0GawV7+l3cvHqmp3gDBY+EN5c1J-pyvk3PqsNHABCFyGcDAYCSY6wRjIUg8DLmdZLLbEIA)

This is a similar situation to recursive functions. The correct lua should declare the variable separately from assigning it in these circumstances.
```lua
local h
h = ____exports.subscribe(
    function()
        print(h.foobar)
    end
)
```

tomblind:
fixes #894

### DIFF
--- a/src/transformation/utils/lua-ast.ts
+++ b/src/transformation/utils/lua-ast.ts
@@ -4,10 +4,10 @@ import * as lua from "../../LuaAST";
 import { assert, castArray } from "../../utils";
 import { TransformationContext } from "../context";
 import { createExportedIdentifier, getIdentifierExportScope } from "./export";
-import { peekScope, ScopeType } from "./scope";
-import { isFunctionType } from "./typescript";
+import { peekScope, ScopeType, Scope } from "./scope";
 import { transformLuaLibFunction } from "./lualib";
 import { LuaLibFeature } from "../../LuaLib";
+import { isArray } from "util";
 
 export type OneToManyVisitorResult<T extends lua.Node> = T | T[] | undefined;
 export function unwrapVisitorResult<T extends lua.Node>(result: OneToManyVisitorResult<T>): T[] {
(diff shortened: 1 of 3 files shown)


src/transformation/utils/lua-ast.ts
```diff
@@ -129,6 +129,19 @@ export function createHoistableVariableDeclarationStatement(
...
 
+function hasMultipleReferences(scope: Scope, identifiers: lua.Identifier | lua.Identifier[]) {
+    const scopeSymbols = scope.referencedSymbols;
+    if (!scopeSymbols) {
+        return false;
+    }
+
+    const referenceLists = isArray(identifiers)
+        ? identifiers.map(i => i.symbolId && scopeSymbols.get(i.symbolId))
+        : [identifiers.symbolId && scopeSymbols.get(identifiers.symbolId)];
```
ark120202, line 140:
```suggestion
    const referenceLists = castArray(identifiers).map(i => i.symbolId && scopeSymbols.get(i.symbolId));
```

test/unit/identifiers.spec.ts
```diff
@@ -788,3 +788,39 @@ test("lua built-in as in constructor assignment", () => {
         export const result = new A("42").error;
     `.expectToMatchJsResult();
 });
+
+test("local variable declaration referencing self indirectly", () => {
```
ark120202, line 792: It doesn't seem to fit the `identifiers.spec.ts` test file, which generally tests identifier printing itself, not how it's used. I think it would make sense to test it along with self-calling function, or general assignment tests

src/transformation/utils/lua-ast.ts
```diff
@@ -129,6 +128,17 @@ export function createHoistableVariableDeclarationStatement(
...
 
+function hasMultipleReferences(scope: Scope, identifiers: lua.Identifier | lua.Identifier[]) {
+    const scopeSymbols = scope.referencedSymbols;
+    if (!scopeSymbols) {
+        return false;
+    }
+
+    const referenceLists = castArray(identifiers).map(i => i.symbolId && scopeSymbols.get(i.symbolId));
+
+    return referenceLists.some(symbolRefs => symbolRefs && symbolRefs.length > 1);
```
Perryvw, line 139: Maybe
```suggestion
    return referenceLists.some(symbolRefs => symbolRefs?.length > 1);
```

tomblind: This change won't work as `symbolRefs?.length` resolves to `number | undefined` which can't be compared to a number like that. I could change it to something like `(symbolRefs?.length || 0) > 1`, but I feel like it looses clarity that way.

Issue #811, opened 2020-01-20
Annotating callback argument as @tuplereturn


ElusiveMori:
```ts
/** @vararg */
interface Vararg<T> extends Array<T> {}

/* Transpiled:
function vararger(...)
end
*/
function vararger(...args: Vararg<any>): void {}

/* Transpiled:
function hello(a, b, ...)
    vararger(...)
end
*/
function hello(a: number, b: number, ...rest: Vararg<any>): void {
    vararger(...rest)
}

/* Transpiled:
function ohno(callback)
    hello(
        1,
        2,
        table.unpack(
            callback()
        )
    )
end
*/
//                             V - ?
function ohno(callback: () => any[]) {
    hello(1, 2, ...callback())
}

/* Transpiled:
ohno(
    function() return {1, 2, 3} end
)
*/
ohno(() => [1, 2, 3])
```

When you have the opportunity, passing arguments directly by varargs is more efficient than packing/unpacking it. In LuaJIT especially, packing/unpacking is a massive pessimization that reduces performance to interpreter-level, whereas direct vararg passing is an optimized case (probably due to not allocating excess tables).

To this end, it would be nice to have a way to make the above code transpile into something like this instead:
```lua
function ohno(callback)
    hello(1, 2, callback())
end

ohno(function() return 1, 2, 3 end)
```

Maybe this could be the same `@tuplereturn` annotation but now also allowed on function parameters?

ark120202:
It's already kind of supported: 
[...]

PR #911, opened 2020-07-31
Enabled @tupleReturn on array-like types

### LINKED ISSUE #893: const assertion on tupleReturn causes pack+unpack on return value
[playground](https://typescripttolua.github.io/play#code/PQKhAIAEBcFcAcA2BTASsuAnAduEwAoAM1mwGNoBLAe1yOuoAoBKcAbwPC-Ew1h3ABtAET1qwgDThhAIwCGmYQF1wcgM7gytNdADcBAL5A)

tomblind:
This PR enables use of the `@tupleReturn` annotation on array-like types, so that functions which return a type with the annotation will be treated as if the function was annotated as such. This is basically a step towards implementing #580, and helps situations like #811.

Incidentally, this also fixes #893, as that was a general problem with casting, which I needed to work for some of the tests.

Examples:
```ts
/** @tupleReturn */
type TupleReturn<A extends unknown[]> = A & { __tupleReturn?: never };
```
*Basic Usage*
```ts
function returnTuple(): TupleReturn<string[]> {
    return ["a", "b", "c"];
}
```
*Inferred Return Type*
```ts
function returnTuple() {
    return ["a", "b", "c"] as TupleReturn<string[]>;
}
```
*Automatic Forwarding*
```ts
function forwardTuple() {
    return returnTuple();
}
const [a, b, c] = forwardTuple(); // No wrapping or unpacking
```
*Function Expression Inference*
```ts
declare function takesTupleReturnFn(cb: () => TupleReturn<string[]>): void;
takesTupleReturnFn(() => ["a", "b", "c"]); // Arrow function will be transpiled as a tupleReturn function
```
*Overloading*
```ts
declare function returnsArray(): string[];
declare function returnsTuple(): TupleReturn<string[]>;

declare function overloaded<R extends TupleReturn<unknown[]>>(cb: () => R): R;
declare function overloaded<R>(cb: () => R): R;

const [a, b, c] = overloaded(returnsArray); // a, b, c = unpack(overloaded(returnsArray))
const [d, e, f] = overloaded(returnsTuple); // d, e, f = overloaded(returnsTuple)
```
This last example is not currently possible with the function annotation version. It will be useful for typing functions such as `pcall` which take arbitrary functions call them, and return their result.

### DIFF
--- a/src/transformation/utils/annotations.ts
+++ b/src/transformation/utils/annotations.ts
@@ -118,6 +118,10 @@ export function isTupleReturnCall(context: TransformationContext, node: ts.Node)
             return true;
         }
 
+        if (isTupleReturnType(signature.getReturnType()) && !isLuaIteratorType(context, node)) {
+            return true;
+        }
+
         // Only check function type for directive if it is declared as an interface or type alias
         const declaration = signature.getDeclaration();
         const isInterfaceOrAlias =
@@ -162,6 +166,10 @@ export function isInTupleReturnFunction(context: TransformationContext, node: ts
     }
 
     return getTypeAnnotations(functionType).has(AnnotationKind.TupleReturn);
+}
+
+export function isTupleReturnType(type: ts.Type): boolean {
+    return getTypeAnnotations(type).has(AnnotationKind.TupleReturn);
 }
 
 export function isLuaIteratorType(context: TransformationContext, node: ts.Node): boolean {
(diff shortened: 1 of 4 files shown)


ark120202:
Personally I'd like to avoid introducing new features under `@tupleReturn` name, since it's going to be renamed in #821

PR #914, opened 2020-09-11
Update to TypeScript 4

### LINKED ISSUE #912: Feature request: typescript 4 labeled tuple support
It's not really clear when defining a tuple without any name
```typescript
type UserInfo = [string, number, boolean]
function getUserInfo(): UserInfo
```
from the lua side it looks much clear
```lua
local username, age, isRegistered = getUserInfo()
```

Since typescript 4 supports labeled tuple, it would be great to have them as
```typescript
type UserInfo = [username: string, age: number, isRegistered: boolean];
```
Link to the announcement https://devblogs.microsoft.com/typescript/announcing-typescript-4-0-beta/#labeled-tuple-elements

hazzard993:
Fixes #912

Tried to keep changes small

Removed tests that no longer passed TypeScript's type checker

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
         "@types/node": "^13.7.7",
         "@types/resolve": "1.14.0",
         "@typescript-eslint/eslint-plugin": "^2.31.0",
-        "@typescript-eslint/parser": "^2.31.0",
+        "@typescript-eslint/parser": "^4.1.0",
         "eslint": "^6.8.0",
         "eslint-plugin-import": "^2.20.1",
         "eslint-plugin-jest": "^23.8.2",
@@ -58,7 +58,7 @@
         "jest-circus": "^25.1.0",
         "lua-types": "^2.8.0",
         "prettier": "^2.0.5",
-        "ts-jest": "^26.2.0",
+        "ts-jest": "^26.3.0",
         "ts-node": "^8.6.2"
     }
 }
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "dependencies": {
         "resolve": "^1.15.1",
         "source-map": "^0.7.3",
-        "typescript": "^4.0.2"
+        "typescript": ">=4.0.2"
     },
     "devDependencies": {
         "@types/fs-extra": "^8.1.0",
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "dependencies": {
         "resolve": "^1.15.1",
         "source-map": "^0.7.3",
-        "typescript": "^3.9.2"
+        "typescript": "^4.0.2"
     },
     "devDependencies": {
         "@types/fs-extra": "^8.1.0",
@@ -58,7 +58,7 @@
         "jest-circus": "^25.1.0",
         "lua-types": "^2.8.0",
         "prettier": "^2.0.5",
-        "ts-jest": "^26.0.0",
+        "ts-jest": "^26.2.0",
         "ts-node": "^8.6.2"
     }
 }
--- a/src/transformation/visitors/binary-expression/compound.ts
+++ b/src/transformation/visitors/binary-expression/compound.ts
@@ -48,7 +48,7 @@ type CompoundAssignmentToken =
     | ts.SyntaxKind.GreaterThanGreaterThanToken
     | ts.SyntaxKind.GreaterThanGreaterThanGreaterThanToken;
 
-const compoundToAssignmentTokens: Record<ts.CompoundAssignmentOperator, CompoundAssignmentToken> = {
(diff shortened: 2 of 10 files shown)


package.json
```diff
@@ -38,7 +38,7 @@
     "dependencies": {
         "resolve": "^1.15.1",
         "source-map": "^0.7.3",
-        "typescript": "^3.9.2"
+        "typescript": "^4.0.2"
```
ark120202, line 41: `^` doesn't make much sense with typescript's versioning scheme anyway
```suggestion
        "typescript": ">=4.0.2"
```

package.json
```diff
@@ -47,7 +47,7 @@
         "@types/node": "^13.7.7",
         "@types/resolve": "1.14.0",
         "@typescript-eslint/eslint-plugin": "^2.31.0",
```
ark120202, line 49: Plugin version should match the parser, otherwise some rules might handle AST incorrectly. Though if it works okay with current codebase, and there would be some rules with changed configuration format, I think it's fine to keep it that way for now. I could update it to match my [base config](https://github.com/ark120202/eslint-config/blob/master/bases/typescript.js) later

test/unit/classes/accessors.spec.ts
```diff
@@ -125,37 +125,6 @@ test("set accessor in base class", () => {
     `.expectToMatchJsResult();
 });
 
-test("set accessor override", () => {
```
ark120202, line 128: Could you check skipped tests tests, I think they shouldn't be relevant now too

src/transformation/visitors/return.ts
```diff
@@ -1,12 +1,51 @@
...
 import * as lua from "../../LuaAST";
-import { FunctionVisitor } from "../context";
+import { FunctionVisitor, TransformationContext } from "../context";
 import { isInTupleReturnFunction, isTupleReturnCall } from "../utils/annotations";
 import { validateAssignment } from "../utils/assignment-validation";
 import { createUnpackCall, wrapInTable } from "../utils/lua-ast";
 import { ScopeType, walkScopesUp } from "../utils/scope";
 import { isArrayType } from "../utils/typescript";
 
+function transformExpressionToTupleExpressions(
```
ark120202, line 10: `TupleExpressions` doesn't make much sense for me. Maybe `transformExpressionsInReturn`, and let it handle https://github.com/hazzard993/TypescriptToLua/blob/update-typescript/src/transformation/visitors/return.ts#L71-L75 too, so that both usages would be simplified to `lua.createReturnStatement(transformExpressionsInReturn(...))` and `const results = transformExpressionsInReturn(...)`?

```ts
if (!isInTupleReturnFunction(context, node)) {
	return [context.transformExpression(statement.expression)];
}

let results: lua.Expression[];
// ...
```

src/transformation/visitors/function.ts
```diff
@@ -52,7 +53,12 @@ function isRestParameterReferenced(context: TransformationContext, identifier: l
     return references.some(r => !r.parent || !ts.isSpreadElement(r.parent) || !isVarargType(context, r));
 }
 
-export function transformFunctionBodyStatements(context: TransformationContext, body: ts.Block): lua.Statement[] {
+export function transformFunctionBodyStatements(context: TransformationContext, body: ts.ConciseBody): lua.Statement[] {
```
ark120202, line 56: `Statements` doesn't make much sense given that now it works on expressions too. `transformFunctionBodyContent`?

test/unit/classes/accessors.spec.ts
hazzard993: Yep they were too, removed them as well

src/transformation/visitors/function.ts
hazzard993: thanks again @ark120202!

src/transformation/visitors/binary-expression/compound.ts
```diff
@@ -48,7 +48,7 @@ type CompoundAssignmentToken =
     | ts.SyntaxKind.GreaterThanGreaterThanToken
     | ts.SyntaxKind.GreaterThanGreaterThanGreaterThanToken;
 
-const compoundToAssignmentTokens: Record<ts.CompoundAssignmentOperator, CompoundAssignmentToken> = {
```
Perryvw, line 51: Which CompoundAssignmentOperator do we not support now?

ark120202: ??=, &&=, ||=

Perryvw: For this PR we should either create an issue to support these compound operators later, or just revert the Partial/ | undefined changes and add them here, but then we also need some tests for them. I'm leaving it up to @hazzard993 to decide.

hazzard993: Created #916 to complete this later

ark120202 (approved):

PR #919, opened 2020-10-02
Add in missing 4.0 compound assignments

### LINKED ISSUE #916: Support all CompoundAssignmentOperators
For this PR we should either create an issue to support these compound operators later, or just revert the Partial/ | undefined changes and add them here, but then we also need some tests for them. I'm leaving it up to @hazzard993 to decide.

_Originally posted by @Perryvw in https://github.com/TypeScriptToLua/TypeScriptToLua/pull/914#discussion_r487413644_

Perryvw:
Closes #916

### DIFF
--- a/src/transformation/visitors/binary-expression/compound.ts
+++ b/src/transformation/visitors/binary-expression/compound.ts
@@ -46,9 +46,12 @@ type CompoundAssignmentToken =
     | ts.SyntaxKind.AsteriskAsteriskToken
     | ts.SyntaxKind.LessThanLessThanToken
     | ts.SyntaxKind.GreaterThanGreaterThanToken
-    | ts.SyntaxKind.GreaterThanGreaterThanGreaterThanToken;
+    | ts.SyntaxKind.GreaterThanGreaterThanGreaterThanToken
+    | ts.SyntaxKind.BarBarToken
+    | ts.SyntaxKind.AmpersandAmpersandToken
+    | ts.SyntaxKind.QuestionQuestionToken;
 
-const compoundToAssignmentTokens: Partial<Record<ts.CompoundAssignmentOperator, CompoundAssignmentToken>> = {
+const compoundToAssignmentTokens: Record<ts.CompoundAssignmentOperator, CompoundAssignmentToken> = {
     [ts.SyntaxKind.BarEqualsToken]: ts.SyntaxKind.BarToken,
     [ts.SyntaxKind.PlusEqualsToken]: ts.SyntaxKind.PlusToken,
     [ts.SyntaxKind.CaretEqualsToken]: ts.SyntaxKind.CaretToken,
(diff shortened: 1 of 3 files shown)


src/transformation/visitors/binary-expression/compound.ts
```diff
@@ -180,13 +181,86 @@ export function transformCompoundAssignmentStatement(
...
+            lua.createIfStatement(
+                lua.createBinaryExpression(lhs, lua.createNilLiteral(), lua.SyntaxKind.EqualityOperator),
+                lua.createBlock([lua.createAssignmentStatement(lhs, context.transformExpression(rhs))]),
+                undefined,
+                node
+            ),
+        ];
+    } else {
+        context.diagnostics.push(unsupportedNodeKind(node, operator));
+        return [];
```
ark120202, line 264: Since `operator` is a union, and all it's values are handled above, you can use `assertNever` to statically verify that this branch would never be executed
```suggestion
	assertNever(operator);
```

src/transformation/visitors/binary-expression/compound.ts
```diff
@@ -180,13 +181,86 @@ export function transformCompoundAssignmentStatement(
...
+                lua.createUnaryExpression(lhs, lua.SyntaxKind.NotOperator),
+                lua.createBlock([lua.createAssignmentStatement(lhs, context.transformExpression(rhs))]),
+                undefined,
+                node
+            ),
+        ];
+    } else if (operator === ts.SyntaxKind.QuestionQuestionToken) {
+        return [
+            lua.createIfStatement(
+                lua.createBinaryExpression(lhs, lua.createNilLiteral(), lua.SyntaxKind.EqualityOperator),
```
ark120202, line 256: All these 3 branches appear to be different only in the first parameter here. Thoughts on just extracting it into a `condition` variable?

test/unit/assignments.spec.ts
```diff
@@ -404,3 +404,56 @@ test("nullish coalescing compound assignment side effect not evaluated", () => {
...
+        const inst = new MyClass();
+        inst.prop ${operator} 8;        
+    `.expectToMatchJsResult();
+});
+
+test.each([
+    { operator: "||=", initialValue: false },
+    { operator: "&&=", initialValue: true },
+    { operator: "??=", initialValue: undefined },
+])("compound assignment side effects", ({ operator, initialValue }) => {
```
ark120202, line 444: Shouldn't it be the case with other compound assignment operators too?

test/unit/assignments.spec.ts
```diff
@@ -404,3 +404,56 @@ test("nullish coalescing compound assignment side effect not evaluated", () => {
...
+        * x.y &&= z is translated to x.y && (x.y = z).
+        * x.y ||= z is translated to x.y !== undefined && (x.y = z).
+        
+        Test if setter in Lua is called same nr of times as in JS.
+    */
+    util.testModule`
+        export let setterCalled = 0;
+
+        class MyClass {
+
```
ark120202, line 425:
```suggestion
```

test/unit/assignments.spec.ts
Perryvw: true, adding some more.

src/transformation/visitors/binary-expression/index.ts
```diff
@@ -45,13 +46,18 @@ export function transformBinaryOperation(
...
     node: ts.Node
 ): lua.Expression {
     if (isBitOperator(operator)) {
         return transformBinaryBitOperation(context, node, left, right, operator);
     }
 
+    if (operator === ts.SyntaxKind.QuestionQuestionToken) {
+        assert(ts.isBinaryExpression(node));
+        return transformNullishCoalescingExpression(context, node);
+    }
```
ark120202, line 59: Do we need this branch? If it's required, we can remove `case ts.SyntaxKind.QuestionQuestionToken` from `transformBinaryExpression`

src/transformation/visitors/binary-expression/compound.ts
```diff
@@ -175,13 +189,77 @@ export function transformCompoundAssignmentStatement(
...
         return transformAssignment(context, lhs, operatorExpression);
     }
 }
+
+/* These setter-skipping operators will not execute the setter if result does not change.
+ * x.y ||= z does NOT call the x.y setter if x.y is already true.
+ * x.y &&= z does NOT call the x.y setter if x.y is already false.
+ * x.y ??= z does NOT call the x.y setter if x.y is already not nullish.
+ */
+type SetterSkippingCompoundAssignmentOperator =
```
ark120202, line 221: After checking out [the proposal](https://github.com/tc39/proposal-logical-assignment) it seems to be intrinsic behavior for logical assignment concept, so I'd call it `LogicalAssignmentOperator`

src/transformation/visitors/binary-expression/index.ts
Perryvw: Yep you're right, it is needed here, I removed the case from transformBinaryExpression.

src/transformation/visitors/binary-expression/compound.ts
Perryvw: Actually TS already has a `type LogicalOperator = AmpersandAmpersandToken | BarBarToken`, but it does not include `??`, I propose we keep this as it is to avoid confusing naming, I did update the type to use `ts.LogicalOperator` instead of the individual tokens.

ark120202 (approved):

PR #920, opened 2020-10-04
Implement parseInt and parseFloat

### LINKED ISSUE #827: Polyfills for parseFloat and parseInt
These are commonly used functions in existing TS codebases. It would be nice to have some polyfills for these to better support the transpilation of TS code without having to use lua specific function calls e.g. `tonumber`.

In addition to that they are just generally helpful functions.

Perryvw:
Closes #827

### DIFF
--- a/src/lualib/ParseFloat.ts
+++ b/src/lualib/ParseFloat.ts
@@ -1,0 +1,4 @@
+function __TS__ParseFloat(this: void, numberString: string): number {
+    const number = tonumber(string.match(numberString, "^-?%d+%.?%d*"));
+    return number ?? NaN;
+}
--- a/src/lualib/ParseFloat.ts
+++ b/src/lualib/ParseFloat.ts
@@ -1,4 +1,11 @@
 function __TS__ParseFloat(this: void, numberString: string): number {
-    const number = tonumber(string.match(numberString, "^-?%d+%.?%d*"));
+    // Check if string is infinity
+    const infinityMatch = string.match(numberString, "^%s*(-?Infinity)");
+    if (infinityMatch) {
+        // eslint-disable-next-line @typescript-eslint/prefer-string-starts-ends-with
+        return infinityMatch[0] === "-" ? -Infinity : Infinity;
+    }
+
+    const number = tonumber(string.match(numberString, "^%s*(-?%d+%.?%d*)"));
     return number ?? NaN;
 }
--- a/src/lualib/ParseInt.ts
+++ b/src/lualib/ParseInt.ts
@@ -1,28 +1,38 @@
-const __TS__parseInt_patterns = {
-    2: "^-?[01]*",
-    3: "^-?[012]*",
-    4: "^-?[0123]*",
-    5: "^-?[01234]*",
-    6: "^-?[012345]*",
-    7: "^-?[0123456]*",
-    8: "^-?[01234567]*",
-    9: "^-?[012345678]*",
-    10: "^-?[0123456789]*",
-    11: "^-?[0123456789aA]*",
-    12: "^-?[0123456789aAbB]*",
-    13: "^-?[0123456789aAbBcC]*",
-    14: "^-?[0123456789aAbBcCdD]*",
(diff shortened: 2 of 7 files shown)


src/lualib/ParseInt.ts
```diff
@@ -0,0 +1,31 @@
...
+    7: "^-?[0123456]*",
+    8: "^-?[01234567]*",
+    9: "^-?[012345678]*",
+    10: "^-?[0123456789]*",
+    11: "^-?[0123456789aA]*",
+    12: "^-?[0123456789aAbB]*",
+    13: "^-?[0123456789aAbBcC]*",
+    14: "^-?[0123456789aAbBcCdD]*",
+    15: "^-?[0123456789aAbBcCdDeE]*",
+    16: "^-?[0123456789aAbBcCdDeEfF]*",
```
ark120202, line 16: Looks like it actually supports bases up to 36

src/lualib/ParseInt.ts
```diff
@@ -0,0 +1,31 @@
...
+};
+
+function __TS__ParseInt(this: void, numberString: string, base?: number): number {
+    const number = tonumber(string.match(numberString, __TS__parseInt_patterns[base ?? 10]), base);
```
ark120202, line 20: When `base` is out of supported range (<2 or >36) it should return `undefined`

src/lualib/ParseFloat.ts
```diff
@@ -0,0 +1,4 @@
+function __TS__ParseFloat(this: void, numberString: string): number {
+    const number = tonumber(string.match(numberString, "^-?%d+%.?%d*"));
```
ark120202, line 2: JS implementation also handles `Infinity` and `-Infinity`

src/lualib/declarations/math.d.ts
```diff
@@ -8,4 +8,7 @@ declare namespace math {
     function atan(y: number, x?: number): number;
 
     function atan2(y: number, x: number): number;
+
+    function ceil(x: number): number;
+    function floor(x: number): number;
```
ark120202, line 13: Why not just `Math.*`?

test/unit/builtins/numbers.spec.ts
```diff
@@ -78,3 +78,41 @@ test("number intersected method", () => {
 test("numbers overflowing the float limit become math.huge", () => {
     util.testExpression`1e309`.expectToMatchJsResult();
 });
+
+describe.each(["parseInt", "parseFloat"])("parse numbers with %s", parseFunction => {
+    const numberStrings = ["3", "3.0", "9", "42", "239810241", "-20391", "3.1415", "2.7182", "-34910.3"];
```
ark120202, line 83: JS also seems to trim string first, so `" 1 "` is `1`

src/lualib/ParseInt.ts
ark120202, line 20: > `?? 10`

Citing [MDN page](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt):

> If radix is undefined, 0, or unspecified, JavaScript assumes the following:
> 
> 1. If the input string begins with "0x" or "0X" (a zero, followed by lowercase or uppercase X), radix is assumed to be 16 and the rest of the string is parsed as a hexidecimal number.
> 2. If the input string begins with "0" (a zero), radix is assumed to be 8 (octal) or 10 (decimal). Exactly which radix is chosen is implementation-dependent. ECMAScript 5 clarifies that 10 (decimal) should be used, but not all browsers support this yet. For this reason, always specify a radix when using parseInt.
> 3. If the input string begins with any other value, the radix is 10 (decimal).

src/lualib/declarations/math.d.ts
Perryvw: I know these methods do what we want, if we can have such extremely complicated functions like these as lualib functions I would rather not risk it and use the lowest level possible in lualib implementations.

ark120202 (approved):

PR #922, opened 2020-10-16
Add devcontainer.json configuration

ark120202:


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,14 @@
+{
+    "name": "TypeScriptToLua",
+    "image": "mcr.microsoft.com/vscode/devcontainers/javascript-node:14",
+    "settings": {
+        "terminal.integrated.shell.linux": "/bin/bash"
+    },
+    "extensions": [
+        "ark120202.vscode-typescript-to-lua",
```
Perryvw, line 8: Why is this needed to develop tstl? Isn't this only for when you want to use it?

ark120202: We have 2 lua projects in this repo - lualib and benchmark

PR #878, opened 2020-05-18
Get benchmark data to visualizer

Perryvw:

PR #935, opened 2020-11-05
Fixed various benchmark issues

lolleko:
Benchmarks now work for pull requests from a fork (at least partially see comment: https://github.com/TypeScriptToLua/TypeScriptToLua/pull/935#issuecomment-723495647)
Fixed comparison info not being emitted correctly after #878
Bump actions/github-script from 0.9.0 to v3
Moved create benchmark check script to standalone js file

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,6 +45,8 @@ jobs:
     steps:
       - name: Lua Install
         run: sudo apt-get install lua5.3 luajit
+      - name: Glow Install
+        run: brew install glow
       # Checkout master & commit
       - name: Checkout master
         uses: actions/checkout@v2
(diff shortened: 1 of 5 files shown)


package-lock.json
```diff
@@ -1,8 +1,11622 @@
 {
     "name": "typescript-to-lua",
     "version": "0.36.0",
-    "lockfileVersion": 1,
+    "lockfileVersion": 2,
```
ark120202, line 4: FWIW that would mean that every contributor would have to use Node >=15, otherwise npm v6 would revert these lockfile changes. But not upgrading also isn't great, since then some contributors would have to rollback node. Alternatively we can wait for [corepack](https://github.com/nodejs/corepack) to land, and lock npm to v6 even on newer node versions.

Perryvw (approved):

package-lock.json
lolleko: Yeah I wasn't sure what to do about this one. I have a new system and did a fresh install of node.

The changelog mentions that the new lockfile version is backwards compatible, but I have not been able to try it:

https://blog.npmjs.org/post/626173315965468672/npm-v7-series-beta-release-and-semver-major

>package-lock.json is updated to a newer format, using "lockfileVersion": 2. This format is backwards-compatible with npm CLI versions using "lockfileVersion": 1, but older npm clients will print a warning about the version mismatch.

.github/workflows/ci.yml
```diff
@@ -93,50 +95,24 @@ jobs:
...
-                }
-              });
-            } else {
-              console.log(summary);
-              console.log(text);
-            }
+            const create_benchmark_check = require(`${process.env.GITHUB_WORKSPACE}/commit/.github/scripts/create_benchmark_check.js`);
+            return create_benchmark_check({github, context, core});
+      - name: Benchmark results
+        run: echo "${{steps.script-combine-results.outputs.result}}" | glow -s dark -w 120 -
```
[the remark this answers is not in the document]
lolleko: I think the tables would be quite a bit of work. Also terminal color handling is often different across platforms and emulators, I am not sure how much of an issue that is, but that might require some effort.

lolleko:
**TL;DR** Tried to use a new GitHub feature to allow the benchmarks check to be posted for fork PRs. I had security concerns so instead the benchmark results are just emitted to the action log.

---

My initial idea was to use GitHub actions new `on: pull_request_target` (https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/).

With `pull_request_target` an actions runner for a PR from a fork gets read/write GITHUB_TOKEN.
This would allow us to use [github.checks.create](https://github.com/TypeScriptToLua/TypeScriptToLua/pull/935/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fL124) in workflows triggered by a PR.
To make sure no modified (malicious) workflows are executed with a read/write token, the runner uses the workflows from the upstream's master branch:
> instead of running against the workflow and code from the merge commit, the event runs against the workflow and code from the base of the pull request. This means the workflow is running from a trusted source and is given access to a read/write token as well as secrets enabling the maintainer to safely comment on or label a pull request.

So why not use `pull_request_target`?
The problem is that we execute arbitrary Lua code from the PR commit during benchmarking. If an attacker is able to break out from the LuaVM he could potentially get access to a read/write token for this repo. That is not good, obviously...

I am summarizing my thought process here, as a reference in case anyone thinks about using `pull_request_target` in the future. 
**`pull_request_target` should not be used for any workflow that execute arbitrary code, for example benchmarks or tests.**

Instead the benchmark results are only posted to the log and no check is created.

package-lock.json
ark120202: Running `npm install` on v6 downgrades lockfile back to `"lockfileVersion": 1` and removes new key, which I wouldn't really call backwards compatible. But I guess that's okay

ark120202 (approved):

PR #937, opened 2020-11-24
Fix export all bug and add NamespaceExport support

### LINKED ISSUE #927: `export * from "..."` shouldn't re-export `default`

### LINKED ISSUE #928: `export * as ns from "..."`
- https://github.com/tc39/proposal-export-ns-from
- https://github.com/microsoft/TypeScript/issues/4813
- https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#export--as-ns-syntax

hazzard993:
Fixes #927
Resolves #928

### DIFF
--- a/src/transformation/visitors/modules/export.ts
+++ b/src/transformation/visitors/modules/export.ts
@@ -8,7 +8,6 @@ import {
     createExportedIdentifier,
 } from "../../utils/export";
 import { createExportsIdentifier } from "../../utils/lua-ast";
-import { createSafeName } from "../../utils/safe-names";
 import { ScopeType } from "../../utils/scope";
 import { transformScopeBlock } from "../block";
 import { transformIdentifier } from "../identifier";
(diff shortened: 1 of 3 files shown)


src/transformation/visitors/modules/export.ts
```diff
@@ -46,24 +47,62 @@ function transformExportAllFrom(context: TransformationContext, node: ts.ExportD
...
+            : createExportsIdentifier();
+
+    const leftAssignment = lua.createAssignmentStatement(
+        lua.createTableIndexExpression(lua.cloneIdentifier(left), lua.cloneIdentifier(forKey)),
+        forValue
+    );
+
+    const ifBody = lua.createBlock([leftAssignment]);
+
+    const ifStatement = lua.createIfStatement(
```
Perryvw, line 79: Could you add a comment explaining what the purpose of this if statement is.

src/transformation/visitors/modules/export.ts
```diff
@@ -46,24 +47,62 @@ function transformExportAllFrom(context: TransformationContext, node: ts.ExportD
     const moduleRequire = createModuleRequire(context, node.moduleSpecifier);
     const tempModuleIdentifier = lua.createIdentifier("____export");
 
+    const result: lua.Statement[] = [];
+
     const declaration = lua.createVariableDeclarationStatement(tempModuleIdentifier, moduleRequire);
+    result.push(declaration);
+
+    if (node.exportClause && ts.isNamespaceExport(node.exportClause)) {
```
Perryvw, line 55: Could you add a comment explaining in what case this if statement is needed and what this branch of this function does?

src/transformation/visitors/modules/export.ts
hazzard993: Added some more comments to transformExportAll, I think it should be much easier to read now

PR #946, opened 2020-12-18
Add @luaName annotation

### LINKED ISSUE #945: Representing functions with a name that is a JavaScript reserved word
Currently I am using namespaces to represent Lua globals with functions inside of them- that is okay, but this is not okay:

```ts
/** @noSelf */
declare namespace fs {
	...

	/** Deletes a file or directory. */
	function delete(path: string): void // <---- delete is a reserved word that cannot be used here

	...
}
```

This is a problem - the function is named `delete` but TypeScript does not have the capability to represent this!

Merging is not allowed:

image

And even without type checking, the transpilation only recognizes the namespace declaration above so it doesn't even work correctly:

image

My only option is to include the function definition inside the namespace, but since function identifiers cannot be strings there is no way to name the function `delete`, so there needs to be a way to communicate to TypeScriptToLua that the function on the Lua side is named that way.

The issue? There _isn't_ yet such a feature.

I propose an `@luaName` annotation, which will allow you to specify a Lua name for the function. This would allow me to name the function something different like `_delete` and then in the transpilation step it is turned back to `delete`.

Example:

```ts
/** @luaName bar */
declare function foo(x: string): void

foo('Hi!')
```

would transpile to

```lua
bar('Hi!')
```

Likewise,

```ts
/** @noSelf */
declare namespace fs {
	/** Deletes a file or directory.
	 * @luaName delete */
	function _delete(path: string): void
}

fs._delete('some path')
```

would transpile to

```lua
fs.delete('some path')
```

LoganDark:
This will allow you to apply an `@luaName` annotation to things such as functions, methods, fields, namespaces, and so on, to change the transpiled name.

Example:

Lua:
```lua
local foo = {}
function foo:baz(str)
    return str
end
```

TS:
```ts
/** @luaName foo */
declare namespace bar {
    function baz(str: string): string
}

return bar.baz('hi'); // actually calls foo:baz
```

Fixes #945

### DIFF
--- a/src/transformation/utils/safe-names.ts
+++ b/src/transformation/utils/safe-names.ts
@@ -70,32 +70,42 @@ function checkName(context: TransformationContext, name: string, node: ts.Node):
 export function hasUnsafeSymbolName(
     context: TransformationContext,
     symbol: ts.Symbol,
-    tsOriginal: ts.Identifier
+    tsOriginal: ts.Identifier,
+    luaName = symbol.name
 ): boolean {
     const isAmbient = symbol.declarations && symbol.declarations.some(d => isAmbientNode(d));
 
     // Catch ambient declarations of identifiers with bad names
-    if (isAmbient && checkName(context, symbol.name, tsOriginal)) {
+    if (isAmbient && checkName(context, luaName, tsOriginal)) {
         return true;
     }
 
     // only unsafe when non-ambient and not exported
-    return isUnsafeName(symbol.name) && !isAmbient && !isSymbolExported(context, symbol);
+    return isUnsafeName(luaName) && !isAmbient && !isSymbolExported(context, symbol);
 }
 
 export function hasUnsafeIdentifierName(
     context: TransformationContext,
     identifier: ts.Identifier,
-    checkSymbol = true
+    checkSymbol = true,
+    luaName?: string
 ): boolean {
     if (checkSymbol) {
         const symbol = context.checker.getSymbolAtLocation(identifier);
         if (symbol) {
-            return hasUnsafeSymbolName(context, symbol, identifier);
+            if (luaName === undefined) {
+                luaName = symbol.name
+            }
+
+            return hasUnsafeSymbolName(context, symbol, identifier, luaName);
         }
     }
 
-    return checkName(context, identifier.text, identifier);
+    if (luaName === undefined) {
+        luaName = identifier.text
+    }
+
+    return checkName(context, luaName, identifier);
 }
 
 const fixInvalidLuaIdentifier = (name: string) =>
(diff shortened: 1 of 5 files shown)


LoganDark:
WIP, do not merge

LoganDark:
Alright, I need help with something.

The LuaName.Variable test is failing, and I've managed to narrow down the cause. It took a bit of time to figure it out but it seems that this getTypeAtLocation call is returning a different type than I expect.

https://github.com/TypeScriptToLua/TypeScriptToLua/pull/946/files#diff-a0b0e354d96cb97ae146c66fc8fc06afaa447d0ef870f7b76447ef49147688d3R185

It's returning the type of the value assigned to the variable, not the type of the variable itself. Which, I guess is expected, but how do I get the annotations of the variable declaration?

Please advise

LoganDark:
It do be looking like that issue I ran into affects other code, namely the `@noSelf` annotation, as well.

```ts
/** @noSelf */
declare const somelib: {
    some_func(): void
}

somelib.some_func()
```

turns into

```lua
somelib:some_func()
```

test/unit/annotations/luaName.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+            /** @luaName baz */
+            bar: string
+        }
+    `;
+
+    const result = util.transpileAndExecute("return foo.bar;", undefined, luaHeader, tsHeader);
+
+    expect(result).toBe("hi");
+});
+
```
lolleko, line 94: Add some tests for classes: (vars, functions,  getter/setter...) and interfaces. Also all of this tests use declare. Would be nice to have some tests with actual namespaces/variables and not declarations of them.

lolleko (changes requested):

test/unit/annotations/luaName.spec.ts
LoganDark: Yes, I agree. I need some sleep before I can continue on this, but thank you for the suggestions :D

src/transformation/utils/safe-names.ts
```diff
@@ -70,32 +70,42 @@ function checkName(context: TransformationContext, name: string, node: ts.Node):
 export function hasUnsafeSymbolName(
     context: TransformationContext,
     symbol: ts.Symbol,
-    tsOriginal: ts.Identifier
+    tsOriginal: ts.Identifier,
+    luaName = symbol.name
```
Perryvw, line 74: I really dislike the signature of this function changing for an annotation, same for hasUnsafeIdentifierName, since you have the symbol and identifier nodes, I think it would be nicer to so `getLuaName` inside, or even just avoiding calling these functions if a luaName was specified. I would prefer to keep the impact of this change limited as much as possible to just annotations.ts, with minor adjustments to call.ts and identifier.ts.

LoganDark: Sorry, was unsure exactly what the guarantees were for these arguments. Unfamiliar codebase syndrome. I'll apply your suggestion ASAP

LoganDark:
Closing due to https://github.com/TypeScriptToLua/TypeScriptToLua/issues/945#issuecomment-748551678

PR #948, opened 2020-12-19
Omit tostring from template literal spans if they are strings

### LINKED ISSUE #940: Unnecessary tostring in string templates
Hey,

I've noticed that `Text: ${message}` string template resolves to `"Text: " .. tostring(message)` even if type of message is string.
It would be great to get rid of that if there's any easy fix :)

Best regards

hazzard993:
Closes #940

Seems like a straight forward fix

Using the existing `isStringType` to determine if the type is a string

### DIFF
--- a/test/unit/templateLiterals.spec.ts
+++ b/test/unit/templateLiterals.spec.ts
@@ -60,3 +60,19 @@ test.each(["func`noSelfParameter`", "obj.func`noSelfParameter`", "obj[`func`]`no
         `.expectToMatchJsResult();
     }
 );
+
+test.each([
+    ["'any string'", "string"],
+    ["'string type'", "'string literal type'"],
+    ["'string intersection'", "string & { x: number }"],
+])("tagged template literal omit tostring for string types (%p)", (input, type) => {
+    util.testFunction`
+        function func(msg: ${type}) {
+            return \`func \${msg}\`;
+        }
+
+        return func(${input} as ${type});
+    `
+        .tap(builder => expect(builder.getMainLuaCodeChunk()).not.toContain("tostring"))
+        .expectToMatchJsResult();
+});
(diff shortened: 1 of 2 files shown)


test/unit/templateLiterals.spec.ts
```diff
@@ -60,3 +60,29 @@ test.each(["func`noSelfParameter`", "obj.func`noSelfParameter`", "obj[`func`]`no
...
+
+test.each([
+    ["string", false],
+    ["'string literal type'", false],
+    ["string & unknown", false],
+    ["number", true],
+    ["string | any", true],
+])("template span expect tostring for type (%p) - %p", (type, expectToString) => {
+    util.testFunction`
+        function func(msg: ${type}) {
```
lolleko, line 72: Can this be simplified to something like this?

```ts
let msg: ${type} ;
// @ts-ignore
return  \`\${msg}\`;
```

test/unit/templateLiterals.spec.ts
```diff
@@ -60,3 +60,29 @@ test.each(["func`noSelfParameter`", "obj.func`noSelfParameter`", "obj[`func`]`no
...
+            return \`\${msg}\`;
+        }
+
+        // @ts-ignore
+        return func("");
+    `
+        .tap(builder => {
```
lolleko, line 79: I would prefer one test case for contains "tostring" and one test case for does not contain "tostring". That will result in a small amount of code duplication, but is a lot easier to read imo.

lolleko (changes requested):

test/unit/templateLiterals.spec.ts
hazzard993: Somewhat, TypeScript is a bit too smart for that case

```ts
let msg: string | number = 5;
msg; // type: number
```

But this gets around it

```ts
let msg = 5 as string | number;
msg; // type: string | number
```

lolleko (approved):

PR #949, opened 2020-12-20
Deprecate phantom, metaExtension, extension, pureAbstract

### LINKED ISSUE #652: Deprecate some compiler directives
* `@compileMembersOnly` - too generic, targeted transforms cover it better and avoid name duplication (now it's `EnumName.ENUM_NAME_ENUM_MEMBER`).
* `@extension` - too generic, originally it was designed as a workaround to add members to existing classes, but interface merging + simple property set works, and it's exactly like it's done in JS.
* `@pureAbstract` - it was added only for dota, interface merging covers it.
---
* `@metaExtension` - looks like it was added only for a single environment, I think we might reevaluate the need for it.
* `@customConstructor` - I personally don't see much point in it, wouldn't it be better to type it as an interface and function?
* `@phantom` - what was an original motivation for it? It looks similar to #65 and #645.

lolleko:
I tried to keep the deprecation warnings seperate from existing code/tests, so they can be easily removed once we delete the the deprecated annotations.

Fixes #652

### DIFF
--- a/src/transformation/utils/diagnostics.ts
+++ b/src/transformation/utils/diagnostics.ts
@@ -3,21 +3,32 @@ import { LuaTarget } from "../../CompilerOptions";
 import { createSerialDiagnosticFactory } from "../../utils";
 import { AnnotationKind } from "./annotations";
 
-const createDiagnosticFactory = <TArgs extends any[]>(message: string | ((...args: TArgs) => string)) =>
+type MessageProvider<TArgs extends any[]> = string | ((...args: TArgs) => string);
+
+const createDiagnosticFactory = <TArgs extends any[]>(
+    category: ts.DiagnosticCategory,
+    message: MessageProvider<TArgs>
+) =>
     createSerialDiagnosticFactory((node: ts.Node, ...args: TArgs) => ({
         file: node.getSourceFile(),
         start: node.getStart(),
         length: node.getWidth(),
         messageText: typeof message === "string" ? message : message(...args),
+        category,
     }));
 
-export const unsupportedNodeKind = createDiagnosticFactory(
+const createErrorDiagnosticFactory = <TArgs extends any[]>(message: MessageProvider<TArgs>) =>
+    createDiagnosticFactory(ts.DiagnosticCategory.Error, message);
+const createWarningDiagnosticFactory = <TArgs extends any[]>(message: MessageProvider<TArgs>) =>
+    createDiagnosticFactory(ts.DiagnosticCategory.Warning, message);
+
+export const unsupportedNodeKind = createErrorDiagnosticFactory(
     (kind: ts.SyntaxKind) => `Unsupported node kind ${ts.SyntaxKind[kind]}`
 );
 
-export const forbiddenForIn = createDiagnosticFactory("Iterating over arrays with 'for ... in' is not allowed.");
+export const forbiddenForIn = createErrorDiagnosticFactory("Iterating over arrays with 'for ... in' is not allowed.");
 
-export const unsupportedNoSelfFunctionConversion = createDiagnosticFactory((name?: string) => {
+export const unsupportedNoSelfFunctionConversion = createErrorDiagnosticFactory((name?: string) => {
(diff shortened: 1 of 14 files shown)


src/transformation/utils/diagnostics.ts
```diff
@@ -3,126 +3,146 @@ import { LuaTarget } from "../../CompilerOptions";
...
 );
 
-export const unsupportedVarDeclaration = createDiagnosticFactory(
+export const unsupportedVarDeclaration = createErrorDiagnosticFactory(
     "`var` declarations are not supported. Use `let` or `const` instead."
 );
+
+export const annotationDeprecated = createWarningDiagnosticFactory(
+    (kind: AnnotationKind) =>
+        `'@${kind}' is deprecated and will be removed in a future update. Please update your code before upgrading to the next release, otherwise your project will no longer compile.`
```
Perryvw, line 147: It might be nice to be able to include some kind of url to documentation on how to upgrade, though it makes this a bit more complicated. Thoughts?

lolleko: Good call. We could link to our annotation docs and add a deprecated label to the annotations in questions. In addition we could add some examples of "how to upgrade" e.g. interface merging for pure abstract...

test/legacy-utils.ts
```diff
@@ -13,7 +13,7 @@ export function transpileString(
     const { diagnostics, file } = transpileStringResult(str, options);
     expect(file.lua).toBeDefined();
 
-    const errors = diagnostics.filter(d => !ignoreDiagnostics || d.source === "typescript-to-lua");
+    const errors = ignoreDiagnostics ? [] : diagnostics.filter(d => d.source === "typescript-to-lua");
```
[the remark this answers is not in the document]
lolleko: I moved it out of the filter function because `!ignoreDiagnostics || d.source === "typescript-to-lua"` is not correct. It will not ignore tstl diagnostics, which I thought was the purpose of this flag.
Or is this flag only supposed to ignore TS errors?

Im also fine with using `d.category === error` instead as a temporary workaround. We should try to get rid of legacy utils anyway, I might do a follow up PR for that.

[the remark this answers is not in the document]
lolleko: like this?

`const errors = diagnostics.filter(d => (!ignoreDiagnostics || d.source === "typescript-to-lua") && d.category === ts.DiagnosticCategory.Error);`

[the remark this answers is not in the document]
Perryvw: yeah that's what I was thinking of

test/unit/annotations/__snapshots__/deprecated.spec.ts.snap
```diff
@@ -0,0 +1,26 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`extension deprecation: code 1`] = `""`;
+
+exports[`extension deprecation: code 2`] = `"local __meta__A = debug.getregistry().A"`;
+
+exports[`extension deprecation: diagnostics 1`] = `"main.ts(4,9): warning TSTL: '@extension' is deprecated and will be removed in a future update. Please update your code before upgrading to the next release, otherwise your project will no longer compile. See https://typescripttolua.github.io/docs/advanced/compiler-annotations#extension for more information."`;
+
+exports[`extension deprecation: diagnostics 2`] = `"main.ts(4,9): warning TSTL: '@metaExtension' is deprecated and will be removed in a future update. Please update your code before upgrading to the next release, otherwise your project will no longer compile. See https://typescripttolua.github.io/docs/advanced/compiler-annotations#metaExtension for more information."`;
```
Perryvw, line 9: think link doesn't work, I think the kind needs to be lowercased

lolleko: lowercased it, seems to work now. Also added https://github.com/TypeScriptToLua/TypeScriptToLua.github.io/pull/19 (WIP)

PR #961, opened 2021-01-03
Changed how language-extensions are detected

Perryvw:


Perryvw:
Actually decided to leave this as it is for now

PR #962, opened 2021-01-04
Allow numeric access to MultiReturn

### LINKED ISSUE #960: Indexing MultiReturn return value leads to incorrect Lua
For example:
```ts
function func() {
    return $multi(1,2,3);
}

print(func()[1])
```
Causes `func()[1]` to be translated to `func()[2]`, which is not correct, instead it should be something like `({ func() })[2]`.

Reported here: https://github.com/TypeScriptToLua/TypeScriptToLua/pull/821#issuecomment-753604417

hazzard993:
Resolves #960

```ts
function result() {
  return $multi(1, 2, 3);
}

const first = result()[0];
```

transforms to

```lua
function result() ... end

local first = select(
  1, -- uses +1 rule
  result()
)
```

But diagnostics will be generated if anything else about the type is attempted to be accessed

```ts
result()["forEach"] // not allowed to access via non-numeric type
result().forEach // not allowed to access via property access expression
```

### DIFF
--- a/src/transformation/utils/diagnostics.ts
+++ b/src/transformation/utils/diagnostics.ts
@@ -164,6 +164,10 @@ export const invalidMultiTypeArrayLiteralElementInitializer = createErrorDiagnos
     "This array literal pattern cannot have initializers."
 );
 
+export const invalidMultiReturnAccess = createErrorDiagnosticFactory(
+    "The MultiReturn type can only be accessed via an element access expression of a numeric type."
+);
+
 export const unsupportedMultiFunctionAssignment = createErrorDiagnosticFactory(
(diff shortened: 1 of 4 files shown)


LoganDark:
That actually may be incorrect, it should be this:
```lua
local first = (select(
  1, -- uses +1 rule
  result()
))
```

Note the parenthesis around the call to `select`, this is necessary because `select` returns *all* arguments at the specific index, rather than just one.

Something like this:

```lua
local first, second = select(1, result()), 'second'
```

is incorrect because if `result()` returns more than one value, the second value will be assigned to `second` and the `'second'` will be dropped.

This has correct behavior:

```lua
local first, second = (select(1, result())), 'second'
```

PR #969, opened 2021-01-24
Operator Mapping Library Extensions

### LINKED ISSUE #964: Overloaded Operator Mapping Proposals
One of the few things remaining that can't be mapped from TS to Lua is operator overloads. Since it seems unlikely JS will support syntax for overloading anytime in the near future, I believe we should add some sort of alternative mapping to allow these to be accessed. This is especially important for environments that provide types with functionality that can only be accessed this way (such as math types like vectors).

There's a few ways this could be achieved, so before actually working on a PR, I'd love to get feedback on what the most appropriate way to implement this would be. Here's a few proposals:

---

## Option A: Helper Types
Use helper types, provided by TSTL to allow declaring operator mappings.
```ts
//Provided by TSTL as helper
declare type LuaAdd<A, B, R> = { (a: A, b: B): R; readonly __luaAdd: unique symbol; }
declare type LuaAddMethod<A, B, R> = { (this: A, b: B): R; readonly __luaAdd: unique symbol; }

declare class Vector {
    add: LuaAddMethod<Vector, Vector, Vector>;
    static add: LuaAdd<Vector, Vector, Vector>;
}

declare const addVectors: LuaAdd<Vector, Vector, Vector>;

// All transpile to 'a + b'
declare const a: Vector, b: Vector;
const c = a.add(b);
const d = Vector.add(a, b);
const e = addVectors(a, b);
```

**Pros**
- Gives declaration writers flexibility to access overloads in situationally appropriate ways
- Good type safety and overload support

**Cons**
- Requires a great number of helper types to cover all of the operators
- Need to ensure declared functions are only used for calls and cannot be assigned to/from
- Obfuscates the fact that things are transpiled specially to end users

---

## Option B: Annotations
Similar to Option A, but using annotations instead of types.
```ts
declare class Vector {
    /** @__add */ add(other: Vector): Vector;
    /** @__add */ static add(a: Vector, b: Vector): Vector;
}

/** @__add */ declare function addVectors(a: Vector, b: Vector): Vector;
```

**Pros**
- Same as helper types: flexible and type-safe

**Cons**
- Like helper types, requires extra checks to ensure they aren't mis-used
- Uses annotations, which we've been trying to move away from

---

## Option C: Builtin Helper Function
Provide simple built-in helpers to access operators (similar to `$multi`).
```ts
declare function $add(a: any, b: any): any;
// or
declare function $add<T>(a: T, b: T): T;
declare function $add<A, B>(a: A, b: B): A;
```

**Pros**
- No need for explicit support from declarations
- Makes it clear to users they are invoking something special

**Cons**
- Limited type-safety. Either we use the `any` version, or prevent support for possible type combinations with the generic version.

---

## Option D: User-Defined Helper Function
Allow declaration writers to supply overloads of the helper functions to map types.
```ts
// Must be defined by declaration writers
declare function $add(a: Vector, b: Vector): Vector;
```

**Pros**
- Since declaring global functions with `$` names is not currently allowed, there's no chance of collisions with existing code
- Provides type safety and overload capabilities that Option C did not

**Cons**
- Potential collision issues with different declarations defining similar overloads.
  * If two declarations have identical types for inputs, but different return types, there will be problems.

---

My preference right now is for option A. But I'd like to see if there are other arguments for/against other options, or if there are other possibilities I'm not thinking of.

tomblind:
Resolves #964

Support for mapping lua's metamethod-based operator overloads to function declarations. This allows operator overloads provided by libraries to be callable from TS.

A common example of this would be a math vector class:
```ts
interface Vector {
    x: number;
    y: number;
}

declare function Vector(x: number, y: number): Vector;

const a = Vector(1, 2);
const b = Vector(3, 4);
const result = a + b; // Not allowed
```

To solve this, declaration writers can now map the add operator to a function:
```ts
declare const vectorAdd: LuaAdd<Vector, Vector, Vector>;

const result = vectorAdd(a, b); // transpiles to 'result = a + b'
```

Mappings don't need to be global functions:
```ts
declare namespace Vector {
    export const add: LuaAdd<Vector, Vector, Vector>;
}
const result = Vector.add(a, b); // result = a + b
```

They can also declared as methods on the object using an alternate type:
```ts
interface Vector {
    add: LuaAddMethod<Vector, Vector>;
}
const result = a.add(b); // result = a + b
```

Overloads are allowed using intersections and work as expected:
```ts
declare const vectorMul: LuaMul<Vector, Vector, Vector> & LuaMul<Vector, number, number>;
const result = vectorMul(a, b); // result is Vector
const result = vectorMul(a, 7); // result is number
```

Currently, the following operators are supported:
- Common binary math operators (add, subtract, etc...)
- Negation operator
- Less-than & greater-than operators
- Table length operator
- Concat operator
- Bitwise operators (5.3+ only)


src/transformation/visitors/language-extensions/operators.ts
```diff
@@ -0,0 +1,176 @@
...
+            extensionKind === extensions.ExtensionKind.FloorDivisionOperatorMethodType
+        ) {
+            context.diagnostics.push(unsupportedForTarget(node, "Floor division operator", luaTarget));
+        }
+    }
+
+    const args = node.arguments.slice();
+    if (binaryOperatorMappings.has(extensionKind)) {
+        if (args.length === 1) {
+            assert(ts.isPropertyAccessExpression(node.expression) || ts.isElementAccessExpression(node.expression));
```
lolleko, line 149: Asserts should probably be removed. If there is still a chance the condition can occur it should be handled properly.

language-extensions/index.d.ts
```diff
@@ -1,2 +1,56 @@
 declare function $multi<T extends any[]>(...values: T): MultiReturn<T>;
 declare type MultiReturn<T extends any[]> = T & { readonly " __multiBrand": unique symbol };
+
```
lolleko, line 3: Some nitpicking on the "operator" naming:

I would prefer complete names without abbreviations, I think that would increase readability and since you don't have to type these Identifiers very often, it shouldn't result in a lot of extra type work for the user. You could call them the same as in `ExtensionKind` (https://github.com/TypeScriptToLua/TypeScriptToLua/pull/969/files#diff-32793566a7017d38a85af44995a5400e56cab8d4894d8d545fd586496312d878R4)

e.g `LuaFloorDivision` instead of `LuaIDiv`
`LuaBitwiseLeftShift` instead of `LuaShl` ...

language-extensions/index.d.ts
```diff
@@ -1,2 +1,56 @@
 declare function $multi<T extends any[]>(...values: T): MultiReturn<T>;
 declare type MultiReturn<T extends any[]> = T & { readonly " __multiBrand": unique symbol };
```
lolleko, line 2: If you want to you can rename `MultiReturn` to `LuaMultiReturn` as discussed on discord. But we could also do that in a separate PR.

src/transformation/visitors/language-extensions/operators.ts
tomblind: The only way these would happen is if someone was messing with the extension declarations themselves. But, since TS won't be happy without them, I'll add in some proper diagnostics.

language-extensions/index.d.ts
tomblind: My thought here was that someone writing these declarations would know the meta-method they were mapping to and be looking for that. I don't feel strongly about it though - I can rename them.

language-extensions/index.d.ts
tomblind: I plan to do a PR to fix some issues with MultiReturn, so I figured I'd rename it in that.

test/unit/language-extensions/operators.spec.ts
```diff
@@ -0,0 +1,379 @@
...
+        public __concat(other: OpTest) { return this.value.toString() + other.value.toString(); }
+
+        public __unm() { return -this.value; }
+        public __bnot() { return ~this.value; }
+        public __len() { return this.value; }
+    }
+`;
+
+const binaryMathOperatorTests: Array<[string, number, number, number]> = [
+    ["LuaAdd", 7, 4, 11],
```
Perryvw, line 51: It's not obvious  what this means at first, might be more readable if you use objects instead of tuples, ie `{ extensionType: "LuaAdd", left: 7, right: 4, expectedResult: 11 }`, similar to https://github.com/TypeScriptToLua/TypeScriptToLua/blob/master/test/unit/assignments.spec.ts#L408

language-extensions/index.d.ts
lolleko: I think for most people this will be the first time interacting with Lua operators.  I for example had trouble figuring out what `unm` is supposed to mean, `UnaryNegation` on the other hand I would have understand immediately.

src/transformation/visitors/language-extensions/operators.ts
tomblind: I removed the assertions, but didn't add diagnostics since I can't even create tests for them. That would require somehow modifying the extension declarations for the tests. I think it's safe to let things just crash once users start messing with TSTL internals.

Perryvw (approved):

lolleko (approved):

PR #978, opened 2021-02-07
Multi-Return update

### LINKED ISSUE #968: MultiReturn not wrapped in table when assigned without spread operator
[Playground](https://typescripttolua.github.io/play/#code/GYVwdgxgLglg9mABMOcAUBKRBvAUIgxAJwFMoQikASAWxABtY0AiAQ2YBpFmAjZjANy4AvrlwB6AFSTEAAXIAHeiQBKZCkknjcoSLASIerIphz5CpcpUQBtNp258AukNG4ICAM5READ0QAvDikngxQAFzIqJjCAoji4lFwiDCeiGBwPgDuRKwKCiQAJilIrIhQrDzKAIS4CkQwYFBovgB0IWE2AAxOgvGJngAWcAzFPCTc7GIeYN6IAJ6BwSShjJFGJhix-YbGKWk5eQXFjYhlFVUkdQ1NaPPtK509grhAA)

MultiReturn should either be wrapped in a table like @tupleReturn currently does, or a diagnostic should be produced when it is assigned without using the spread operator.

tomblind:
This turned into a somewhat major refactor on the multi-return stuff, as I ran into more unhandled cases along the way.

Here's the main things this PR does:
- Rename `MultiReturn` to `LuaMultiReturn` for consistency with other lib extensions
- Added doc comments to multi-return stuff in lib extension declarations
- Fixed #968 
- Enabled support for assigning multi-return values without destructuring, by wrapping call in a table
- Fixed broken code generation with assignment expressions and try/catch blocks
- Added a diagnostic for cases where `$multi` could be implicitly cast to something else (such as when the function explicitly defines a non-multireturn return type)

### DIFF
--- a/src/transformation/visitors/language-extensions/multi.ts
+++ b/src/transformation/visitors/language-extensions/multi.ts
@@ -11,11 +11,7 @@ const isMultiTypeDeclaration = (declaration: ts.Declaration): boolean =>
     extensions.getExtensionKind(declaration) === extensions.ExtensionKind.MultiType;
 
 export function isMultiReturnType(type: ts.Type): boolean {
-    return (
-        (type.symbol?.declarations?.some(isMultiTypeDeclaration) ||
-            type.aliasSymbol?.declarations?.some(isMultiTypeDeclaration)) ??
-        false
-    );
+    return type.aliasSymbol?.declarations?.some(isMultiTypeDeclaration) ?? false;
 }
 
 export function isMultiFunction(context: TransformationContext, expression: ts.CallExpression): boolean {
--- a/src/transformation/visitors/language-extensions/multi.ts
+++ b/src/transformation/visitors/language-extensions/multi.ts
@@ -14,7 +14,7 @@ export function isMultiReturnType(type: ts.Type): boolean {
     return type.aliasSymbol?.declarations?.some(isMultiTypeDeclaration) ?? false;
 }
 
-export function isMultiFunction(context: TransformationContext, expression: ts.CallExpression): boolean {
+export function isMultiFunctionCall(context: TransformationContext, expression: ts.CallExpression): boolean {
     const type = context.checker.getTypeAtLocation(expression.expression);
     return type.symbol?.declarations?.some(isMultiFunctionDeclaration) ?? false;
 }
--- a/src/transformation/visitors/language-extensions/multi.ts
+++ b/src/transformation/visitors/language-extensions/multi.ts
@@ -10,6 +10,14 @@ const isMultiFunctionDeclaration = (declaration: ts.Declaration): boolean =>
 const isMultiTypeDeclaration = (declaration: ts.Declaration): boolean =>
     extensions.getExtensionKind(declaration) === extensions.ExtensionKind.MultiType;
 
+export function isMultiReturnType(type: ts.Type): boolean {
+    return (
+        (type.symbol?.declarations?.some(isMultiTypeDeclaration) ||
+            type.aliasSymbol?.declarations?.some(isMultiTypeDeclaration)) ??
+        false
+    );
+}
+
 export function isMultiFunction(context: TransformationContext, expression: ts.CallExpression): boolean {
     const type = context.checker.getTypeAtLocation(expression.expression);
     return type.symbol?.declarations?.some(isMultiFunctionDeclaration) ?? false;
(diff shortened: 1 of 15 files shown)


src/transformation/visitors/return.ts
```diff
@@ -6,13 +6,42 @@ import { validateAssignment } from "../utils/assignment-validation";
...
+import { invalidMultiFunctionReturnType } from "../utils/diagnostics";
 
 function transformExpressionsInReturn(
     context: TransformationContext,
     node: ts.Expression,
     insideTryCatch: boolean
 ): lua.Expression[] {
+    if (ts.isCallExpression(node)) {
+        // $multi(...)
+        if (isMultiFunction(context, node)) {
```
Perryvw, line 25: This naming is a little confusing to me actually, you are checking if the function call isMultiFunction, I would expect node.expression (the name of the function itself) to be the multiFunction here though.

test/unit/language-extensions/__snapshots__/multi.spec.ts.snap
```diff
@@ -54,24 +58,78 @@ exports[`invalid $multi call (({ $multi });): code 1`] = `"local ____ = nil"`;
...
+
+exports[`invalid $multi call (const [a = 0] = $multi()): diagnostics 1`] = `"main.ts(2,25): error TSTL: The $multi function must be called in an expression that is returned."`;
+
+exports[`invalid $multi call (const {} = $multi();): code 1`] = `
+"local ____ = {
+    ____(_G)
+}"
+`;
+
+exports[`invalid $multi call (const {} = $multi();): diagnostics 1`] = `"main.ts(2,20): error TSTL: The $multi function must be called in an expression that is returned."`;
```
Perryvw, line 76: This diagnostic seems unclear, would it be clearer to change it to `The $multi function must be called in a return statement or variable declaration.`?

test/unit/language-extensions/__snapshots__/multi.spec.ts.snap
```diff
@@ -54,24 +58,78 @@ exports[`invalid $multi call (({ $multi });): code 1`] = `"local ____ = nil"`;
...
+exports[`invalid $multi call (const a = $multi();): diagnostics 1`] = `"main.ts(2,19): error TSTL: The $multi function must be called in an expression that is returned."`;
+
+exports[`invalid $multi implicit cast: code 1`] = `
+"function badMulti(self)
+    return \\"foo\\", 42
+end"
+`;
 
-exports[`invalid $multi call (const [a = 0] = $multi()): diagnostics 1`] = `"main.ts(2,15): error TSTL: This array binding pattern cannot have initializers."`;
+exports[`invalid $multi implicit cast: diagnostics 1`] = `"main.ts(3,20): error TSTL: The $multi function can only be used to return a LuaMultiReturn type."`;
```
Perryvw, line 92: Would `The $multi function cannot be cast to a non-LuaMultiReturn type.` be clearer or am I forgetting some cases for this diagnostic here?

src/transformation/visitors/return.ts
tomblind: I agree (this is how it was named already). I think `isMultiFunctionCall` would make more sense.

test/unit/language-extensions/__snapshots__/multi.spec.ts.snap
tomblind: This is what it was previously. $multi can't be called in a variable declaration, so maybe we change it to just `The $multi function must be called in a return statement`.

test/unit/language-extensions/__snapshots__/multi.spec.ts.snap
tomblind: Sounds good to me.

Perryvw (approved):

src/transformation/visitors/language-extensions/multi.ts
```diff
@@ -1,169 +1,92 @@
...
-    context: TransformationContext,
-    declaration: ts.VariableDeclaration
-): lua.Statement[] {
-    assert(declaration.initializer);
-    assert(ts.isCallExpression(declaration.initializer));
-
-    if (!ts.isArrayBindingPattern(declaration.name)) {
-        context.diagnostics.push(invalidMultiTypeToNonArrayBindingPattern(declaration.name));
-        return [];
+export function multiReturnCallShouldBeWrapped(context: TransformationContext, node: ts.CallExpression) {
```
lolleko, line 47: I would prefer
```suggestion
export function shouldMultiReturnCallBeWrapped(context: TransformationContext, node: ts.CallExpression) {
```

But I am also okay with the current version.

lolleko (approved):

Issue #963, opened 2021-01-05
Move annotations to language extensions where possible


Perryvw:
Now that we have language extensions, we can move some of the more hacky annotations to language extensions instead.

Some of the annotations that can be moved:

- `@forRange`
- `@luaIterator`
- `@luaTable`(?)
- `@vararg`

Did I miss any?

@tupleReturn is already moved, should we deprecate the annotation at some point?

lolleko:
Reopened because we still have other candidates for example `@customConstrcutor`

tomblind:
A problem I've found with that proposal for LuaConstructor is that it can't handle the case there an object type has a static factory function (due to circular reference).

[See Playground Example](https://typescripttolua.github.io/play/#code/CYUwxgNghgTiAEAXAngBwQGQK5QKIA9EQA7AZwEsB7YgHgBUAhGKY4eEQk4U+UxGcsQDmAPngBeeAG94AbTrxB8Rs1YBdAFzT4cKMGoRk8APrGIOAkTJViAZWQBbAEaUIWrMXIBHLAlKOXCHgAXxCAbgAoUEhYBBR0eGwoAGFqPhgsMERKGHp2TlYeAAoigDpy2CFSLRZkWTUASgkxWoaxSSLiEAB3eDKKmCqtAAVYKAcQIhhSehEm8TEAJUmsGGI6NBBZpoAyRIsCimoaACJTcxS0-kzsmCYWYBORSKjwaDh4GNIeAFlkZOg33yVm48D+AKg31KXV6Ugi8AR8FQWCcEHIYF4iCgiHR8BhRQAZpRKFp0oIhAAaeBOWBaYhYZwgGANLTgwGkSLBCJgK7wBz-dkSPE9MECyGkIonImUE5UgAsACYGpEgA)

I'm struggling to find a good way to deal with this (via class or interface declarations). So far, the only workable solution I can think of is to provide an optional "rename" type parameter to LuaConstructor.

[...]

PR #979, opened 2021-02-07
$range language extension

tomblind:
This adds the `$range` language extension as a replacement for the `@forRange` annotation, as mentioned in #963

### DIFF
--- a/src/transformation/utils/diagnostics.ts
+++ b/src/transformation/utils/diagnostics.ts
@@ -86,6 +86,14 @@ export const invalidForRangeCall = createErrorDiagnosticFactory(
     (message: string) => `Invalid @forRange call: ${message}.`
 );
 
+export const invalidRangeUse = createErrorDiagnosticFactory(
+    "$range can only be used as a single expression in a for...of loop."
+);
+
+export const invalidRangeControlVariable = createErrorDiagnosticFactory(
+    "For loop using $range must declare a single control variable."
+);
+
 export const luaTableMustBeAmbient = createErrorDiagnosticFactory(
     "Classes with the '@luaTable' annotation must be ambient."
 );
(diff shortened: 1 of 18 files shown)


src/transformation/utils/diagnostics.ts
```diff
@@ -86,6 +86,14 @@ export const invalidForRangeCall = createErrorDiagnosticFactory(
     (message: string) => `Invalid @forRange call: ${message}.`
 );
 
+export const invalidRangeUse = createErrorDiagnosticFactory(
+    "$range can only be used as a single expression in a for...of loop."
```
Perryvw, line 90: Not sure what you mean with `as a single expression`, maybe just remove that part?

test/unit/language-extensions/range.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+		for (i of $range(1, 5)) {}
+	`
+        .setOptions(rangeProjectOptions)
+        .expectDiagnosticsToMatchSnapshot([invalidRangeControlVariable.code]);
+});
+
+test.each([
+    ["for (const i in $range(1, 10, 2)) {}", [invalidRangeUse.code]],
+    ["const x = $range(1, 10);", [invalidRangeUse.code]],
+    ["const range = $range;", [invalidRangeUse.code, invalidRangeUse.code]],
```
Perryvw, line 58: Same code mentioned twice here, could that be reduced to 1? Then is there a need to specify them as part of the each() at all or could you just put it in the test body?

tomblind: This is an odd side effect of how the $range function was being detected. I found a way to do that better and avoid the redundant diagnostic.

lolleko:
I think this should also deprecate `@forRange`.

For that push a `annotationDeprecated ` diagnostic when it's used and add a test to https://github.com/TypeScriptToLua/TypeScriptToLua/blob/master/test/unit/annotations/deprecated.spec.ts

lolleko (approved):

PR #980, opened 2021-02-11
Implemented enum merging

Perryvw:


src/transformation/utils/typescript/nodes.ts
```diff
@@ -23,3 +24,12 @@ export function isInDestructingAssignment(node: ts.Node): boolean {
             (ts.isBinaryExpression(node.parent) && ts.isArrayLiteralExpression(node.parent.left)))
     );
 }
+
+export function getNodeSymbol(context: TransformationContext, node: ts.Node): ts.Symbol | undefined {
```
lolleko, line 28: Maybe add a comment that using getSymbolAtLocation directly is preferred. Since this is a bit hacky after all.

And I would prefer `getSymbolForNode`.

test/unit/enum.spec.ts
```diff
@@ -154,3 +154,33 @@ test("toString", () => {
...
+        }
+
+        enum TestEnum {
+            C = 1,
+            D
+        }
+
+        return ${serializeEnum("TestEnum")}
+    `.expectToMatchJsResult();
+});
```
lolleko, line 186: Add a test for multiple files, **if** you can split enums across files and out test framework supports that.

lolleko (approved):

PR #981, opened 2021-02-12
LuaIterable language extension

tomblind:
Another language extension for #963.

This one replaces `@luaIterator` with a new type: `LuaIterable`. To iterate multiple values, you can use `LuaIterable<LuaMultiReturn<T>>` (similar to `@luaIterator`+`@tupleReturn`).

This type can be returned by functions or properties. It can be used in `for...of` loops or manually by calling the returned function (and passing the appropriate state/value variables for iterators which require them).

I avoided deprecating `@luaIterator` for now, since it is heavily depended on by [lua-types](https://github.com/TypeScriptToLua/lua-types). Once we are able to update the npm package, this will be deprecated along with `@tupleReturn`.

### DIFF
--- a/src/transformation/visitors/language-extensions/operators.ts
+++ b/src/transformation/visitors/language-extensions/operators.ts
@@ -49,7 +49,7 @@ const unaryOperatorMappings = new Map<extensions.ExtensionKind, lua.UnaryOperato
     [extensions.ExtensionKind.LengthOperatorMethodType, lua.SyntaxKind.LengthOperator],
 ]);
 
-const operatorMapExtensions = [...binaryOperatorMappings.keys(), ...unaryOperatorMappings.keys()];
+const operatorMapExtensions = new Set([...binaryOperatorMappings.keys(), ...unaryOperatorMappings.keys()]);
 
 const bitwiseOperatorMapExtensions = new Set<extensions.ExtensionKind>([
     extensions.ExtensionKind.BitwiseAndOperatorType,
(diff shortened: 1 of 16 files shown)


test/unit/language-extensions/iterable.spec.ts
```diff
@@ -0,0 +1,231 @@
...
+import {
+    invalidIterableUse,
+    invalidMultiIterableWithoutDestructuring,
+} from "../../../src/transformation/utils/diagnostics";
+
+const iterableProjectOptions: tstl.CompilerOptions = {
+    types: [path.resolve(__dirname, "../../../language-extensions")],
+};
+
+const testIterable = `
```
Perryvw, line 13: I get you don't want to repeat youself, but this is not very readable in my opinion. An improvement would be to move every iterable definition you use here closer to the tests that actually use them, maybe it might even be nice to group them with `describe`s

test/unit/language-extensions/iterable.spec.ts
```diff
@@ -0,0 +1,231 @@
...
+        ([call, code]): Array<[string, string]> => [
+            [`for (const s in ${call}) {}`, code],
+            [`const i = ${call};`, code],
+            [`function foo(i: any) {} foo(${call});`, code],
+        ]
+    )
+)("invalid use of LuaIterable (%p)", (statement, code) => {
+    util.testFunction`
+        ${code}
+        ${statement}
```
Perryvw, line 215: It is so hard to understand what this test does, wouldn't it be easier to just declare a simple iterator function and call it here instead of the current 3-fold combinations, maybe splitting this up into 3 different tests for the different iterable types.

src/transformation/visitors/language-extensions/operators.ts
```diff
@@ -84,25 +81,15 @@ function getOperatorMapExtensionKindForCall(context: TransformationContext, node
...
+    const type = context.checker.getTypeFromTypeNode(typeDeclaration.type);
+    return operatorMapExtensions.find(extensionKind => extensions.isExtensionType(type, extensionKind));
 }
 
 function isOperatorMapType(context: TransformationContext, type: ts.Type): boolean {
     if (type.isUnionOrIntersection()) {
         return type.types.some(t => isOperatorMapType(context, t));
     } else {
-        return type.symbol?.declarations?.some(isOperatorMapDeclaration);
+        return operatorMapExtensions.some(extensionKind => extensions.isExtensionType(type, extensionKind));
```
Perryvw, line 92: I don't understand why this change is required (in combination with the removal of the Set above). This turns this function from O(1) into O(n) (n being the nr of supported map extensions). Since this function is (I think?) executed for all function and method calls, that seems like a big performance hit I would like to avoid if possible.

tomblind: I agree the increased complexity is not ideal here. The issue comes down to the change in how extensions are identified. Previously it was by name which was easy to directly map to a kind and check for in a set. Now it's by brand which is a differently named property on each type so we have to iterate through the kinds we care about and test for each brand. Let me think about this a bit - there may be a way to refactor things to get the complexity back down.

Perryvw (approved):

PR #991, opened 2021-02-21
Lua table extensions

tomblind:
Next extension for #963 

This adds a set of new extensions to replace functionality provided by `@luaTable`.

It adds `LuaTableGet` & `LuaTableSet` (and method variants) to allow declaring functions which map to `value = table[index]` and `table[index] = value`, similar to how operator overload mappings work.

It also adds a `LuaTable` interface and `LuaTableConstructor` extension to allow tables to be created directly:
```ts
const t = new LuaTable(); // local t = {}
```

### DIFF
--- a/src/transformation/utils/diagnostics.ts
+++ b/src/transformation/utils/diagnostics.ts
@@ -174,6 +174,14 @@ export const invalidOperatorMappingUse = createErrorDiagnosticFactory(
     "This function must always be directly called and cannot be referred to."
 );
 
+export const invalidTableExtensionUse = createErrorDiagnosticFactory(
+    "This function must always be directly called and cannot be referred to."
+);
+
+export const invalidTableSetExpression = createErrorDiagnosticFactory(
+    "Table set extension can only be called as a stand-alone statement. It cannot be used as an expression in another statement."
+);
+
 export const annotationDeprecated = createWarningDiagnosticFactory(
     (kind: AnnotationKind) =>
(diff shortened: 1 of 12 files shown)


test/unit/language-extensions/table.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+    test("table length", () => {
+        util.testFunction`
+            const tbl = new LuaTable<number, string>();
+            tbl.set(1, "foo");
+            tbl.set(3, "bar");
+            return tbl.length();
+        `
+            .setOptions(tableProjectOptions)
+            .expectToEqual(1);
+    });
```
lolleko, line 129: Not sure why someone would do that but could you add a test for `new LuaTable().get/set("test")`. Just to ensure that works and does not produce invalid Lua (`{}["test"]`).

src/transformation/utils/diagnostics.ts
```diff
@@ -174,6 +174,14 @@ export const invalidOperatorMappingUse = createErrorDiagnosticFactory(
     "This function must always be directly called and cannot be referred to."
 );
 
+export const invalidTableExtensionUse = createErrorDiagnosticFactory(
+    "This function must always be directly called and cannot be referred to."
```
lolleko, line 178:
```suggestion
    "This function must be called directly and cannot be referred to."
```

lolleko (approved):

PR #992, opened 2021-02-28
Vararg Update

### LINKED ISSUE #963: Move annotations to language extensions where possible

tomblind:
This PR:
- Deprecates `@vararg` annotation
- Adds automatic vararg optimization, removing the need for a similar language extension
- Adds `$vararg` constant language extension for accessing the global ellipsis operator

Vararg optimization will occur in any case that's viable. It will NOT be optimized:
- Inside of IIFEs generated by TSTL (assignment expressions, etc...)
- After the vararg's identifier has been accessed directly (since the array may have been modified)
- If the vararg's identifier is accessed between the use and definition of a function that will be hoisted (the function may modify the array)

closes #963


src/transformation/visitors/template.ts
```diff
@@ -65,29 +65,35 @@ export const transformTaggedTemplateExpression: FunctionVisitor<ts.TaggedTemplat
 
     // Construct table with strings and literal strings
 
-    const rawStringsTable = lua.createTableExpression(
-        rawStrings.map(text => lua.createTableFieldExpression(lua.createStringLiteral(text)))
+    const rawStringsObject = ts.factory.createArrayLiteralExpression(
```
Perryvw, line 68: `rawStringsArray` would be a better name since it's a TS array, not an object.

src/transformation/visitors/template.ts
```diff
@@ -65,29 +65,35 @@ export const transformTaggedTemplateExpression: FunctionVisitor<ts.TaggedTemplat
...
+        rawStrings.map(text => ts.factory.createStringLiteral(text))
     );
 
-    const stringTableLiteral = lua.createTableExpression([
-        ...strings.map(partialString => lua.createTableFieldExpression(lua.createStringLiteral(partialString))),
-        lua.createTableFieldExpression(rawStringsTable, lua.createStringLiteral("raw")),
+    const stringObject = ts.factory.createObjectLiteralExpression([
```
Perryvw, line 72: Is this really necessary? Shouldnt transforming the array literal already take care of all of this?

Perryvw (commented):
Review to be continued

src/transformation/visitors/spread.ts
```diff
@@ -0,0 +1,70 @@
...
+    context: TransformationContext,
+    symbol: ts.Symbol,
+    identifier: ts.Identifier
+) {
+    if (!ts.isSpreadElement(identifier.parent)) {
+        return false;
+    }
+    // Check for scopes which would stop optimization
+    const scope = findScope(context, ScopeType.Function | ScopeType.Try | ScopeType.Catch | ScopeType.File);
+    if (scope?.type === ScopeType.File && isExtensionValue(context, symbol, ExtensionKind.VarargConstant)) {
```
lolleko, line 23: This if statement and the next one (L26) are really hard to understand, maybe they can be refactored into an extra function?

src/transformation/visitors/binary-expression/compound.ts
```diff
@@ -146,27 +140,39 @@ export function transformCompoundAssignmentExpression(
...
-        return createImmediatelyInvokedFunctionExpression(assignStatements, left, expression);
+        const statements = transformAssignment(context, lhs, operatorExpression);
+        return { statements, result: left };
     }
 }
 
+export function transformCompoundAssignmentExpression(
+    context: TransformationContext,
+    expression: ts.Expression,
+    // TODO: Change type to ts.LeftHandSideExpression?
```
lolleko, line 163: Is this a question or a todo? Either way I think we should do a separate PR where we address all the TS depuration issues. IRRC we have a few places where we use old API.

src/transformation/visitors/spread.ts
```diff
@@ -0,0 +1,70 @@
...
+        return true; // $vararg global constant
+    }
+    if (!scope?.node || !ts.isFunctionLike(scope.node)) {
+        return false;
+    }
+    // Identifier must be a var arg in the local function scope
+    function isSpreadParameter(p: ts.ParameterDeclaration) {
```
lolleko, line 30: Not sure what out policy is about inline functions is, but personally I prefer arrow functions for that.

test/unit/spread.spec.ts
```diff
@@ -128,3 +129,303 @@ describe("in object literal", () => {
...
+describe("vararg spread optimization", () => {
+    test("basic use", () => {
+        util.testFunction`
+            function pick(...args: string[]) { return args[1]; }
+            function test(...args: string[]) {
+                return pick(...args);
+            }
+            return test("a", "b", "c");
+        `
+            .expectLuaToMatchSnapshot()
```
lolleko, line 142: Not really sure, if we need expectLuaToMatchSnapshot and expectToMatchJsResult. I think if the behaviour of JS and Lua matches, it does not matter what Lua output looks like.

test/unit/language-extensions/vararg.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+
+const varargProjectOptions: tstl.CompilerOptions = {
+    types: [path.resolve(__dirname, "../../../language-extensions")],
+};
+
+test.each([
+    'const result = [...$vararg].join("")',
+    'let result: string; { result = [...$vararg].join(""); }',
+    'let result: string; if (true) { result = [...$vararg].join(""); }',
+    'let result: string; do { result = [...$vararg].join(""); } while (false);',
```
Perryvw, line 14: What about a return statement? Think we should be able to test that.

test/unit/spread.spec.ts
Perryvw: You could argue in this case it does matter, since the current unoptimized lua also works, but is not exactly what we want. So we should either expect it to match a snapshot (I'm fine with this), or somehow assert the resulting code indeed has the optimizations.

lolleko: Good Point. I think checking if the generated code does not contain `unpack` could be more explicit. But I am also fine with using a snapshot.

src/transformation/utils/lua-ast.ts
```diff
@@ -73,6 +75,22 @@ export function createImmediatelyInvokedFunctionExpression(
     return lua.createCallExpression(iife, [], tsOriginal);
 }
 
+export interface ImmediatelyInvokedFunctionParameters {
+    statements: lua.Statement | lua.Statement[];
+    result: lua.Expression | lua.Expression[];
+}
+
+export function transformToImmediatelyInvokedFunctionExpression(
```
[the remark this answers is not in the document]
tomblind: Any suggestions on where to move this? Maybe I should make a new file for common transform utils?

src/transformation/visitors/template.ts
tomblind: The object requires both array indices and the "raw" field, so I'm sure there's a way to reuse the array literal to construct this in a single expression.

src/transformation/visitors/binary-expression/compound.ts
tomblind: Should I remove this? It was just copied over from where the code was previously.

test/unit/spread.spec.ts
tomblind: Yeah, this was my thinking - we need to test that the optimization occurs and that it functions correctly. Testing for unpack feels a bit messy. If something breaks it will be clearly visible in the snapshot.

test/unit/language-extensions/vararg.spec.ts
tomblind: Is there a way to do a return statement in the global scope in ts?

src/transformation/visitors/binary-expression/compound.ts
lolleko: Oh, in that case just keep it. Until we figure out what exactly was meant by this.

src/transformation/utils/lua-ast.ts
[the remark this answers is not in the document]
Perryvw: That would be fine with me yeah, I was thinking src/transformation/utils/function-context.ts or maybe .scope.ts but I'm not sure it fits in those, even though it is related. I'd be okay with adding a new file in transformation/utils for this.

src/transformation/visitors/binary-expression/compound.ts
Perryvw: Yeah keep it for now.

src/transformation/utils/symbols.ts
```diff
@@ -44,7 +45,9 @@ export function trackSymbolReference(
         symbolInfo.set(symbolId, { symbol, firstSeenAtPos: identifier.pos });
     }
 
-    markSymbolAsReferencedInCurrentScopes(context, symbolId, identifier);
+    if (!isOptimizableVarArgSpread(context, symbol, identifier)) {
```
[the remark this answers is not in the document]
Perryvw: I'm sorry but I still don't understand this. Could you explain it from the perspective of this `trackSymbolReference`? Why do we not track this as a reference if the identifier is a vararg? Isn't that a reference?

Perryvw (commented):
Could you add the following test case to your spread tests:
```ts
function foo(...args: string[]) {
    return args;
}
```
This should be optimized right?

Other than that looks good to me.

tomblind:
@Perryvw That won't be optimized. The function returns an array so the vararg needs to be wrapped. The only way to do what you're thinking there is using `$multi` (which there is a test for).

PR #997, opened 2021-03-04
Fix destructuring assignment of multiReturn (closes #995)

### LINKED ISSUE #995: Bug when destructuring a result directly.
I reproduce this bug using `string.find`, but maybe it applies to all `@tupleReturn`?

The following code reproduces this error on the tstl playground:
```
// works as it should
const result = string.find("a", "b");
const [c, ...d] = result;

// Doesn't work
const [a, ...b] = string.find("a", "b");
```

Gets translated to
```
result = {
    string.find("a", "b")
}
c = result[1]
d = __TS__ArraySlice(result, 1)
local ____ = string.find("a", "b")
a = ____[1]
b = __TS__ArraySlice(____, 1)
```

On the second case, it uses `string.find` result directly without putting it into a table, which causes a nil error.

yancouto:
Fixed the issue. Turns out we were dealing with this problem on normal assignments, so something like this works:
```js
let a, b;
[a, ...b] = string.find("a", "b");
```
This is handled here: https://github.com/TypeScriptToLua/TypeScriptToLua/blob/master/src/transformation/visitors/binary-expression/assignments.ts#L94-L97

Maybe this is an indicative that this needs some refactoring, as that code should be common. However, I just fixed the issue locally, and added a test showing it is now working.

The only other change is a snapshot of a test that's supposed to fail, so that shouldn't be a problem I believe.

`npm test` works.

### DIFF
--- a/test/unit/language-extensions/multi.spec.ts
+++ b/test/unit/language-extensions/multi.spec.ts
@@ -24,6 +24,19 @@ test("multi example use case", () => {
         .expectToEqual({ a: "foo", b: 5 });
 });
 
+test("multi #995", () => {
+    util.testModule`
+        function multiReturn(): LuaMultiReturn<[number, number, number]> {
+            return $multi(1, 2, 3);
+        }
+
+        const [a, ...b] = multiReturn();
+        export {a, b};
+    `
+        .setOptions(multiProjectOptions)
+        .expectToEqual({ a: 1, b: [2, 3] });
+});
+
 test.each<[string, any]>([
     ["$multi()", undefined],
     ["$multi(true)", true],
--- a/test/unit/language-extensions/multi.spec.ts
+++ b/test/unit/language-extensions/multi.spec.ts
@@ -24,7 +24,8 @@ test("multi example use case", () => {
         .expectToEqual({ a: "foo", b: 5 });
 });
 
-test("multi #995", () => {
+// https://github.com/TypeScriptToLua/TypeScriptToLua/issues/995
+test("Destructuring assignment of LuaMultiReturn", () => {
     util.testModule`
         function multiReturn(): LuaMultiReturn<[number, number, number]> {
             return $multi(1, 2, 3);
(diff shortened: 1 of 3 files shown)


test/unit/language-extensions/multi.spec.ts
```diff
@@ -24,6 +24,19 @@ test("multi example use case", () => {
         .expectToEqual({ a: "foo", b: 5 });
 });
 
+test("multi #995", () => {
+    util.testModule`
+        function multiReturn(): LuaMultiReturn<[number, number, number]> {
+            return $multi(1, 2, 3);
```
Perryvw, line 30: Could you add another test where this function returns nil? (like the string.find in the issue)

test/unit/language-extensions/multi.spec.ts
```diff
@@ -24,6 +24,19 @@ test("multi example use case", () => {
         .expectToEqual({ a: "foo", b: 5 });
 });
 
+test("multi #995", () => {
```
Perryvw, line 27: Our convention is to give the test a descriptive name, i.e `test("destructure LuaMultiReturn with rest"` and then add a comment with the PR link. (for example see https://github.com/TypeScriptToLua/TypeScriptToLua/blob/master/test/unit/builtins/object.spec.ts#L232 )

Perryvw (commented):
Thanks for the PR, looks good! I only have some minor comments.

test/unit/language-extensions/multi.spec.ts
```diff
@@ -24,6 +24,30 @@ test("multi example use case", () => {
...
+    util.testModule`
+        function multiReturn(): LuaMultiReturn<[number, number, number]> {
+            return;
+        }
+
+        const [a, ...b] = multiReturn();
+        export {a, b};
+    `
+        .setOptions(multiProjectOptions)
+        .expectToEqual({ b: [] });
```
Perryvw, line 48: Could you also check the value of `a` here? (should be undefined)

test/unit/language-extensions/multi.spec.ts
```diff
@@ -24,6 +24,30 @@ test("multi example use case", () => {
...
+        function multiReturn(): LuaMultiReturn<[number, number, number]> {
+            return $multi(1, 2, 3);
+        }
+
+        const [a, ...b] = multiReturn();
+        export {a, b};
+    `
+        .setOptions(multiProjectOptions)
+        .expectToEqual({ a: 1, b: [2, 3] });
+    util.testModule`
```
Perryvw, line 39: Tests should only test one thing, could you split this second testModule into a second `test("Destructuring assignment of LuaMultiReturn returning nil", () => {`

Perryvw (approved):

Issue #953, opened 2020-12-22
Remove deprecated annotations from transpiler


Perryvw:
Continuation of #652. A warning was added to warn users no longer to use these compiler directives, at some point in the future we should actually remove these directives from our code base.

It might be a good idea to choose a set date to aim for end-of-life on these, something like May 1?

PR #998, opened 2021-03-05
Remove phantom, metaExtension, extension, pureAbstract

lolleko:
And replace the deprecation warning with an error

Progress on #953


Perryvw (approved):
👌

PR #1001, opened 2021-03-06
Add actual multi version tests

lolleko:
Tests now select the appropriate Lua runtime based on luaTarget option
This works for all supported Lua versions except JIT

### DIFF
--- a/test/util.ts
+++ b/test/util.ts
@@ -29,6 +29,9 @@ function getLuaBindingsForVersion(target: tstl.LuaTarget): { lauxlib: LauxLib; l
         const { lauxlib, lua, lualib } = require("lua-wasm-bindings/dist/lua.53");
         return { lauxlib, lua, lualib };
     }
+    if (target === tstl.LuaTarget.LuaJIT) {
+        throw Error("Can't use executeLua() or expectToMatchJsResult() wit LuaJIT as target!");
+    }
 
     const { lauxlib, lua, lualib } = require("lua-wasm-bindings/dist/lua.54");
     return { lauxlib, lua, lualib };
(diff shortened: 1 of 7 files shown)


test/util.ts
```diff
@@ -17,6 +16,24 @@ const luaLib = fs.readFileSync(path.resolve(__dirname, "../dist/lualib/lualib_bu
...
+    if (target === tstl.LuaTarget.Lua52) {
+        const { lauxlib, lua, lualib } = require("lua-wasm-bindings/dist/lua.52");
+        return { lauxlib, lua, lualib };
+    }
+    if (target === tstl.LuaTarget.Lua53) {
+        const { lauxlib, lua, lualib } = require("lua-wasm-bindings/dist/lua.53");
+        return { lauxlib, lua, lualib };
+    }
+
+    const { lauxlib, lua, lualib } = require("lua-wasm-bindings/dist/lua.54");
```
Perryvw, line 33: What if target is LuaJIT?

lolleko: Mhh we can use 5.1 or 5.2 for JIT. But not sure that will help much. Most of our version specific tests are snapshot tests or test for exceptions. So no Lua code is actually executed in those.

Perryvw: I would be okay by simply erroring for this case with a message we cannot run JIT in tests.

lolleko: But we can run JIT tests as long as they don't use any JIT specific functions or features that are only available since 5.2.
Not sure however what the point is of testing non JIT functionality with JIT as target. So I guess erroring is fine?


Throwing for JIT breaks the following 3 test cases:
https://github.com/TypeScriptToLua/TypeScriptToLua/blob/master/test/unit/conditionals.spec.ts#L371
Any idea why we test that for JIT specifically?

lolleko: Added the throw in a3511622df77eeb5b49565182227dd12a405edf1

PR #1008, opened 2021-03-25
Replaced deprecated factory functions

lolleko:
Also enabled some eslint rules & fixed violations

### DIFF
--- a/src/LuaLib.ts
+++ b/src/LuaLib.ts
@@ -87,6 +87,7 @@ export enum LuaLibFeature {
     Unpack = "Unpack",
 }
 
+/* eslint-disable @typescript-eslint/naming-convention */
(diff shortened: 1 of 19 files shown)


src/LuaLib.ts
```diff
@@ -87,6 +87,7 @@ export enum LuaLibFeature {
     Unpack = "Unpack",
 }
 
+/* eslint-disable @typescript-eslint/naming-convention */
```
Perryvw, line 90: Why is it complaining here, because of the key names?

Perryvw (approved):

src/LuaLib.ts
lolleko: Yes PascalCase key names are not allowed. I didn't change them to camelCase because I wasn't sure what the implications of that would be.

PR #1011, opened 2021-04-17
Module resolution from node_modules

### LINKED ISSUE #432: New module resolution and Compilation structure
Continuation of #270.

- Remove everything related to module resolving from Transformer
- Add a new class - `Compilation`, that would accept all files compiled with transformer, process all of them with a regular expression, find require calls and resolve them.
	- All resolved .lua files would be loaded as well
	- Require class would be replaced with real resolved paths.
	- Since all .lua files would get loaded, that would be a step towards resolving #357
- Add a new option that would control how `Compilation`. If it's in lib mode it won't transform require paths (except for resolution of "paths" tsconfig options). If it's in app mode it would output dependencies to `node_modules` directory of build target. That would also allow to reuse npm resolution technique to avoid version conflicts.

### LINKED ISSUE #751: Transpiling a project from another directory  will incorrectly resolve require paths
i.e running having a directory 'project' with tsconfig and file:
```ts
import * as foo from "./bar/baz";
```

And then running `transpileProject("project/tsconfig.json")` from the parent directory (of the project dir), will result in `require("project.bar.baz")` because Emit falls back on `process.cwd()` instead of the project (tsconfig) dir: https://github.com/TypeScriptToLua/TypeScriptToLua/blob/master/src/Emit.ts#L24

### LINKED ISSUE #823: Broken require paths in bundles
If a TSTL bundled project is built from outside the project directory, the require paths will be broken.

Example:
```
tstltest/
  tsconfig.json
  foo.ts
  bar.ts
```

```json
// tsconfig.json
{
    "include": [
        "./*.ts"
    ],
    "tstl":
    {
        "noHeader": true,
        "luaBundle": "bundle.lua",
        "luaBundleEntry": "foo.ts"
    }
}
```

```ts
// foo.ts
import {Bar} from "./bar";
console.log(Bar.bar);
```

```ts
// bar.ts
export namespace Bar {
    export const bar = "bar";
}
```

```lua
-- bundle.lua, if built from parent directory ('tstl -p tstltest')
local ____modules = {}
local ____moduleCache = {}
local ____originalRequire = require
local function require(file)
    if ____moduleCache[file] then
        return ____moduleCache[file]
    end
    if ____modules[file] then
        ____moduleCache[file] = ____modules[file]()
        return ____moduleCache[file]
    else
        if ____originalRequire then
            return ____originalRequire(file)
        else
            error("module '" .. file .. "' not found")
        end
    end
end
____modules = {
["bar"] = function() local ____exports = {}
____exports.Bar = {}
local Bar = ____exports.Bar
do
    Bar.bar = "bar"
end
return ____exports
end,
["foo"] = function() local ____exports = {}
local ____bar = require("tstltest.bar") -- <----- this will fail
local Bar = ____bar.Bar
print(Bar.bar)
return ____exports
end,
}
return require("foo")
```

This can be worked around right now with a `"rootDir": "."` in the tsconfig.json, but requiring files that are part of the bundle should always work, regardless of the build environment.

### LINKED ISSUE #976: Bug: packages with index.ts
Either code generator or custom `require` in lua works incorrectly:

When importing `import * from './pkg` if `pkg` is a directory with `index.ts` tstl generates module for lua named `pkg.index` and generates import referencing just `pkg` without `.index`.

Sample:

`main.ts`:
```typescript
import { hello } from './pkg';

print(hello);
```

`pkg/index.ts`:
```typescript
export const hello = 'hello';
```

results in:

```lua
local ____modules = {}
local ____moduleCache = {}
local ____originalRequire = require
local function require(file)
    if ____moduleCache[file] then
        return ____moduleCache[file]
    end
    if ____modules[file] then
        ____moduleCache[file] = ____modules[file]()
        return ____moduleCache[file]
    else
        if ____originalRequire then
            return ____originalRequire(file)
        else
            error("module '" .. file .. "' not found")
        end
    end
end
____modules = {
["pkg.index"] = function() local ____exports = {}
____exports.hello = "hello"
return ____exports
end,
["main"] = function() local ____exports = {}
local ____pkg = require("pkg")
local hello = ____pkg.hello
print(hello)
return ____exports
end,
}
return require("main")
```

tstl version: `0.37.1`

### LINKED ISSUE #1005: Add generated file header to bundle same as regular transpiled files
Greetings,

In the configuration management world (e.g. Puppet, Chef), it is best practice for the tool to append a comment to the beginning of the file.

Something like:

```bash
# DO NOT EDIT THIS SCRIPT - THIS SCRIPT IS MANAGED BY PUPPET AND ANY CHANGES WILL BE OVERWRITTEN
mkdir foo1
mkdir foo2
mkdir foo3
```

Currently, this is the top 5 lines of my "main.lua" file that was generated by TypeScriptToLua:

```lua

local ____modules = {}
local ____moduleCache = {}
local ____originalRequire = require
local function require(file)
```

I propose that TypeScriptToLua should do the same thing for the lua files it outputs:

```lua
--[[

This file was automatically generated with TypeScriptToLua.
https://typescripttolua.github.io/
DO NOT EDIT THIS FILE, as any changes will be overwritten.

--]]

local ____modules = {}
local ____moduleCache = {}
local ____originalRequire = require
local function require(file)
```

I think this would be valuable for various reasons. For example, someone just submitted a pull request to one of my projects that edited the generated Lua file instead of the actual TypeScript files, and maybe this header would have prevented that.

Thoughts?

Perryvw:
Add extra resolve step when emitting to resolve Lua files that might be referenced but not part of the project output. This includes Lua files placed independently in the source directory, and also lua files in `node_modules`.

Upon resolution, any `require()` calls will be resolved to search for the lua file they are including. If this file is not yet part of the output it is added. Require paths are rewritten so the lua output and requires match.

Lua files from `node_modules` are placed into `lua_modules` to avoid overwriting them when rewriting require paths.

To create libraries, a buildMode config setting is added,. If this setting is set to "library", files from external packages in node_modules are not included, and their requires are not rewritten. This will be done once the package is included and used in a client project.

Closes #909, closes #1005, fixes #976, closes, fixes #823, fixes #751, closes #432

Documentation to go with this functionality:
https://github.com/TypeScriptToLua/TypeScriptToLua.github.io/pull/44


src/transpilation/diagnostics.ts
```diff
@@ -1,8 +1,20 @@
...
-const createDiagnosticFactory = <TArgs extends any[]>(getMessage: (...args: TArgs) => string) =>
-    createSerialDiagnosticFactory((...args: TArgs) => ({ messageText: getMessage(...args) }));
+const createDiagnosticFactory = <TArgs extends any[]>(
+    getMessage: (...args: TArgs) => string,
+    category: ts.DiagnosticCategory = ts.DiagnosticCategory.Error
+) => createSerialDiagnosticFactory((...args: TArgs) => ({ messageText: getMessage(...args), category }));
+
+export const couldNotResolveRequire = createDiagnosticFactory(
+    (require: string, containingFile: string) =>
+        `Could not resolve require path '${require}' in file ${containingFile}.`,
```
lolleko, line 11:
```suggestion
    (requirePath: string, containingFile: string) =>
        `Could not resolve require path '${requirePath}' in file ${containingFile}.`,
```

src/transpilation/resolve.ts
```diff
@@ -0,0 +1,206 @@
...
+    useSyncFileSystemCalls: true,
+});
+
+interface ResolutionResult {
+    resolvedFiles: ProcessedFile[];
+    diagnostics: ts.Diagnostic[];
+}
+
+// Cache for getting source files from the program
+const projectFileCache = new Set<string>();
```
lolleko, line 24: Is there any good way to avoid having this in global/module scope? maybe store it in the transformation context?

src/transpilation/resolve.ts
```diff
@@ -0,0 +1,206 @@
...
+            // Could not resolve dependency, add a diagnostic and make some fallback path
+            diagnostics.push(couldNotResolveRequire(required, path.relative(projectRootDir, file.fileName)));
+
+            const fallbackRequire = fallbackResolve(required, projectRootDir, fileDir);
+            replaceRequireInCode(file, required, fallbackRequire);
+            replaceRequireInSourceMap(file, required, fallbackRequire);
+        }
+    }
+    return { resolvedFiles: dependencies, diagnostics };
+}
```
lolleko, line 110: This function is a bit long and contains to many nested conditionals. Would be nice to split it into multiple functions, that might also make some of the comments obsolete.

src/transpilation/resolve.ts
```diff
@@ -0,0 +1,206 @@
...
+    // eslint-disable-next-line @typescript-eslint/ban-types
+    let match: RegExpExecArray | null;
+    while ((match = pattern.exec(code))) {
+        paths.push(match[1]);
+    }
+
+    return paths;
+}
+
+function resolveDependency(
```
lolleko, line 154: The only place where this function is used is inside resolveFileDependencies (https://github.com/TypeScriptToLua/TypeScriptToLua/pull/1011/files#diff-d2c750f93b463ac383b3f4a3127e379381ce0e0ff1a5934c12bd2216f84d4125R53) Therefore I think this function should be moved closer/below resolveFileDependencies, to improve readability.

src/transpilation/resolve.ts
```diff
@@ -0,0 +1,206 @@
...
+    fileDirectory: string,
+    rootDirectory: string,
+    dependency: string,
+    emitHost: EmitHost
+): string | undefined {
+    // Check if file is a file in the project
+    const resolvedPath = path.resolve(fileDirectory, dependency);
+
+    if (isProjectFile(resolvedPath)) {
```
lolleko, line 163: Is JSON the only file type that can be resolved directly? In that case I would prefer something more explicit like this instead of the comment
```suggestion
    if (path.extension(resolvedPath).toLoweCase() == ".json" && isProjectFile(resolvedPath)) {
```

test/transpile/module-resolution.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+    });
+
+    test("can resolve dependencies and bundle files with sourceDir", () => {
+        const mainFile = path.join(projectPath, "src", "main.ts");
+        util.testProject(path.join(projectPath, "tsconfig.json"))
+            .setMainFileName(mainFile)
+            .setOptions({ luaBundle: "bundle.lua", luaBundleEntry: mainFile })
+            .expectToEqual(expectedResult);
+    });
+});
```
lolleko, line 165: Maybe add a few tests, with filenames that have uppercase chars as well.

test/transpile/module-resolution/project-with-dependency-chain/tsconfig.json
```diff
@@ -0,0 +1,12 @@
...
+        "strict": true,
+        "moduleResolution": "Node",
+        "noUnusedLocals": true,
+        "noUnusedParameters": true,
+        "target": "esnext",
+        "lib": ["esnext"],
+        "types": [],
+        "rootDir": "."
+    }
+}
```
lolleko, line 12: Maybe try another test project that uses an `module` option other than `commonJS` (e.g. `esnext)`.
Honestly not sure if that will even works with our current import/export implementation, but it would be nice to knowing any case, so we can document that limitation.

src/transpilation/transpiler.ts
```diff
@@ -53,28 +55,79 @@ export class Transpiler {
...
+    const relativeOutputPath = getEmitPathRelativeToOutDir(file, program);
+    const outDir = getEmitOutDir(program);
+
+    return path.join(outDir, relativeOutputPath);
+}
+
+export function getEmitPathRelativeToOutDir(fileName: string, program: ts.Program): string {
+    const sourceDir = getSourceDir(program);
+    // Default output path is relative path in source dir
+    let emitPath = path.relative(sourceDir, fileName).split(path.sep);
```
lolleko, line 96:
```suggestion
    let emitPathSplits = path.relative(sourceDir, fileName).split(path.sep);
```
?

lolleko (changes requested):

src/transpilation/resolve.ts
Perryvw: Removed this completely

src/transpilation/resolve.ts
Perryvw: I got it as short as I could now, it's still quite long but any further separation just hurts readability imo.

test/transpile/module-resolution.spec.ts
Perryvw: the lua source test has file names with upper case names in it. I did some further tests and everything seems to work  fine, adding more tests seems useless. I even tried some cases where the module name cases mismatched the files on disk. This worked fine on windows (paths are case insensitive) and failed on linux, but on linux TS will also fail to resolve these modules, giving you an error.

lolleko (approved):

PR #1014, opened 2021-04-26
Fix namespace and module resulting in invalid Lua when modified with TS transformer plugin

### LINKED ISSUE #1013: Exported namespace transformed by a TS transformer plugin generates invalid Lua
TS:
```ts
export namespace ns {
    export function returnsBool() {
        return false;
    }
}
```
Produces the following incorrect Lua, **ONLY** if changing an AST node with a TS transformer plugin. For example rewriting false to true:
```lua
local ____exports = {}
do
    function ns.returnsBool(self)
        return true
    end
end
return ____exports
```
Resulting in failure:
`attempt to index a nil value (global 'ns')`

Probably caused by a similar issue as solved in #1010

Perryvw:
Fixes #1013


test/transpile/plugins/transformer-plugin.js
```diff
@@ -0,0 +1,18 @@
+/**
```
lolleko, line 1: Where is the `.ts` source for this file?

Perryvw: I did not add TS source as it would mean adding another step to our test or test run to compile it. We could do that at some point if we have more code like this, but for now I don't know if it would add much.

lolleko (approved):

PR #1031, opened 2021-06-12
Implement array.entries()

### LINKED ISSUE #1028: Implement array.entries
Hi, this is a feature request for `array.entires()`. (The TSTL VSCode extension currently reports that it is unimplemented.)

I think this would be relatively quick to add!

Perryvw:
Closes #1028

### DIFF
--- a/src/lualib/ArrayEntries.ts
+++ b/src/lualib/ArrayEntries.ts
@@ -1,0 +1,14 @@
+// https://262.ecma-international.org/10.0/#sec-array.prototype.entries
+function __TS__ArrayEntries<T>(this: void, array: T[]): IterableIterator<[number, T]> {
+    let key = 0;
+    return {
+        [Symbol.iterator](): IterableIterator<[number, T]> {
+            return this;
+        },
+        next(): IteratorResult<[number, T]> {
+            const result = { done: array[key] === undefined, value: [key, array[key]] as [number, T] };
+            key++;
+            return result;
+        },
+    };
+}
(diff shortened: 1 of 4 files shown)


src/lualib/ArrayEntries.ts
```diff
@@ -0,0 +1,14 @@
...
+        [Symbol.iterator](): IterableIterator<[number, T]> {
+            return this;
+        },
+        next(): IteratorResult<[number, T]> {
+            const result = { done: array[key] === undefined, value: [key, array[key]] as [number, T] };
+            key++;
+            return result;
+        },
+    };
+}
```
tomblind, line 14: Why not just this?
```ts
function __TS__ArrayEntries<T>(this: void, array: T[]) {
    return ipairs(array);
}
```

Perryvw: Because TS defines the type of entries as returning `IterableIterator<[number, T]>`, and I don't think ipairs `LuaIterable<LuaMultiReturn<[number, NonNullable<T>]>>` is 1:1 assignable to that since it's a different iterator kind and also the lack of LuaMultiReturn.

tomblind: Ah yeah, that would be a problem if used indirectly. It's a shame since this version is so much more complex.

It might be good to add a test for indirect use:
```
...
const entries = array.entries();
for (const [i, v] of entries) {
...
```

Perryvw: Good call, added a test for indirect use as well as destructuring, those both fail if ipairs is used

tomblind (approved):

PR #1032, opened 2021-06-12
Remove deprecated annotations

### LINKED ISSUE #953: Remove deprecated annotations from transpiler

lolleko:
Removed the following annotations:
- luaIterator
- vararg
- luaTable
- forRange

Closes #953 we should open a new issue if we decide to remove more annotations.


Perryvw (commented):
Looks great, nice simplification. Could you make sure the docs on these removed functions mention how to upgrade?

lolleko:
Docs have been updated to reflect these changes: https://github.com/TypeScriptToLua/TypeScriptToLua.github.io/pull/48

PR #1041, opened 2021-07-04
Optional chaining

Perryvw:
Added support for optional chaining.

I tried to come up with a good test suite but I feel like I might still be missing some stuff, feel free to suggest extra tests.

Also removed a stray .debug() I left in the module-resolution tests, as well as fixing an undiscovered quadratic (actually worse) complexity bug with `??`.


test/unit/optionalChaining.spec.ts
```diff
@@ -1,9 +1,130 @@
...
+
+test.each(["[1, 2, 3, 4]", "undefined"])("optional array access (%p)", value => {
+    util.testFunction`
+        const arr: number[] | undefined = ${value};
+        return arr?.[2];
+    `.expectToMatchJsResult();
+});
+
+test("optional array access nested", () => {
+    util.testFunction`
```
tomblind, line 42: It might be good to have additional test cases here with undefined array elements.

test/unit/optionalChaining.spec.ts
```diff
@@ -1,9 +1,130 @@
...
+        const arr: [number, [number, [number, [number, number]]]] = [1, [2, [3, [4, 5]]]];
+        return arr[1]?.[1][1]?.[0];
+    `.expectToMatchJsResult();
+});
+
+test("optional nested access properties", () => {
+    util.testFunction`
```
tomblind, line 49: Also here: would be good to have test cases that would produce "not found".

src/transformation/visitors/binary-expression/index.ts
```diff
@@ -125,6 +125,12 @@ export const transformBinaryExpression: FunctionVisitor<ts.BinaryExpression> = (
             );
         }
 
+        case ts.SyntaxKind.QuestionQuestionToken: {
+            // Even though this is also handled in transformBinaryOperation, also handle it here
+            // to avoid transforming node.left and node.right twice (unexpected quadratic complexity)
+            return transformNullishCoalescingExpression(context, node);
```
tomblind, line 131: Should we still handle this in `transformBinaryOperation`? We might want to remove that or throw an error there in case it's still happening somehow.

Perryvw: I tried that, but unfortunately `transformBinaryOperation is used by `transformCompoundAssignment` which can get `??=`

test/unit/optionalChaining.spec.ts
Perryvw: The second element in the returned array should be not found as it uses [2] that doesn't exist

Perryvw: I added some more :)

test/unit/optionalChaining.spec.ts
```diff
@@ -1,9 +1,130 @@
...
+
+    test("optional method of optional method result", () => {
+        util.testFunction`
+            const obj: { a?: () => {b: {c?: () => number }}} = {};
+            const obj2: { a?: () => {b: {c?: () => number }}} = { a: () => ({b: {}})};
+            const obj3: { a?: () => {b: {c?: () => number }}} = { a: () => ({b: { c: () => 5 }})};
+
+            return [obj.a?.().b.c?.() ?? "nil", obj2.a?.().b.c?.() ?? "nil", obj3.a?.().b.c?.() ?? "nil"];
+        `.expectToMatchJsResult();
+    });
```
tomblind, line 132: Another one I thought of: element access method calls
```ts
const obj: { value: string; foo?(): string; bar?(): string; } = {
    value: "foobar",
    foo() { return this.value; }
}
const fooKey = "foo";
const barKey = "bar";
return obj[barKey]?.() ?? obj[fooKey]?.();
```

src/transformation/visitors/binary-expression/index.ts
Perryvw: No, the duplicate transforms are line 137 and 138 in this file (`context.transformExpression(node.left)` and same for node.right), that are ignored and repeated in transformBinaryOperation, only in case of `??` afaik.

tomblind: What I mean is, it looks like using `??=` calls `transformCompoundAssignment`, which transforms the left and right expressions and then calls `transformBinaryOperation` which then calls `transformNullishCoalescingExpression` which re-transforms those expressions again. It seems like we should also handle the redundancy in this case.

tomblind (approved):

PR #1049, opened 2021-07-20
Add support for Promises

Perryvw:
**Out of scope:** async/await, will be done in follow-up PR

Implemented promises based on https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise and a little bit of https://promisesaplus.com/

Note: For now we are deviating from the spec in an important aspect: currently the promise constructor function is executed immediately, instead of waiting until there is no more user code on stack, like the promise spec mandates.


Perryvw:
Woops, putting this back into draft after realising I forgot the case where `then()` is called on an already-resolved promise.

Perryvw:
> Woops, putting this back into draft after realising I forgot the case where `then()` is called on an already-resolved promise.

Problem solved.

src/lualib/PromiseAll.ts
```diff
@@ -0,0 +1,53 @@
...
+        i++;
+    }
+
+    // If there are no remaining pending promises, return a resolved promise with the results
+    if (numToResolve === 0) {
+        return Promise.resolve(results);
+    }
+
+    return new Promise((resolve, reject) => {
+        for (const [index, promise] of pairs(toResolve)) {
```
lolleko, line 35: Maybe replace with ipairs to preserve order? Not sure if that's mandated by the specs.

src/lualib/Promise.ts
```diff
@@ -0,0 +1,185 @@
...
+    public static reject(this: void, reason: any): Promise<never> {
+        // Create and return a promise instance that is already rejected
+        const promise = new __TS__Promise<never>(() => {});
+        promise.state = __TS__PromiseState.Rejected;
+        promise.rejectionReason = reason;
+        return promise;
+    }
+
+    constructor(executor: (resolve: (data: T) => void, reject: (reason: any) => void) => void) {
+        executor(this.resolve.bind(this), this.reject.bind(this));
```
lolleko, line 60: This should be wrapped in a try catch if the executor fails the promise should be rejected.
```suggestion
try {
        executor(this.resolve.bind(this), this.reject.bind(this));
} catch (error) {
        reject(error)
}
```

lolleko: Probably also good idea to add a test for that, haven't looked at those yet.

Perryvw: Good catch, I missed this

src/lualib/PromiseAll.ts
Perryvw: toResolve is not an array so `ipairs` is not appropriate

test/unit/builtins/promise.spec.ts
```diff
@@ -0,0 +1,1022 @@
...
+            requests.set(id, callback);
+        }
+        function emulateRequestReturn(id: number, data: UserData) {
+            requests.get(id)!(data);
+        }
+    `;
+
+    util.testFunction`
+        // Wrap function getUserData(id: number, callback: (userData: UserData) => void) into a Promise
+        async function getUserDataAsync(id: number): Promise<UserData> {
```
tomblind, line 516: dumb question: how does this work if we don't support async yet?

tomblind (commented):
One thing I don't see represented in the tests is direct chaining:
`promise.then(...).then(...).catch(...)`
It might be good to test this to ensure these calls are returning the correct thing.

test/unit/builtins/promise.spec.ts
Perryvw: It works because async keyword has no special function yet and will just be ignored. You are right that this is a a strange async, and in fact I think it is removed again in my async/await branch. I will remove it here too.

lolleko (approved):

tomblind (approved):

PR #1052, opened 2021-07-29
JSX support

### LINKED ISSUE #597: JSX support
There should be some Lua environments with UI features where it would be useful. A simple solution would be just to use `transformJsx`, but it may have issues with factory function context.

GlassBricks:
Provides basic JSX support (#597).
 - Only `jsx: "react"` compiler option is supported.
 - The `reactNamespace` compiler option is ignored; it is deprecated anyways (use `jsxFactory` option instead).
 - ~~JSX pragma comments is _not_ (yet) supported. This could be implemented in the future using annotations.~~
 JSX pragma comments support added in second commit.

This implementation uses typescript's jsx transformation as reference and for tests. This means that this implementation is _different_ from some exisiting jsx efforts for tstl in some areas:
 - The the "children" parameter passed to `createElement` should be accepted as a rest (`...`) parameter, not an array.
 - Comments inside jsx elements are omitted, instead of translated to "nil".
 - Text in jsx elements is handled differently: whitespace in multiline strings are trimmed properly, and escaped entities are allowed.

These should probably be documented if released.

Some jsx text processing implementations were copied from [sucrase](https://github.com/alangpierce/sucrase), which is released under the MIT liscence.

Closes #597


src/transformation/visitors/literal.ts
```diff
@@ -62,78 +63,99 @@ const transformNumericLiteralExpression: FunctionVisitor<ts.NumericLiteral> = ex
...
-const transformObjectLiteralExpression: FunctionVisitor<ts.ObjectLiteralExpression> = (expression, context) => {
-    const violations = findMultiAssignmentViolations(context, expression);
-    if (violations.length > 0) {
-        context.diagnostics.push(...violations.map(e => invalidMultiFunctionUse(e)));
-        return lua.createNilLiteral(expression);
-    }
-
-    let properties: lua.TableFieldExpression[] = [];
-    const tableExpressions: lua.Expression[] = [];
+const transformObjectLiteralExpressionOrJsxAttributes: FunctionVisitor<ts.ObjectLiteralExpression | ts.JsxAttributes> =
```
GlassBricks, line 66: The git diff here really got confused due to just one missing newline. Is this worth changing?

Perryvw: Seems like this is prettier adding another indent due to the formatting after the first line became so much longer. I don't really mind as long as Prettier is happy, but it might be nicer not to touch all of this code for our git history/blames. You could try to somehow get prettier to format it differently, or maybe by aliasing the type for example.

Like I said, as long as prettier is happy I don't really mind.

src/transformation/visitors/jsx/jsx.ts
```diff
@@ -0,0 +1,252 @@
...
+    defaultValue: string
+): lua.Expression {
+    const annotation = getFileAnnotations(node.getSourceFile()).get(fileAnnotation);
+    const [first, second] = (annotation?.args[0] ?? option ?? defaultValue).split(".");
+    return second === undefined
+        ? lua.createIdentifier(first)
+        : lua.createTableIndexExpression(lua.createIdentifier(first), lua.createStringLiteral(second));
+}
+
+/* Implementations of jsx text processing modified from sucrase (https://github.com/alangpierce/sucrase). */
```
lolleko, line 30: Please include the full licence of sucrase here in a comment.

src/transformation/visitors/jsx/jsx.ts
```diff
@@ -0,0 +1,252 @@
...
+/**
+ * Turn the given jsxText string into a JS string literal. Leading and trailing
+ * whitespace on lines is removed, except immediately after the open-tag and
+ * before the close-tag. Empty lines are completely removed, and spaces are
+ * added between lines after that.
+ *
+ * We use JSON.stringify to introduce escape characters as necessary, and trim
+ * the start and end of each line and remove blank lines.
+ */
+function formatJSXTextLiteral(text: string): string {
```
Perryvw, line 44: I really do not like the prospect of having to maintain this code, does TS not already do this?

src/transformation/visitors/jsx/xhtml.ts
```diff
@@ -0,0 +1,257 @@
+/* eslint-disable @typescript-eslint/naming-convention */
+const entities: { [name: string]: string } = {
```
Perryvw, line 2: Why not just do this? Instead of declaring the export separately? (I think you could also drop the eslint disable?)
```suggestion
export const XHTMLEntities: { [name: string]: string } = {
```

src/transformation/visitors/literal.ts
GlassBricks: I think I'll leave it as is, since I don't want to fight with prettier.

src/transformation/visitors/jsx/jsx.ts
GlassBricks: TS does do this, but it's buried under several layers of internal code.

I adapted this implementation from sucrase instead (a babel-like processor), as it has a more straightforward implementation.

src/transformation/visitors/jsx/jsx.ts
GlassBricks: sucrase uses the MIT licence, and I copied some portions of their implementation.
How would I properly attribute such code? Still not too familiar with the particulars of OSS licensing.

GlassBricks: Full text of license was added.

lolleko: The correct way would be to add the license to our binary as well e.g. to `tstl -h`. But adding the full text or a permalink in code is good enough I think. That's what we've done in other places where third party code was used. As long as no one complains this should be fine.

thanks for updating.

Perryvw (approved):

PR #1087, opened 2021-08-15
fix sourceMapTraceback

### LINKED ISSUE #1086: Sourcemap traceback doesn't work (`_G` is returned from debug.traceback)
SourceMap traceback currently causes `_G` to be returned from debug.traceback.

It seems that this is because `globalThis.__TS__originalTraceback` still thinks it has a `self` parameter, and so is translated to `globalThis:__TS__originalTraceback()`.

GlassBricks:
Fixes #1086

### DIFF
--- a/src/lualib/SourceMapTraceBack.ts
+++ b/src/lualib/SourceMapTraceBack.ts
@@ -1,5 +1,6 @@
 // TODO: In the future, change this to __TS__RegisterFileInfo and provide tstl interface to
 // get some metadata about transpilation.
+declare function __TS__originalTraceback(this: void, thread?: LuaThread, message?: string, level?: number);
 function __TS__SourceMapTraceBack(this: void, fileName: string, sourceMap: { [line: number]: number }): void {
     globalThis.__TS__sourcemap = globalThis.__TS__sourcemap || {};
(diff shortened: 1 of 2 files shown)


src/lualib/SourceMapTraceBack.ts
```diff
@@ -9,9 +10,13 @@ function __TS__SourceMapTraceBack(this: void, fileName: string, sourceMap: { [li
         debug.traceback = ((thread, message, level) => {
             let trace: string;
             if (thread === undefined && message === undefined && level === undefined) {
-                trace = globalThis.__TS__originalTraceback();
+                trace = (globalThis.__TS__originalTraceback as (this: void, ...args: any) => string)();
```
Perryvw, line 13: These duplicate casts seem pretty ugly (actually it's ugly these function are any in the first place), would it work to just add
```ts
declare function __TS__originalTraceback(this: void, thread?: LuaThread, message?: string, level?: number);
```
At the top of the file?

GlassBricks: Yes, that does work...

For some reason, before, I was trying to do `declare let __Ts__originalTraceback : (//)=>any` and that didn't work, so I did the cast instead.

PR #1090, opened 2021-08-15
Remove tupleReturn annotation

### LINKED ISSUE #1058: @link JSDoc annotation cause to fatal error
**Description**
I have updated TypeScriptToLua from `0.39.3` to the latest, `0.41.0`, and received an error:
```console
error TS2304: Cannot find name 'https'
```

However, I use https only in my JSDoc comment. 
I have prepared the simple example to reproduce the error.

**Code**
```typescript
/**
 * @see {@link https://toliak.ru/ SomeWebsite }
 */
function helloWorld() {
    return 'Hello World';
}
```

Also those lines cause to error:
- ` * @see a a a a {@link https://toliak.ru/ SomeWebsite }`
- ` * @asdadasd {@link https://toliak.ru/ SomeWebsite }`
- ` * @asdadasd a a a {@link https://toliak.ru/ SomeWebsite }`

And those lines works fine:
- ` * aboba {@link https://toliak.ru/ SomeWebsite }`
- `* {@link https://toliak.ru/ SomeWebsite }`

I think that is a bug, because I use https in the comment and there is no rules, that prohibit me from using link here.

My environment version info:

**NPM version**
```console
{
  p_20210801: '1.0.0',
  npm: '6.14.10',
  ares: '1.16.1',
  brotli: '1.0.9',
  cldr: '37.0',
  icu: '67.1',
  llhttp: '2.1.3',
  modules: '83',
  napi: '7',
  nghttp2: '1.41.0',
  node: '14.15.4',
  openssl: '1.1.1i',
  tz: '2020a',
  unicode: '13.0',
  uv: '1.40.0',
  v8: '8.4.371.19-node.17',
  zlib: '1.2.11'
}
```

**NPM Packages**
```console
+-- typescript@4.3.5
`-- typescript-to-lua@0.41.0
  +-- enhanced-resolve@5.8.2
  | +-- graceful-fs@4.2.6
  | `-- tapable@2.2.0
  +-- resolve@1.20.0
  | +-- is-core-module@2.5.0
  | | `-- has@1.0.3
  | |   `-- function-bind@1.1.1
  | `-- path-parse@1.0.7
  +-- source-map@0.7.3
  `-- typescript@4.3.5 deduped
```

**What about `tsc`**
It works correctly

```console
> tsc -b --verbose

[6:34:37 PM] Projects in this build:
    * tsconfig.json

[6:34:37 PM] Project 'tsconfig.json' is out of date because output file 'index.js' does not exist

[6:34:37 PM] Building project 'C:/Users/Toliak/WebstormProjects/p_20210801/tsconfig.json'...

Process finished with exit code 0
```

Perryvw:
As a side effect of removing `isInTupleReturn`, this also fixes #1058


test/unit/annotations/deprecated.spec.ts
```diff
@@ -60,6 +60,14 @@ test("luaiterator removed", () => {
     `.expectDiagnosticsToMatchSnapshot([annotationRemoved.code]);
 });
 
+test("tuplereturn removed", () => {
+    util.testFunction`
+        /** @tupleReturn */
+        function tuple(): [number, number, number] { return [3, 5, 1]; }
+        return tuple()[2];
+    `.expectDiagnosticsToMatchSnapshot([annotationRemoved.code]);
```
lolleko, line 68: Add a test for just:

```
        /** @tupleReturn */
         function tuple(): [number, number, number] { return [3, 5, 1]; };
```
(without the call). This should also result in a diagnostic

src/transformation/utils/annotations.ts
```diff
@@ -118,7 +117,7 @@ export function isTupleReturnCall(context: TransformationContext, node: ts.Node)
     const signature = context.checker.getResolvedSignature(node);
     if (signature) {
         if (getSignatureAnnotations(context, signature).has(AnnotationKind.TupleReturn)) {
-            context.diagnostics.push(annotationDeprecated(node, AnnotationKind.TupleReturn));
+            context.diagnostics.push(annotationRemoved(node, AnnotationKind.TupleReturn));
```
lolleko, line 120: I think diagnostic.push should be removed from isTupleReturnCall(). isTupleReturnCall should just return true or false without side effects.
The diagnostic should only be pushed at the call location of isTupleReturnCall().
e.g.
```
if (isTupleReturnCall()) {
    context.diagnostic.push...
}
```
Which seems to already be the case???? https://github.com/TypeScriptToLua/TypeScriptToLua/pull/1090/files#diff-877ef39cd04cebdd5a0baf7a2090e119498765d5afc71abe08f32422d1dd1984R252
I wonder why the test does not fail because with this implementation, the diagnostic should get pushed multiple times?

lolleko (changes requested):

lolleko (approved):

PR #1093, opened 2021-08-18
Async await

Perryvw:
Implemented support for async/await using generators/yields as explained [in this article](https://hackernoon.com/async-await-generators-promises-51f1a6ceede2). The actual implementation is based on the TypeScript polyfill as included in the comments in the lualib source.

Generators and yield are implemented in tstl using coroutines, for now it is not possible to use await at the very top level of your code because that would be outside of coroutines, for now `await` must appear in an `async` function.


test/unit/builtins/async-await.spec.ts
```diff
@@ -0,0 +1,242 @@
...
+
+        a(); // Call async function (but cannot await)
+    `
+        .setOptions({ module: ModuleKind.ESNext, target: ScriptTarget.ES2017 })
+        .expectToEqual({
+            aStarted: true,
+        });
+});
+
+test.each(["await a();", "const b = await a();", "export const b = await a();"])(
```
tomblind, line 228: Some suggestions:
```ts
foo(await a())
const b = [await a()];
const b = (await a(), await a());
const b = condition ? await a() : someOtherValue;
```

tomblind (changes requested):
It would be good to have tests of throwing errors inside async functions as well as catching them with with try...catch blocks.

Also, while it should work fine, a good test would be awaiting inside of a statement that gets wrapped in an IIFE, like a complex conditional `:?`.

src/transformation/visitors/function.ts
```diff
@@ -196,7 +197,7 @@ export function transformFunctionToExpression(
         node
     );
     const functionExpression = lua.createFunctionExpression(
-        lua.createBlock(transformedBody),
+        lua.createBlock(isAsyncFunction(node) ? wrapInAsyncAwaiter(context, transformedBody) : transformedBody),
```
lolleko, line 200: Can you move this to an extra variable or fucntion?
It should be more clear that async is handled here, otherwise it's easy to over read.

test/unit/builtins/async-await.spec.ts
```diff
@@ -0,0 +1,242 @@
+import { ModuleKind, ScriptTarget } from "typescript";
+import { notAllowedTopLevelAwait } from "../../../src/transformation/utils/diagnostics";
+import * as util from "../../util";
+
+const promiseTestLib = `
```
lolleko, line 5: Can this be shared with the promises tests?

test/unit/builtins/async-await.spec.ts
```diff
@@ -0,0 +1,242 @@
...
+            return 42;
+        }
+
+        ${awaitUsage}
+        export {} // Required to make TS happy, cannot await without import/exports
+    `
+            .setOptions({ module: ModuleKind.ESNext, target: ScriptTarget.ES2017 })
+            .expectToHaveDiagnostics([notAllowedTopLevelAwait.code]);
+    }
+);
```
lolleko, line 242: Maybe a test (or a few) testing that await does not have any effect when used on non promises.
e.g.

```
const x = 1;
const y = 2;

test((await x) + y == 3)
```

I am not sure if that is possible, but maybe you can detect awaits like this at compile time and ommit them completely (don't wrap them).

TS emits a warning/info diagnostic for such awaits: https://www.typescriptlang.org/play?#code/MYewdgzgLgBAHjAvDAjAbgFCkrAnkmAJky3GhgC8CBDAd2oEtYEBqGXEgUzgAcQAnWAG8YAX0xA

test/unit/builtins/async-await.spec.ts
```diff
@@ -0,0 +1,242 @@
...
+        .expectToEqual({
+            aStarted: true,
+        });
+});
+
+test.each(["await a();", "const b = await a();", "export const b = await a();"])(
+    "cannot await at top-level (%p)",
+    awaitUsage => {
+        util.testModule`
+        async function a() {
```
lolleko, line 232: Maybe a test for async lambdas? Seems like only async functions are tested.

lolleko (changes requested):

test/unit/builtins/async-await.spec.ts
Perryvw: We cannot export code from .spec.ts files, because next to the exported thing, when importing it all tests from the other file are also imported. Jest has a eslint rule that prevents this. I think just having the code on top of both files is clearer than putting it in a shared .ts file, even though it is a bit of duplication.

lolleko (approved):

src/transformation/visitors/sourceFile.ts
```diff
@@ -43,3 +50,14 @@ export const transformSourceFileNode: FunctionVisitor<ts.SourceFile> = (node, co
     const trivia = node.getFullText().match(/^#!.*\r?\n/)?.[0] ?? "";
     return lua.createFile(statements, getUsedLuaLibFeatures(context), trivia, node);
 };
+
+function isTopLevelAwait(statement: ts.Statement) {
+    // Check if expression contains an await child, without going into function declarations
+    let containsAwait = false;
+    traverseWithoutCrossingFunction(statement, node => {
```
tomblind, line 57: It seems pretty intense to recurse the ast downwards from every statement to find top-level awaits. Why not something along the lines of `findFirstNodeAbove(node, n => isAsyncFunction(n))` in `transformAwaitExpression`?

src/lualib/Await.ts
```diff
@@ -0,0 +1,56 @@
...
+function __TS__AsyncAwaiter(this: void, generator: (this: void) => void) {
+    return new Promise((resolve, reject) => {
+        const asyncCoroutine = coroutine.create(generator);
+
+        // eslint-disable-next-line @typescript-eslint/promise-function-async
+        function adopt(value: unknown) {
+            return value instanceof __TS__Promise ? value : Promise.resolve(value);
+        }
+        function fulfilled(value) {
+            try {
```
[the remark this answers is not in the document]
Perryvw: Meant to remove this but seems I forgot, it's gone now :)

src/transformation/visitors/sourceFile.ts
Perryvw: You're right, I updated how this check is done

tomblind (changes requested):
Was about to approve this when I realized something. This isn't going to play with varargs at all. Currently it will optimize and try to directly use the ellipses. But even if you disable vararg optimzation for async functions, the wrapping of the ellipsis will still occur inside the function expression, which will break. The wrapping of the ellipsis needs to get lifted out somehow.
```ts
async function foo(...args: unknown[]) {
    console.log(...args);
}
```
```lua
function foo(self, ...)
    return __TS__AsyncAwaiter(
        function()
            local args = {...}
            print(
                table.unpack(args)
            )
        end
    )
end
```

tomblind (approved):

hubol:
I'm so excited to use this!!! Thank you

PR #1098, opened 2021-08-21
feat: switch statements for lua 5.1+ (universal)

### LINKED ISSUE #194: Switch statements not allowed in Lua 5.1

thejustinwalsh:
Resolves #194 (or rather brings support for Lua 5.1 and universal).

Implemented using `repeat...until true`, `if`, and some conditional `or` chaining. Relies on switch fallthrough amalgamation / unrolling for default case fallthrough.

```ts
switch (0 as number) {
    case 0:
    case 1:
    case 2:
        out.push("0,1,2");
    default:
        out.push("default");
    case 3:
        out.push("3");
}
```

```lua
repeat
    local ____switch3 = 0
    local ____cond3 = ((____switch3 == 0) or (____switch3 == 1)) or (____switch3 == 2)
    if ____cond3 then
        __TS__ArrayPush(out, \\"0,1,2\\")
    end
    if ____cond3 then
        __TS__ArrayPush(out, \\"default\\")
    end
    ____cond3 = ____cond3 or (____switch3 == 3)
    if ____cond3 then
        __TS__ArrayPush(out, \\"3\\")
    end
    if not ____cond3 then
        __TS__ArrayPush(out, \\"default\\")
        __TS__ArrayPush(out, \\"3\\")
    end
until true
```

## Update:

The final approach was to `or` together the conditions into a variable and check that condition in the if statement to ensure that we do not add any additional comparisons that would trigger undesired side effects.

In order to handle the default case, we must still amalgamate the fallthrough cases that a default would fall into, and provide that as the final `or` clause, by checking that no previous condition had passed.

I am making my best attempt to ensure that any redundant or useless cases are removed, and empty cases coalesced.

Unforeseen bonus, the generated code is a bit easier to reason about!

### Orignal:

The approach I took is to gather the fall-through bodies based on the defined order in the switch statement and amalgamate them until I hit a break statement that would break the control flow of the switch case. I then move the resulting default block to a new else statement. It is crucial to handle the code generation for the default case before moving it into the else block, as the rules of fall through would dictate that the case directly after the default would execute.

Took lots of testing until it clicked in my head that this method is indeed conformant with how a switch statement executes. `switchWithBracketsBreakInInternalLoop` was the real kicker, and made me realize handling break statements in their proper scope was missing. By skipping the child tree of any loop or switch the code will not pick up unnecessary breaks of switch control flow.

Performance would be the only open question, although I would not expect this method to be any slower than jumping with goto, as it is really just an unrolled switch statement.

Also send me your code clarity, clean-up, cognitive load lightening, and readability suggestions, please!

### DIFF
--- a/src/transformation/visitors/break-continue.ts
+++ b/src/transformation/visitors/break-continue.ts
@@ -8,6 +8,7 @@ import { findScope, ScopeType } from "../utils/scope";
 export const transformBreakStatement: FunctionVisitor<ts.BreakStatement> = (breakStatement, context) => {
     const breakableScope = findScope(context, ScopeType.Loop | ScopeType.Switch);
     if (breakableScope?.type === ScopeType.Switch) {
+        // Break is handled by the switch statement (see transformSwitchStatement)
         return lua.createBreakStatement(breakStatement);
     } else {
         return lua.createBreakStatement(breakStatement);
--- a/src/transformation/visitors/break-continue.ts
+++ b/src/transformation/visitors/break-continue.ts
@@ -8,7 +8,7 @@ import { findScope, ScopeType } from "../utils/scope";
 export const transformBreakStatement: FunctionVisitor<ts.BreakStatement> = (breakStatement, context) => {
     const breakableScope = findScope(context, ScopeType.Loop | ScopeType.Switch);
     if (breakableScope?.type === ScopeType.Switch) {
-        return lua.createGotoStatement(`____switch${breakableScope.id}_end`);
+        return undefined;
     } else {
         return lua.createBreakStatement(breakStatement);
     }
--- a/src/transformation/visitors/break-continue.ts
+++ b/src/transformation/visitors/break-continue.ts
@@ -6,13 +6,8 @@ import { unsupportedForTarget } from "../utils/diagnostics";
 import { findScope, ScopeType } from "../utils/scope";
 
 export const transformBreakStatement: FunctionVisitor<ts.BreakStatement> = (breakStatement, context) => {
-    const breakableScope = findScope(context, ScopeType.Loop | ScopeType.Switch);
-    if (breakableScope?.type === ScopeType.Switch) {
-        // Break is handled by the switch statement (see transformSwitchStatement)
-        return lua.createBreakStatement(breakStatement);
-    } else {
-        return lua.createBreakStatement(breakStatement);
-    }
+    void context;
+    return lua.createBreakStatement(breakStatement);
 };
 
 export const transformContinueStatement: FunctionVisitor<ts.ContinueStatement> = (statement, context) => {
--- a/src/transformation/visitors/break-continue.ts
+++ b/src/transformation/visitors/break-continue.ts
@@ -8,7 +8,7 @@ import { findScope, ScopeType } from "../utils/scope";
 export const transformBreakStatement: FunctionVisitor<ts.BreakStatement> = (breakStatement, context) => {
     const breakableScope = findScope(context, ScopeType.Loop | ScopeType.Switch);
     if (breakableScope?.type === ScopeType.Switch) {
-        return undefined;
+        return lua.createBreakStatement(breakStatement);
     } else {
         return lua.createBreakStatement(breakStatement);
     }
--- a/src/transformation/visitors/switch.ts
+++ b/src/transformation/visitors/switch.ts
@@ -1,6 +1,5 @@
 import * as ts from "typescript";
 import * as lua from "../../LuaAST";
-import { assert } from "../../utils";
 import { FunctionVisitor } from "../context";
 import { performHoisting, popScope, pushScope, ScopeType } from "../utils/scope";
 
@@ -86,8 +85,15 @@ export const transformSwitchStatement: FunctionVisitor<ts.SwitchStatement> = (st
                           )
                 );
                 isInitialCondition = false;
-            } else {
-                assert(!isInitialCondition, "Default clause should never be the initial condition");
+            } else if (isInitialCondition) {
+                // If we have fallen through to the default here, we need to initialize the default condition
+                statements.push(
+                    lua.createVariableDeclarationStatement(
+                        conditionVariable,
+                        condition ?? lua.createBooleanLiteral(false)
+                    )
+                );
+                isInitialCondition = false;
             }
 
             // Push if statement for case
@@ -102,19 +108,25 @@ export const transformSwitchStatement: FunctionVisitor<ts.SwitchStatement> = (st
             condition = undefined;
         }
 
-        // Amalgamate the default w/ fallthrough clauses and execute if nothing else executed above
+        // If no conditions above match, we need to create the final default case code-path
+        // As we only handle fallthrough into defaults in the previous if statement chain
         const start = clauses.findIndex(c => ts.isDefaultClause(c));
         if (start >= 0) {
-            const end = statement.caseBlock.clauses
-                .slice(start)
-                .findIndex(clause => containsBreakOrReturn([...clause.statements]));
-            const defaultStatements = statement.caseBlock.clauses
-                .slice(start, end >= 0 ? end + start + 1 : undefined)
-                .reduce<lua.Statement[]>(
-                    (statements, clause) => [...statements, ...context.transformStatements(clause.statements)],
-                    []
-                );
+            // Find the last clause that we can fallthrough to
+            const end = statement.caseBlock.clauses.findIndex(
+                (clause, index) => index >= start && containsBreakOrReturn([...clause.statements])
+            );
 
+            // Combine the default and all fallthrough statements
+            const defaultStatements: lua.Statement[] = [];
+            for (const { statements } of statement.caseBlock.clauses.slice(
+                start,
+                end >= 0 ? end + start + 1 : undefined
+            )) {
+                defaultStatements.push(...context.transformStatements(statements));
+            }
+
+            // Add the default clause if it has any statements
             if (defaultStatements.length) {
                 statements.push(
                     lua.createIfStatement(
--- a/src/transformation/visitors/switch.ts
+++ b/src/transformation/visitors/switch.ts
@@ -93,24 +93,32 @@ export const transformSwitchStatement: FunctionVisitor<ts.SwitchStatement> = (st
                           )
                 );
                 isInitialCondition = false;
-            } else if (isInitialCondition) {
-                // If we have fallen through to the default here, we need to initialize the default condition
-                statements.push(
-                    lua.createVariableDeclarationStatement(
-                        conditionVariable,
-                        condition ?? lua.createBooleanLiteral(false)
-                    )
-                );
-                isInitialCondition = false;
+            } else {
+                // If the default is not fallen into skip it and handle in the final default output instead
+                if (previousClause && containsBreakOrReturn(previousClause.statements)) {
+                    continue;
+                }
+
+                // If the default is proceeded by empty clauses and will be emitted we may need to initialize the condition
+                if (isInitialCondition) {
+                    statements.push(
+                        lua.createVariableDeclarationStatement(
+                            conditionVariable,
+                            condition ?? lua.createBooleanLiteral(false)
+                        )
+                    );
+                    isInitialCondition = false;
+                }
+            }
+
+            // Transform the clause and append the final break statement if necessary
+            const clauseStatements = context.transformStatements(clause.statements);
+            if (i === clauses.length - 1 && !containsBreakOrReturn(clause.statements)) {
+                clauseStatements.push(lua.createBreakStatement());
             }
 
             // Push if statement for case
-            statements.push(
-                lua.createIfStatement(
-                    conditionVariable,
-                    lua.createBlock(context.transformStatements(clause.statements))
-                )
-            );
+            statements.push(lua.createIfStatement(conditionVariable, lua.createBlock(clauseStatements)));
 
             // Clear condition for next clause
             condition = undefined;
@@ -132,13 +140,9 @@ export const transformSwitchStatement: FunctionVisitor<ts.SwitchStatement> = (st
                 .forEach(c => defaultStatements.push(...context.transformStatements(c.statements)));
 
             // Add the default clause if it has any statements
+            // The switch will always break on the final clause and skip execution if valid to do so
             if (defaultStatements.length) {
-                statements.push(
-                    lua.createIfStatement(
-                        lua.createUnaryExpression(conditionVariable, lua.SyntaxKind.NotOperator),
-                        lua.createBlock(defaultStatements)
-                    )
-                );
+                statements.push(lua.createDoStatement(defaultStatements));
             }
         }
     }
--- a/src/transformation/visitors/switch.ts
+++ b/src/transformation/visitors/switch.ts
@@ -65,10 +65,11 @@ export const transformSwitchStatement: FunctionVisitor<ts.SwitchStatement> = (st
         // Default is the not of all known case expressions
         let previousClause: ts.CaseOrDefaultClause;
         let condition: lua.Expression | undefined;
+        let isInitialCondition = true;
         statement.caseBlock.clauses.forEach(clause => {
             if (!condition || (previousClause && containsBreakOrReturn([...previousClause.statements]))) {
                 if (ts.isDefaultClause(clause)) {
-                    // If the default is first, or followed by a break, we can't fall into it, skip.
+                    condition = undefined;
                 } else {
                     condition = lua.createBinaryExpression(
                         switchVariable,
@@ -76,39 +77,46 @@ export const transformSwitchStatement: FunctionVisitor<ts.SwitchStatement> = (st
                         lua.SyntaxKind.EqualityOperator
                     );
                 }
-            } else {
-                if (ts.isDefaultClause(clause)) {
-                    // use the previous condition for the default clause
-                } else {
-                    condition = lua.createBinaryExpression(
-                        condition,
-                        lua.createBinaryExpression(
-                            switchVariable,
-                            context.transformExpression(clause.expression),
-                            lua.SyntaxKind.EqualityOperator
-                        ),
-                        lua.SyntaxKind.OrOperator
-                    );
-                }
+            } else if (!ts.isDefaultClause(clause)) {
+                condition = lua.createBinaryExpression(
+                    condition,
+                    lua.createBinaryExpression(
+                        switchVariable,
+                        context.transformExpression(clause.expression),
+                        lua.SyntaxKind.EqualityOperator
+                    ),
+                    lua.SyntaxKind.OrOperator
+                );
             }
 
             if (clause.statements.length) {
                 if (!ts.isDefaultClause(clause) && condition) {
                     statements.push(
-                        lua.createAssignmentStatement(
-                            conditionVariable,
-                            lua.createBinaryExpression(conditionVariable, condition, lua.SyntaxKind.OrOperator)
-                        )
+                        isInitialCondition
+                            ? lua.createVariableDeclarationStatement(conditionVariable, condition)
+                            : lua.createAssignmentStatement(
+                                  conditionVariable,
+                                  lua.createBinaryExpression(conditionVariable, condition, lua.SyntaxKind.OrOperator)
+                              )
                     );
+                    isInitialCondition = false;
                     condition = undefined;
                 }
 
-                statements.push(
-                    lua.createIfStatement(
-                        conditionVariable,
-                        lua.createBlock(context.transformStatements(clause.statements))
+                if (
+                    !(
+                        ts.isDefaultClause(clause) &&
+                        previousClause &&
+                        containsBreakOrReturn([...previousClause.statements])
                     )
-                );
+                ) {
+                    statements.push(
+                        lua.createIfStatement(
+                            conditionVariable,
+                            lua.createBlock(context.transformStatements(clause.statements))
+                        )
+                    );
+                }
             }
 
             previousClause = clause;
@@ -144,7 +152,6 @@ export const transformSwitchStatement: FunctionVisitor<ts.SwitchStatement> = (st
 
     // Add the switch expression after hoisting
     const expression = context.transformExpression(statement.expression);
-    statements.unshift(lua.createVariableDeclarationStatement(conditionVariable, lua.createBooleanLiteral(false)));
     statements.unshift(lua.createVariableDeclarationStatement(switchVariable, expression));
 
     // Wrap the statements in a repeat until true statement to facilitate dynamic break/returns
--- a/src/transformation/visitors/switch.ts
+++ b/src/transformation/visitors/switch.ts
@@ -21,9 +21,11 @@ const containsBreakOrReturn = (statements: ts.Node[]): boolean => {
 export const transformSwitchStatement: FunctionVisitor<ts.SwitchStatement> = (statement, context) => {
     const scope = pushScope(context, ScopeType.Switch);
 
-    // Give the switch a unique name to prevent nested switches from acting up.
+    // Give the switch and condition accumulator a unique name to prevent nested switches from acting up.
     const switchName = `____switch${scope.id}`;
+    const conditionName = `____cond${scope.id}`;
     const switchVariable = lua.createIdentifier(switchName);
+    const conditionVariable = lua.createIdentifier(conditionName);
 
     // Collect all the expressions into a single expression for use in the default clause
     let allExpressions: lua.BinaryExpression;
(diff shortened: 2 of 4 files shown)


src/transformation/visitors/break-continue.ts
```diff
@@ -8,7 +8,7 @@ import { findScope, ScopeType } from "../utils/scope";
 export const transformBreakStatement: FunctionVisitor<ts.BreakStatement> = (breakStatement, context) => {
     const breakableScope = findScope(context, ScopeType.Loop | ScopeType.Switch);
     if (breakableScope?.type === ScopeType.Switch) {
-        return lua.createGotoStatement(`____switch${breakableScope.id}_end`);
+        return undefined;
     } else {
```
Perryvw, line 12: Maybe add a comment stating switch breaks are handled by the switch tranfromation.

src/transformation/visitors/switch.ts
```diff
@@ -1,61 +1,89 @@
...
+export const transformSwitchStatement: FunctionVisitor<ts.SwitchStatement> = (statement, context) => {
     const scope = pushScope(context, ScopeType.Switch);
 
     // Give the switch a unique name to prevent nested switches from acting up.
     const switchName = `____switch${scope.id}`;
     const switchVariable = lua.createIdentifier(switchName);
 
-    let statements: lua.Statement[] = [];
+    // Collect the fallthrough bodies for each case as defined by the switch.
+    const caseBody: lua.Statement[][] = [];
```
Perryvw, line 36: So if I understand correctly: you copy/duplicate the fallthrough statements into individual clauses without any fallthrough? Might be nice to add that as a comment, maybe even with a small 2-clause example like:
```ts
// case A: 
//    foo(); 
// case B: 
//    bar(); 
//    break;
// ->
// if (A) { 
//     foo();
//     bar();
// } else if (B) { 
//     bar(); 
// }
```

test/unit/__snapshots__/switch.spec.ts.snap
```diff
@@ -1,52 +1,25 @@
...
-        end
-    end
-    ::____switch3_case_2::
     do
-        do
-            result = 1
-            goto ____switch3_end
+        local ____switch3 = 2
+        if ____switch3 == 0 then
+            do
```
Perryvw, line 10: Do we still need these `do` blocks now that we encapsulate everything in if scopes already?

Perryvw (commented):
Seems like a smart solution! Could we maybe optimize this a little bit by aggregating empty fallthrough cases into a single if clause? Would be a nice snapshot test too. I'm thinking of something like:
```ts
switch (a) {
    case b:
    case c:
    case d:
        foo();
    default:
        bar();
}
```
translating into something like:
```lua
if (a == b or a == c or a == d) then
    foo()
else
    bar();
end
```

src/transformation/visitors/switch.ts
thejustinwalsh: Correct. Although, after looking at some more uses of a switch in the wild, this method can create more code duplication than anticipated when a return statement is used over a break.

test/unit/__snapshots__/switch.spec.ts.snap
thejustinwalsh: These do statements are here because the test defines the cases like so...
```ts
case 1: {
  result = 100
  break;
}
```

The parens create the do in the above case, the below would be sans do block.
```ts
case 1:
  result = 100
  break;
```

thejustinwalsh:
> Seems like a smart solution! Could we maybe optimize this a little bit by aggregating empty fallthrough cases into a single if clause? Would be a nice snapshot test too. I'm thinking of something like:
> 
> ```ts
> switch (a) {
>     case b:
>     case c:
>     case d:
>         foo();
>         break;
>     default:
>         bar();
>         break;
> }
> ```
> 
> translating into something like:
> 
> ```lua
> if (a == b or a == c or a == d) then
>     foo()
> else
>     bar();
> end
> ```

Great minds think alike! ;)

I didn't think of the empty case grouping, but I was onto this idea that if I actually didn't use an else statement at all, I could handle all cases with `if` and `or`. for example...

```ts
switch (a) {
    case b:
    case c:
        foo();
    case d:
        bar();
        break;
    default:
        baz();
        break;
}
```
```lua
if (a == b || a == c || a == d) then
    foo()
end
if ( a == d) then
    bar()
end
if (a == b or a == c or a == d) then
    foo()
end
if (not (a == b or a == c or a == d))
    bar();
end
```

With this approach, you still collect the fallthroughs, but then instead of duplicating the code blocks, you let the if's cascade, similar to how it was done with goto statements. I am fairly certain that the hoisting would allow for this to result in the same execution as the code duplication technique.

The default case would then always have to be the not of all the cases present in the switch, which could be further optimized by storing the right-hand side of the compare in a table as the key where the absence of the key means it is the default case. Special care would need to be taken around nil, as that can't be a key in the table, but could be a case clause.

thejustinwalsh:
Test added in 9172e88 with the contents of this comment https://github.com/TypeScriptToLua/TypeScriptToLua/issues/1099#issuecomment-903689466 to ensure proper scoping. implementation on master fails this test.

thejustinwalsh:
@tomblind `repeat...until` in 504fd1a

test/unit/__snapshots__/switch.spec.ts.snap
```diff
@@ -1,52 +1,56 @@
...
+            break
+        end
+        if not (((((____switch3 == 0) or (____switch3 == 1)) or (____switch3 == 2)) or (____switch3 == 3)) or (____switch3 == 4)) then
+            __TS__ArrayPush(out, \\"default\\")
+        end
+        if (not (((((____switch3 == 0) or (____switch3 == 1)) or (____switch3 == 2)) or (____switch3 == 3)) or (____switch3 == 4))) or (____switch3 == 4) then
+            __TS__ArrayPush(out, \\"4\\")
+        end
+    until true
+    return out
```
thejustinwalsh, line 24: @Perryvw any idea on how to get rid of all those extra parentheses?
It appears that using `lua.createBinaryExpression` is causing a benign nesting here.

thejustinwalsh:
Ready for review.
Any other tests we should run through for good measure?

src/transformation/visitors/break-continue.ts
```diff
@@ -8,7 +8,8 @@ import { findScope, ScopeType } from "../utils/scope";
 export const transformBreakStatement: FunctionVisitor<ts.BreakStatement> = (breakStatement, context) => {
     const breakableScope = findScope(context, ScopeType.Loop | ScopeType.Switch);
     if (breakableScope?.type === ScopeType.Switch) {
-        return lua.createGotoStatement(`____switch${breakableScope.id}_end`);
+        // Break is handled by the switch statement (see transformSwitchStatement)
+        return lua.createBreakStatement(breakStatement);
```
tomblind, line 12: Isn't this redundant?

src/transformation/visitors/switch.ts
```diff
@@ -1,61 +1,152 @@
...
 
-    // Give the switch a unique name to prevent nested switches from acting up.
+    // Give the switch and condition accumulator a unique name to prevent nested switches from acting up.
     const switchName = `____switch${scope.id}`;
+    const conditionName = `____cond${scope.id}`;
     const switchVariable = lua.createIdentifier(switchName);
+    const conditionVariable = lua.createIdentifier(conditionName);
+
+    // Collect all the expressions into a single expression for use in the default clause
+    let allExpressions: lua.BinaryExpression;
```
tomblind, line 31: This isn't needed anymore, right?

src/transformation/visitors/switch.ts
```diff
@@ -1,61 +1,152 @@
...
+                            lua.createBinaryExpression(conditionVariable, condition, lua.SyntaxKind.OrOperator)
+                        )
+                    );
+                    condition = undefined;
+                }
 
-        const goto = lua.createGotoStatement(`${switchName}_case_${index}`);
-        return lua.createIfStatement(condition, lua.createBlock([goto]), previousCondition);
-    }, undefined as lua.IfStatement | undefined);
+                statements.push(
```
tomblind, line 106: We should skip this on default clauses that aren't fallen into so we don't output an erroneous if block that won't be hit.

src/transformation/visitors/switch.ts
```diff
@@ -1,61 +1,152 @@
...
 
-    statements.push(lua.createLabelStatement(`${switchName}_end`));
-
+    // Hoist the variable, function, and import statements to the top of the switch
     statements = performHoisting(context, statements);
     popScope(context);
 
+    // Add the switch expression after hoisting
     const expression = context.transformExpression(statement.expression);
+    statements.unshift(lua.createVariableDeclarationStatement(conditionVariable, lua.createBooleanLiteral(false)));
```
tomblind, line 147: Would it be possible to initialize the condition variable to the first case's expression?

test/unit/switch.spec.ts
```diff
@@ -359,3 +365,61 @@ test("switch fallthrough stops after default", () => {
...
+    `
+        .expectLuaToMatchSnapshot()
+        .expectToMatchJsResult();
+});
+
+test("switch handles side-effects", () => {
+    util.testFunction`
+        const out = [];
+        
+        let y = 0;
```
tomblind, line 407: It would be good to have variants of this test with different starting y values.

tomblind (changes requested):

src/transformation/visitors/switch.ts
thejustinwalsh: Correct will clean up.

src/transformation/visitors/break-continue.ts
thejustinwalsh: True. was leaving it there for a smaller delta. If I remove everything else out of that function, the context is no longer needed, guessing I should just do a `void context;` to silence the unused error?

src/transformation/visitors/switch.ts
thejustinwalsh: Agreed, I spotted this one after I got tests passing.

src/transformation/visitors/switch.ts
thejustinwalsh: Sure. I was initially concerned that hoisting was going to mess with me. so opted for the most straightforward.
Also, it saves an if in the code above and allows for the same logic for each assignment. Going to clean up everything else first, then take a look at how much more complicated the code might look.

src/transformation/visitors/switch.ts
```diff
@@ -1,61 +1,136 @@
...
         }
+    } else {
+        // Build up the condition for each if statement
+        // Fallthrough is handled by accepting the last condition as an additional or clause
+        // Default is the not of all known case expressions
+        let previousClause: ts.CaseOrDefaultClause;
+        let condition: lua.Expression | undefined;
+        let isInitialCondition = true;
+        statement.caseBlock.clauses.forEach(clause => {
+            if (!condition || (previousClause && containsBreakOrReturn([...previousClause.statements]))) {
```
tomblind, line 47: I'm confused about the logic here. If the previous clause contained a break/return, the condition is simply assigned, but it will still be wrapped in an `or` down below. Why not do that here like clauses that don't follow a break?

tomblind (changes requested):

src/transformation/visitors/switch.ts
thejustinwalsh: Could use a refactor now that all tests are passing. This is just churn on solving the problem.
You sure you don't want to get in on this... ;)

src/transformation/visitors/switch.ts
```diff
@@ -1,61 +1,136 @@
...
 
-        const goto = lua.createGotoStatement(`${switchName}_case_${index}`);
-        return lua.createIfStatement(condition, lua.createBlock([goto]), previousCondition);
-    }, undefined as lua.IfStatement | undefined);
+                if (
+                    !(
+                        ts.isDefaultClause(clause) &&
+                        previousClause &&
+                        containsBreakOrReturn([...previousClause.statements])
+                    )
```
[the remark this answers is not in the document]
thejustinwalsh: Fixed.

test/unit/__snapshots__/switch.spec.ts.snap
thejustinwalsh: I went ahead and removed this snapshot btw, created a better one to illustrate the many cases we had talked through already.

src/transformation/visitors/switch.ts
thejustinwalsh: Fixed in 8d1f74a105e1994e5ee6a7df0a2851d217fae06b

src/transformation/visitors/switch.ts
```diff
@@ -1,61 +1,139 @@
...
+            // Skip redundant default clauses, will be handled in final default case
+            if (i === 0 && ts.isDefaultClause(clause)) continue;
+            if (ts.isDefaultClause(clause) && previousClause && containsBreakOrReturn([...previousClause.statements])) {
+                continue;
+            }
 
-        const goto = lua.createGotoStatement(`${switchName}_case_${index}`);
-        return lua.createIfStatement(condition, lua.createBlock([goto]), previousCondition);
-    }, undefined as lua.IfStatement | undefined);
+            // Compute the condition for the if statement
```
Perryvw, line 54: Candidate for refactoring to its own function?

src/transformation/visitors/switch.ts
```diff
@@ -1,61 +1,139 @@
...
+                    lua.createBlock(context.transformStatements(clause.statements))
+                )
+            );
+
+            // Clear condition for next clause
+            condition = undefined;
+        }
 
-    statements.push(lua.createLabelStatement(`${switchName}_end`));
+        // Amalgamate the default w/ fallthrough clauses and execute if nothing else executed above
```
Perryvw, line 105: The code below seems like some black magic, and this comment is not really helping understand it, it might be making it worse actually.

tomblind (changes requested):
Thanks for the refactor - this is much easier to follow now. There's still a few tricky situations to deal with, though.

src/transformation/visitors/switch.ts
```diff
@@ -1,61 +1,139 @@
...
+                .slice(start, end >= 0 ? end + start + 1 : undefined)
+                .reduce<lua.Statement[]>(
+                    (statements, clause) => [...statements, ...context.transformStatements(clause.statements)],
+                    []
+                );
+
+            if (defaultStatements.length) {
+                statements.push(
+                    lua.createIfStatement(
+                        lua.createUnaryExpression(conditionVariable, lua.SyntaxKind.NotOperator),
```
tomblind, line 121: Something I was thinking about with this is that in almost all cases, the `if not ____cond` check is unecessary. If code reached this point we definitely want the default clause to run. The one exception is when the last case in the switch is not default and doesn't explicitly break:

```ts
switch (x) {
    default:
        console.log("default");
    case 1:
        console.log(1);
}
```

To compensate for this, we could just add a `break` to the last case if it's not default.

src/transformation/visitors/switch.ts
```diff
@@ -1,61 +1,139 @@
...
+            const previousClause: ts.CaseOrDefaultClause | undefined = statement.caseBlock.clauses[i - 1];
 
-        // If the clause condition holds, go to the correct label
-        const condition = lua.createBinaryExpression(
-            switchVariable,
-            context.transformExpression(clause.expression),
-            lua.SyntaxKind.EqualityOperator
-        );
+            // Skip redundant default clauses, will be handled in final default case
+            if (i === 0 && ts.isDefaultClause(clause)) continue;
```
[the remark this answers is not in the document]
thejustinwalsh: It is, updating tests and working on a fix.

test/unit/__snapshots__/switch.spec.ts.snap
```diff
@@ -1,53 +1,66 @@
...
-    ::____switch3_end::
-    return result
+    local out = {}
+    repeat
+        local ____switch3 = 0
+        local ____cond3 = ((____switch3 == 0) or (____switch3 == 1)) or (____switch3 == 2)
+        if ____cond3 then
+            __TS__ArrayPush(out, \\"0,1,2\\")
+        end
+        if ____cond3 then
```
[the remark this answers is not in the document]
thejustinwalsh: The case for the default clause when it is fallen into will always be the condition for the previous condition. There is no additional condition that is needed to be satisfied. In the case above if you passed in any number > 2 it would not satisfy the condition of fallthrough into default and the `not ____cond` default at the bottom is would be triggered.

The `not ___cond` clause is the execution path for default unless there is a situation where default can be fallen into.

Rather than

src/transformation/visitors/switch.ts
```diff
@@ -1,61 +1,139 @@
...
+                    isInitialCondition
+                        ? lua.createVariableDeclarationStatement(conditionVariable, condition)
+                        : lua.createAssignmentStatement(
+                              conditionVariable,
+                              lua.createBinaryExpression(conditionVariable, condition, lua.SyntaxKind.OrOperator)
+                          )
+                );
+                isInitialCondition = false;
+            } else {
+                assert(!isInitialCondition, "Default clause should never be the initial condition");
```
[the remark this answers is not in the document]
thejustinwalsh: True. Initially, the assert was here to prove to me that the code was correct and this was not happening. Maybe just removing the block entirely?

How would I go about testing a situation with two defaults to ensure something sane happens.

src/transformation/visitors/switch.ts
thejustinwalsh: Sure! Could be easier on the eyes here.

src/transformation/visitors/switch.ts
thejustinwalsh: Is it the word `amalgamate`...?

I can work to comment through this section better. Do you have any suggestions on how to make it feel less black magic?

Essentially, we have to reconstruct the default clause and combine all of the statements from the clauses that the default can fall into.

Without a goto statement, this was the only approach I could dream up, as discovered by the initial try at this where combining clauses was proven to pass all tests too.

src/transformation/visitors/switch.ts
```diff
@@ -1,61 +1,139 @@
...
+                        context.transformExpression(clause.expression),
+                        lua.SyntaxKind.EqualityOperator
+                    );
+                }
 
-    if (concatenatedIf) {
-        statements.push(concatenatedIf);
-    }
+                // Skip empty clauses
+                if (clause.statements.length === 0) continue;
```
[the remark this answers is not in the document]
thejustinwalsh: Test agrees. Working it out now.

src/transformation/visitors/switch.ts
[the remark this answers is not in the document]
Perryvw: Just removing it would be fine, it might produce some weird lua but we don't guarantee working lua if you have broken TS anyway. I don't think there is really a nice way to check this for right now, and tbh I don't really care about this situation either, as long as we don't just throw and exit.

src/transformation/visitors/switch.ts
```diff
@@ -1,61 +1,139 @@
...
 import { performHoisting, popScope, pushScope, ScopeType } from "../utils/scope";
 
-export const transformSwitchStatement: FunctionVisitor<ts.SwitchStatement> = (statement, context) => {
-    if (context.luaTarget === LuaTarget.Universal || context.luaTarget === LuaTarget.Lua51) {
-        context.diagnostics.push(unsupportedForTarget(statement, "Switch statements", LuaTarget.Lua51));
+const containsBreakOrReturn = (statements: ts.Node[]): boolean => {
+    for (const s of statements) {
+        if (ts.isBreakStatement(s) || ts.isReturnStatement(s)) {
+            return true;
+        } else if (!ts.isBlock(s)) {
```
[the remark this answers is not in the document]
thejustinwalsh: Thanks for this, was too caught up in other parts of the code to validate this works. 

It is an optimization really, and the test given fails in the same way that the version without the break would.

`SyntaxList` you say! haha, I'll work on adding tests and ensuring this all works.

src/transformation/visitors/switch.ts
Perryvw: > Essentially, we have to reconstruct the default clause and combine all of the statements from the clauses that the default can fall into.

This would be a good start, maybe also explaining what (and why) end is.

src/transformation/visitors/switch.ts
```diff
@@ -1,61 +1,139 @@
...
-    statements.push(lua.createLabelStatement(`${switchName}_end`));
+        // Amalgamate the default w/ fallthrough clauses and execute if nothing else executed above
+        const start = clauses.findIndex(c => ts.isDefaultClause(c));
+        if (start >= 0) {
+            const end = statement.caseBlock.clauses
+                .slice(start)
+                .findIndex(clause => containsBreakOrReturn([...clause.statements]));
+            const defaultStatements = statement.caseBlock.clauses
+                .slice(start, end >= 0 ? end + start + 1 : undefined)
+                .reduce<lua.Statement[]>(
```
Perryvw, line 113: I see that this reduce is nice, but I would prefer if you just made defaultStatements an array and then do:
```ts
for (const { statements } of statement.caseBlock.clauses.slice(start, end >= 0 ? end + start + 1 : undefined)) {
    defaultStatements.push(...statements)
}
```
Because right now you are doing a lot of duplicate copying of arrays into each other.

src/transformation/visitors/switch.ts
thejustinwalsh: So if the default clause was fallen into from the block above, the default clause has already run. I don't want to execute it twice, right?

```ts
switch (1) {
    case 1:
        console.log("1 falls through")
    default:
        console.log("default is triggered and condition is true")
}

// Do not run default again
```

tomblind: Do we need the 2 default case handlers in this example?

thejustinwalsh: Oh, I understand now.
We can detect when we need the `not ___cond` default and potentially prune it.

```ts
switch (0) {
    case 1:
        console.log("1 falls through")
    default:
        console.log("default is triggered and condition is true")
    case 3:
        console.log("default falls through")
}
```

You would need the `not ____cond` default in this case.

tomblind: Right, that's the edge case. But what I'm suggesting is instead of doing the `not ____cond`, put a `break` at the end of `case 3`. This would only be necessary if the last case is not default and does not already break itself. At this point I'm getting into less-important optimizations, though 😄

src/transformation/visitors/switch.ts
```diff
@@ -1,61 +1,160 @@
...
+                        ? lua.createVariableDeclarationStatement(conditionVariable, condition)
+                        : lua.createAssignmentStatement(
+                              conditionVariable,
+                              lua.createBinaryExpression(conditionVariable, condition, lua.SyntaxKind.OrOperator)
+                          )
+                );
+                isInitialCondition = false;
+            } else {
+                // If the default is not fallen into skip it and handle in the final default output instead
+                if (previousClause && containsBreakOrReturn(previousClause.statements)) {
```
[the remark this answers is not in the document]
thejustinwalsh: Yeah, I don't need that. Must of refactored it a few times and didn't catch that.

thejustinwalsh:
> There's still a redundant default clause, although in a different way now:
> 
> ```lua
> local ____switch2 = x
> local ____cond2 = ____switch2 == 1
> if ____cond2 then
>     print(1)
> end
> if ____cond2 then
>     print("default")
>     break
> end
> do
>     print("default")
> end
> ```
> 
> Would just skipping the condition default clause if it's the last clause in the switch solve this? Or are there other cases that would break?

You know more tests might no the answer... ;) Lets find out.

tomblind:
Sorry, found another way to break it:
```ts
let y = 0;
function foo() {
    return y++;
}

let x = 0;
switch (x) {
    case 1:
        console.log(1);
    case foo():
    default:
        console.log("default");
}
console.log(y); // should print 1, but doesn't because foo() is never evaluated
```

tomblind:
maybe instead of trying to coalesce the conditions for empty statements, we can just eval them separately to keep the logic simpler:
```ts
switch (x) {
    case 1:
    case 2:
    case 3:
        // clause for case 3
}
```
```lua
____cond = ____cond or ____switch == 1
____cond = ____cond or ____switch == 2
____cond = ____cond or ____switch == 3
if ____cond then
    --clause for case 3
end
```

thejustinwalsh:
@tomblind can't believe that slipped through the tests. I swear that WAS working in one of these revisions.
Also, no way, too much time was put into making perfect Lua output. We must push forth.

thejustinwalsh:
@tomblind I mislead myself with my own comment, and after reading through the code again immediately realized the missing condition. I don't think I have it in me to hoist the second check to the initial condition in this special case though, as it would require another refactor or complicate it further. LMK.

tomblind (approved):
ok, I've run out of ways to break it 😄

Issue #1109, opened 2021-08-28
Bundle mode with sourceMapTraceback does not replace `{#SourceMapTraceback}`


Toliak:
# Description

Enabling bundle-mode (`luaBundle` and `luaBundleEntry`) with `sourceMapTraceback`) will cause to appear `{#SourceMapTraceback}` in the result file.

## Steps to reproduce

1. Create a project with dependencies:
```json
{
  "typescript": "^4.4.2",
  "typescript-to-lua": "^1.0.0"
}
```
2. Create `tsconfig.json` with listed content:
```json
{
  "$schema": "https://raw.githubusercontent.com/TypeScriptToLua/vscode-typescript-to-lua/master/tsconfig-schema.json",
  "compilerOptions": {
    "target": "es5",
    "module": "commonjs",
    "esModuleInterop": true,
    "forceConsistentCasingInFileNames": true,
    "strict": true,
    "skipLibCheck": true
  },
  "tstl": {
    "luaTarget": "JIT",
    "sourceMapTraceback": true,
    "luaBundleEntry": "src/index.ts",
    "luaBundle": "dist/index.lua"
  }
}
```
3. Create an empty `src/index.ts` file
4. Run `tstl`
5. Check the received `dist/index.lua`. It looks like:
```lua
--[[ Generated with https://github.com/TypeScriptToLua/TypeScriptToLua ]]

local ____modules = {}
local ____moduleCache = {}
local ____originalRequire = require
local function require(file)
    --- require code here, ok .....
end
____modules = {
["lualib_bundle"] = function() 
    --- lualib code here, ok .....
end,
["index"] = function() --[[ Generated with https://github.com/TypeScriptToLua/TypeScriptToLua ]]
require("lualib_bundle");
{#SourceMapTraceback}     -- <<--- Look here
 end,
}
return require("index")
```

# Why does it happen

The "macro-string" `{#SourceMapTraceback}` should be replaced here:
https://github.com/TypeScriptToLua/TypeScriptToLua/blob/691c9cc7bde8dfd89f44e61cf0777e4bdfbd3fcd/src/LuaPrinter.ts#L150-L153

## Debugging

I will try to debug the compiler and list the ideas below. I believe it will help to determine the problem,

[...]

PR #1110, opened 2021-08-28
Fixes bundle mode with enabled sourceMapTraceback

Toliak:
See #1109 for details


src/transpilation/bundle.ts
```diff
@@ -77,11 +77,11 @@ export function getBundleResult(program: ts.Program, files: ProcessedFile[]): [t
     ];
 }
 
-function moduleSourceNode({ code, sourceMapNode }: ProcessedFile, modulePath: string): SourceNode {
+function moduleSourceNode({ code }: ProcessedFile, modulePath: string): SourceNode {
     const tableEntryHead = `[${modulePath}] = function() `;
     const tableEntryTail = " end,\n";
 
-    return joinSourceChunks([tableEntryHead, sourceMapNode ?? code, tableEntryTail]);
```
Perryvw, line 84: Seems like you're just throwing away the source map information of the file? Did you test this? Seems to me like this would just break the source map completely.

Perryvw (changes requested):
Would expect to see at least one test added to verify this actually fixes the problem.

src/transpilation/bundle.ts
Toliak: The code already contains source map information. I'll figure it out and write tests tomorrow.

Toliak: I've changed the source map tracebacks behavior for bundles. The details are below

Toliak:
## Main changes
- `__TS__originalTraceback`  is able to save two types of source maps.  
  **First type** is default: `Record<number, number>` lua line is being mapped into TS line.  
  **The second type** is: `Record<number, { line : number, file: string }`: lua line is mapped into TS line and source file name.
  The first type has been already applied to single file compilation. And I have applied the second type for bundle compilation. The reason -- in bundle mode runtime Lua should know about target file (because bundle filename != source file)
- I have added `printStackTraceBundleOverride` function into `bundle.ts`. This function generates the **second type** of source map.
- `LuaPrinter`'s method `printFile` will not generate `{#SourceMapTraceback}` in bundle mode anymore. Because Source Map Traceback should be constructed for the whole bundled file
- `getBundleResult` function inserts the traceback before the `entryPoint`

## Misc changes
- Moved `{#SourceMapTraceback}` magic-macro-string into a static field `sourceMapTracebackPlaceholder` of `LuaPrinter` class
- Haven't touched `sourceMapNode` in moduleSourceNode (because it is not necessary)
- Moved helpful `lineAndColumnOf` from `test\unit\printer\sourcemaps.spec.ts` into `test\unit\printer\utils.ts`

## Tests coverage
- Moved default bundle test into `bundle/bundle-two-files`
- Tested source map generation with `bundle/bundle-source-maps` test
- Both tests are located in `test\transpile\bundle.spec.ts` and wrapped into `describe`

PR #1116, opened 2021-09-03
Preceding Statements

### LINKED ISSUE #654: Multi-statement expressions
The missing ability to transform an expression to multiple statements is a blocker for many features:

- Removing IIFEs around many common JS patterns
- ...which would be even more important once we'll get to [optional chaining proposal](https://github.com/tc39/proposal-optional-chaining)
- Implicit vararg optimization https://github.com/TypeScriptToLua/TypeScriptToLua/pull/628#issuecomment-500520700
- Many other optimizations we could do
- As noted in https://github.com/TypeScriptToLua/TypeScriptToLua/issues/614#issuecomment-508148732, JS truthiness

[roblox-ts implementation](https://github.com/roblox-ts/roblox-ts/pull/340) could be useful as a reference.

Some ideas:

Each `transformStatement` (and some other functions) should push a new context. A new method would push statements there. All extra statements would be prepended to result when `transformStatement` finishes. So:

```ts
foo(x ? x() : y());
```

->

```lua
local ____0
if x then
    ____0 = x()
else
    ____0 = y()
end
foo(____0)
```

where related part of `transformConditionalExpression` would look like:

```ts
const temporaryIdentifier = tstl.createIdentifier("____0");
this.pushStatements(/* create preceding statements with temporaryIdentifier */);
return temporaryIdentifier;
```

The more complex part begins when we have expressions that may have side effects, for example:

```ts
foo()(x ? x() : y());
```

the expected result for that would be:

```lua
local ____0 = foo()
local ____1
if x then
    ____1 = x()
else
    ____1 = y()
end
____0(____1)
```

I have few ideas about possible implementations:

1. When `transformStatement` finishes, and it's context has any preceding statements, traverse result, find all side-effect expressions and extract it to variables.

    I'm not really sure that it's possible to implement, the traversing function should be quite complex to be able to extract it without breaking expected flow.

2. Before pushing new context, traverse TS AST, and, if there is some expression that requires multi-statement transform, set a flag in the context. If that flag is set, any side-effect expression would be extracted to a variable.

    This one should be simpler, but it doesn't sound as a good design, since we'll have to implement logic for deciding it twice (during initial traverse and actual transform).

3. Transform expression as usual, and, if we have any preceding statements, set some flag and transform it again.

    So, it's like the second option, but doesn't require additional logic for traversing. However that means that we should be sure that our transform functions don't have any side effects, because original result would be left unused.

### LINKED ISSUE #1127: Incorrect destructuring assignment order
This will produce different results in TSTL vs TS:
```ts
const a: string[] = [];
[a[0], a[0]] = ["A", "B"];
console.log(a[0]); // "A" in Lua, "B" in JS
```
[Playground](https://typescripttolua.github.io/play/#code/MYewdgzgLgBAhgLhtATgSzAcwNoF0YC8MeA3AFDZzYAMuANPDbvkdgEQCCbDbAQm7nKhIIADYBTAHSiQmABRVaASnJA)

This is because of a peculiar behavior of Lua where, even though expressions are evaluated in the same order as JS, the actual assignments happen in reverse order. This is demonstrated with this:
```ts
const a: string[] = setmetatable([], {__newindex(k, v) { console.log(v); }});
function first() { console.log("first"); return 0; }
function second() { console.log("second"); return 1; }
function third() { console.log("third"); return "A"; }
function fourth() { console.log("fourth"); return "B"; }
[a[first()], a[second()]] = [third(), fourth()]; // first, second, third, fourth, B, A
```
[Playground](https://typescripttolua.github.io/play#code/MYewdgzgLgBAhgLhtATgSzAcwNoF0YC8yAplALalxRwBGANsQBR4A0MA3gPqdjEDuGACbEAHowDWbAG4BKDjFCQQDAHR0QmRrIDcMAL56Z2gFAAzAK5hgUNOBim0KaIznsF4CMuJqNjAEQOTlB+RjAopOYoYDAADLp6ZpbWttEQxIqCLvKKnqrqmn5pGSG64VCR0QCM8YlWNnZQABaOma7uSnm+fk0tJWERUTB+AIJ+NRZ1KfYgkU1ZbjlePgWmMyhNfWUVQwBCY-rG2HDYgc4yuGzHReCtuPhE2D0orWyrs40uuCZAA)

### LINKED ISSUE #1142: using __TS_ArrayConcat for spreads breaks when nil values are present
```ts
const a = ["a"];
let b: unknown = null;
const result = [...a, b, "c"];
print(result[0], result[1], result[2]); // Should be "a", nil, "c"
```
[Playground](https://typescripttolua.github.io/play#code/MYewdgzgLgBAhjAvDA2gIjmgugbgFAA2AprAEYBcMArmANZggDuYSMYVBB+oksATkQgdYyFADoJcADQxSMtMGz4ADnwCWYKAAoBQglBQAGLDN3CUARhMwz+lACYsAShwwA9G5gBlABYgOACayRDAYaDJgagTyinhAA)

tomblind:
Closes #654

This PR fully replaces our IIFEs with preceding statements. To do this, it uses a stack based system akin to what is detailed out here: roblox-ts/roblox-ts#537. Statements can be added before the current expression being transformed using `context.addPrecedingStatements()`. If nodes higher up want to catch and manipulate these statements, they can do so by pushing and popping a statement collection with `context.pushPrecedingStatements()` and `context.popPrecedingStatements()`.

Significant refactoring  was required in a few places to pull this off. Call expression handling has been modified a lot, but should be understandable. Dealing with assignments was particularly tricky, though, as there's a complex web of cases (especially destructuring assignments). At some point I imagine a full rewrite of assignment handling might be needed to keep it understandable and maintainable, but I'm not even sure what that would look like right now 😅 

## Evaluation Order
The biggest challenge with this system is maintaining evaluation order in statements with multiple expressions. Consider this example:
```ts
function foo(a: number, b: number) {
    let i = 0;
    let o = {
        method(...args: unknown[]) { /* ... */ }
    };

    function bar(): number {
        // Could modify a, b, i, o, or o.method!
    }

    o.method(a, b, i += bar());
}
```
Once `i += bar()` is lifted into a preceding statement, any expressions before it in the statement could be affected, and thus need to be cached in temps ahead of time to maintain evaluation order.

Unfortunately, without deeper inspection of the preceding statements being generated, we have to aggressively do this caching, even in cases which don't seem to need it upon visual inspection. For example:
```ts
declare function foo(x: number): void;
let i = 0;
foo(i++);
```
Transpiles to:
```lua
local i = 0
local ____foo_2 = foo
local ____G_1 = _G
local ____i_0 = i
i = ____i_0 + 1
____foo_2(____G_1, ____i_0)
```
Both `foo` and even `_G` have to be cached in temps because we can't guarantee that the preceding statements generated by `++i` won't modify them. It's possible we could add some specific optimizations for cases like this to reduce the temps, but it would be tricky and a bit messy. Also, thanks to operator overloading, it is possible `++i` could have actually modified those values.

`const` variables won't be cached like this however. Those looking to reduce temps should prefer `const` over `let` whenever possible.

## Sparse Array Lib
To help with situations which could generate extreme numbers of temps, this PR also adds some new lib functions for manipulating sparse arrays. Consider this:
```ts
foo(a, b, c, d, e++, f, g);
```
Instead of caching all of those variables in temps, we can push them into an array while preserving evaluation order and unpack them into the call's arguments. But, since any of these could have nil values, we use a custom array that tracks its true size.
```lua
local ____foo_2 = foo
local ____array_1 = __TS__SparseArrayNew(_G, a, b, c, d)
local ____e_0 = e
e = ____e_0 + 1
__TS__SparseArrayPush(____array_1, ____e_0, f, g)
____foo_2(
    __TS__SparseArraySpread(____array_1)
)
```
This technique can also be used to handle spread expressions in the middle of arguments, so it also fixes #1142 as a bonus.

Right now, argument lists and array expressions which would generate 3 or more temps when transformed will revert to this behavior. That number can be adjusted, if needed.

## Other Notes
- Some helpers for generating useful temp names has been added to make the output lua a bit more readable
- Some things like optional chaining could be reworked to take advantage of this, but I'll save those for separate PRs
- This also fixes #1127


GlassBricks:
When calling a function with preceding statement arguments, the function itself and the `self` parameter also need to be stored to temp variables.

I've made a PR to the _source_ branch with tests for this:
https://github.com/tomblind/TypescriptToLua/pull/1

src/transformation/visitors/access.ts
```diff
@@ -8,39 +8,45 @@ import { addToNumericExpression } from "../utils/lua-ast";
...
     const type = context.checker.getTypeAtLocation(node.expression);
     const argumentType = context.checker.getTypeAtLocation(node.argumentExpression);
     if (isStringType(context, type) && isNumberType(context, argumentType)) {
-        const index = context.transformExpression(node.argumentExpression);
-        return transformLuaLibFunction(context, LuaLibFeature.StringAccess, node, table, index);
+        return transformLuaLibFunction(context, LuaLibFeature.StringAccess, node, table, accessExpression);
     }
 
-    const accessExpression = transformElementAccessArgument(context, node);
+    accessExpression = getElementAccessArgument(context, node, accessExpression);
```
Perryvw, line 49: This no longer makes sense to me, why are we overwriting this here rather than making a new variable or using the old one?

src/transformation/visitors/conditional.ts
```diff
@@ -27,32 +27,34 @@ function canBeFalsy(context: TransformationContext, type: ts.Type): boolean {
...
+    const falseStatements = context.popPrecedingStatements();
+    falseStatements.push(lua.createAssignmentStatement(lua.cloneIdentifier(tempVar), val2, expression.whenFalse));
 
-    // (condition and (() => v1) or (() => v2))()
-    const conditionAnd = lua.createBinaryExpression(condition, val1Function, lua.SyntaxKind.AndOperator);
-    const orExpression = lua.createBinaryExpression(conditionAnd, val2Function, lua.SyntaxKind.OrOperator);
-    return lua.createCallExpression(orExpression, [], expression);
+    context.addPrecedingStatements([
+        lua.createVariableDeclarationStatement(tempVar, undefined, expression.condition),
+        lua.createIfStatement(
```
Perryvw, line 50: Wouldn't this cause the condition and one of  the branches to be evaluated even if they should not be, if they are part of an expression like for example `const a = true || (a ? b : c)`?

src/transformation/visitors/conditional.ts
```diff
@@ -78,8 +80,15 @@ export function transformIfStatement(statement: ts.IfStatement, context: Transfo
 
     if (statement.elseStatement) {
         if (ts.isIfStatement(statement.elseStatement)) {
+            context.pushPrecedingStatements();
```
Perryvw, line 83: I don't get why this is here. I assumed you were now forced to pop preceding statements whenever we visit a block, but then why is that not done on the `ifBlock` in this function?

src/transformation/visitors/loops/do-while.ts
```diff
@@ -1,23 +1,51 @@
...
-        statement
-    );
+export const transformWhileStatement: FunctionVisitor<ts.WhileStatement> = (statement, context) => {
+    const body = transformLoopBody(context, statement);
+
+    context.pushPrecedingStatements();
+    let condition = context.transformExpression(statement.expression);
+    const precedingStatements = context.popPrecedingStatements();
+
+    // Change from 'while condition' to 'while true - if not condition then break'
```
Perryvw, line 13: I guess the reason for this is to re-evaluate the preceding statements every iteration? Could you add that as a comment?

src/transformation/visitors/binary-expression/destructuring-assignments.ts
```diff
@@ -36,28 +38,31 @@ export function isArrayLength(
...
-    return transformAssignmentPattern(context, node.left, root);
+    return transformAssignmentPattern(context, node.left, root, rootHasPrecedingStatements);
 }
 
 export function transformAssignmentPattern(
     context: TransformationContext,
     node: ts.AssignmentPattern,
-    root: lua.Expression
+    root: lua.Expression,
+    rootHasPrecedingStatements: boolean
```
Perryvw, line 51: Why does this have to be a separate argument, couldnt we just inspect the context to see if this is the case?

src/transformation/visitors/function.ts
```diff
@@ -53,8 +52,10 @@ function isRestParameterReferenced(identifier: lua.Identifier, scope: Scope): bo
 
 export function transformFunctionBodyContent(context: TransformationContext, body: ts.ConciseBody): lua.Statement[] {
     if (!ts.isBlock(body)) {
+        context.pushPrecedingStatements();
         const returnStatement = transformExpressionBodyToReturnStatement(context, body);
-        return [returnStatement];
+        const precedingStatements = context.popPrecedingStatements();
+        return [...precedingStatements, returnStatement];
```
Perryvw, line 58: I see this same pattern recurring a lot, it might be nice to refactor this to something like this?
```ts
function transformInPrecedingStatementScope(context: TransformationContext, transformer: () => lua.Statement | lua.Statements[]) {
        context.pushPrecedingStatements();
        const statementOrStatements = transformer();
        const precedingStatements = context.popPrecedingStatements();
        return [precedingStatements, statementOrStatements];
}
```
So this function could write just
```ts
const [precedingStatements, returnStatement] = transformInPrecedingStatementScope(context, () => transformExpressionBodyToReturnStatement(context, body));
```
(Assuming you could reuse the same refactor at other occurrences of this push/pop/concat pattern)

src/transformation/visitors/sourceFile.ts
```diff
@@ -13,7 +13,10 @@ export const transformSourceFileNode: FunctionVisitor<ts.SourceFile> = (node, co
         const [statement] = node.statements;
         if (statement) {
             assert(ts.isExpressionStatement(statement));
-            statements.push(lua.createReturnStatement([context.transformExpression(statement.expression)]));
+            context.pushPrecedingStatements();
+            const expression = context.transformExpression(statement.expression);
+            statements.push(...context.popPrecedingStatements());
+            statements.push(lua.createReturnStatement([expression]));
```
Perryvw, line 19: Also seems like a  candidate for the push/pop/concat refactor I mentioned earlier

Perryvw (commented):
image
I have skipped the more complicated uses cases (such as call/assignment etc) for now. I hope discussing/reworking the comments I have placed now will give me enough understanding to tackle the remaining files.

lolleko:
Wow, great work! First of all, sorry that I probably won't have the time to review this.
I just have one quick question regarding the "caching" of tables like `self` and `_G` (as described in the PR description).
Since assigning a table only copies the reference, not the table, does that even work/make sense?

src/transformation/visitors/conditional.ts
tomblind: That wouldn't happen because the short-circuit logic for the or `||` opertaor would wrap the right side in another if clause. Here's the output of your example:
```lua
-- const x = true || (a ? b : c)
local ____temp_1 = true
if not ____temp_1 then
    local ____a_0
    if a then
        ____a_0 = b
    else
        ____a_0 = c
    end
    ____temp_1 = ____a_0
end
x = ____temp_1
```

src/transformation/visitors/conditional.ts
tomblind: If an `else if` statement generates preceding statements, it needs to execute those in an else clause before doing the second conditional.
Example:
```ts
if (i) {
} else if (i) { // No preceding statements
}

if (i) {
} else if (i++) { // i++ creates preceding statements
}
```

```lua
if i then
elseif i then -- elseif is fine
end

if i then
else -- elseif won't work - we do an else, execute preceding statements, then do the if check
    local ____i_0 = i
    i = ____i_0 + 1
    if ____i_0 then
    end
end
```

I'll add some comments here explaining this.

src/transformation/visitors/function.ts
tomblind: Good call. I was able to replace a ton of instances of this pattern with the helper.

tomblind:
@lolleko 
Thanks! Regarding tables, what is assigned to the variable name is what might be changed by preceding statements, so we need to cache what was originally assigned to prevent incorrect behavior.

Consider this:
```ts
function changeG(this: void) { _G = {}; return 1; };
foo(i += changeG());
```
```lua
function changeG()
    _G = {}
    return 1
end
local ____foo_1 = foo
local ____G_0 = _G
i = i + changeG()
____foo_1(____G_0, i)
```

`changeG()` should be executed after `_G` was already evaluated as the first parameter here, but it's bumped back because of preceding statements, so `foo` would now be called with the new value of `_G` instead of the old one. So we have to cache the old one and pass that instead.

src/transformation/visitors/loops/do-while.ts
```diff
@@ -1,23 +1,53 @@
...
-        statement
+export const transformWhileStatement: FunctionVisitor<ts.WhileStatement> = (statement, context) => {
+    const body = transformLoopBody(context, statement);
+
+    let [precedingStatements, condition] = transformInPrecedingStatementScope(context, () =>
+        context.transformExpression(statement.expression)
     );
 
+    // Change from 'while condition' to 'while true - [preceding statements] if not condition then break'
+    // This is so that preceding statements are evaluated every iteration.
```
Perryvw, line 15: Could you specify here this is about the preceding statements of the condition? Could you maybe also add this 'reason' comment to the other instances where you do this? (line 39 of this file, 31 in for.ts)

src/transformation/visitors/binary-expression/assignments.ts
```diff
@@ -55,15 +69,16 @@ export function transformAssignment(
...
+        lhs.parent && ts.isShorthandPropertyAssignment(lhs.parent)
+            ? context.checker.getShorthandAssignmentValueSymbol(lhs.parent)
+            : context.checker.getSymbolAtLocation(lhs);
 
     const dependentSymbols = symbol ? getDependenciesOfSymbol(context, symbol) : [];
 
-    const left = transformAssignmentLeftHandSideExpression(context, lhs);
+    const left = transformAssignmentLeftHandSideExpression(context, lhs, rightHasPrecedingStatements);
 
-    const rootAssignment = lua.createAssignmentStatement(left, right, lhs.parent);
```
Perryvw, line 66: Is this changed correct? Wouldn't this 'break' source maps?

src/transformation/visitors/binary-expression/assignments.ts
```diff
@@ -179,7 +190,9 @@ const canBeTransformedToLuaAssignmentStatement = (
         }
 
         if (ts.isPropertyAccessExpression(element) || ts.isElementAccessExpression(element)) {
-            return true;
```
Perryvw, line 182: This is a pretty drastic change, was this incorrect before, or what changed for this to get the opposite value now?

src/transformation/visitors/expression-list.ts
```diff
@@ -0,0 +1,189 @@
...
+    context: TransformationContext,
+    expression: lua.Expression,
+    tsOriginal?: ts.Node
+): lua.Expression {
+    if (!shouldMoveToTemp(context, expression, tsOriginal)) {
+        return expression;
+    }
+    const tempIdentifier = context.createTempNameForLuaExpression(expression);
+    const tempDeclaration = lua.createVariableDeclarationStatement(tempIdentifier, expression);
+    lua.setNodePosition(tempDeclaration, lua.getOriginalPos(expression));
```
Perryvw, line 27: Doesn't the lua create* function already do this?

src/transformation/visitors/expression-list.ts
```diff
@@ -0,0 +1,189 @@
...
+    context.addPrecedingStatements(tempDeclaration);
+    const tempClone = lua.cloneIdentifier(tempIdentifier);
+    lua.setNodePosition(tempClone, lua.getOriginalPos(tempIdentifier));
+    return tempClone;
+}
+
+function transformExpressions(
+    context: TransformationContext,
+    expressions: readonly ts.Expression[]
+): [lua.Expression[], lua.Statement[][], number] {
```
Perryvw, line 37:
```suggestion
): { transformedExpressions: lua.Expression[], precedingStatements: lua.Statement[][], lastPrecedingStatementsIndex: number} {
```

src/transformation/visitors/expression-list.ts
```diff
@@ -0,0 +1,189 @@
...
+    const precedingStatements: lua.Statement[][] = [];
+    const transformExpressions: lua.Expression[] = [];
+    let lastPrecedingStatementsIndex = -1;
+    for (let i = 0; i < expressions.length; ++i) {
+        context.pushPrecedingStatements();
+        transformExpressions.push(context.transformExpression(expressions[i]));
+        const expressionPrecedingStatements = context.popPrecedingStatements();
```
Perryvw, line 44: Guess you could still use the `transformInPrecedingStatementScope` helper here

Perryvw (commented):
image

GlassBricks:
Nullish coalescing is not implemented with preceding statements.

This led me to see a larger problem: `transformBinaryOperation` (in `visitors/binary-expression/index.ts`) takes left and right lua.expressions rather than ts expressions. This means there is no way to obtain the preceding statements, and every usage of transformBinaryExpression that uses a short-circuit binary operation is potentially wrong.

E.g. see this failing test:
```
test("nullish-coalescing assignment expression with side effect rhs", () => {
    util.testFunction`
        let i = 0;
        let t = { foo: 0 }
        const bar = (t.foo ??= i++)
        return [i, t.foo, bar]
    `
        .expectToMatchJsResult();
});
```

GlassBricks:
If I can comment further, I think an issue with this preceding statements implementation is that it makes the code _much_ more stateful; _every_ transform now has the possibility of adding preceding statements, and so can easily lead bugs like above.

An idea for a less stateful implementation would be to introduce some interface ExpressionAndPrecedingStatements (with a shorter name) which contains both a lua.Expression and preceding statements needed when evaluated, and have parts of the code base that return `lua.Expression` return that interface instead. This way preceding statements would need to be handled explicitly, and can be treated together with an expression, e.g. the logic of`transformBinaryOperation` doesn't need to be modified too much to work with preceding statements.

Utilities such as `transformExpressionAndPutStatements(ts.Expression, precedingStatements: lua.Statement[]` could help with simple cases.

This would touch more of the existing code, but might be worth it. Preceding statements already has an impact on almost all parts of transformation, and need be accounted for in every single transform.

lolleko:
@tomblind 
> `changeG()` should be executed after `_G` was already evaluated as the first parameter here, but it's bumped back because of preceding statements, so `foo` would now be called with the new value of `_G` instead of the old one. So we have to cache the old one and pass that instead.

Thanks for clarifying, I understand now.

src/transformation/visitors/binary-expression/assignments.ts
tomblind: I changed this temporarily at some point for a reason I can't even remember but it was supposed to be reverted. 😅

src/transformation/visitors/binary-expression/assignments.ts
tomblind: See #1127

tomblind:
@GlassBricks 

> every usage of transformBinaryExpression that uses a short-circuit binary operation is potentially wrong.

Good catch on this one. It's going to take some refactoring to get right. 😞 

> An idea for a less stateful implementation would be to introduce some interface 
ExpressionAndPrecedingStatements (with a shorter name) which contains a lua.Expression and preceding statements needed when evaluated.

This was my original plan for this, but it's a fundamental change that would require refactoring pretty much all transformations, so I avoided it in the hopes of keeping the scope of this PR down.

GlassBricks:
> This was my original plan for this, but it's a fundamental change that would require refactoring pretty much all transformations, so I avoided it in the hopes of keeping the scope of this PR down.

The scope of this PR is already pretty big... Even without refactoring everything, using this interface still may be beneficial.

Double-checking every function that takes a lua.Expression but not preceding statements, `transformNullishCoalescingOperator` in `binaryExpression/index.ts` is also potentially wrong (and still uses an IIFE).

Perryvw:
> > This was my original plan for this, but it's a fundamental change that would require refactoring pretty much all transformations, so I avoided it in the hopes of keeping the scope of this PR down.
> 
> The scope of this PR is already pretty big... Even without refactoring everything, using this interface still may be beneficial.
> 
> Double-checking every function that takes a lua.Expression but not preceding statements, `transformNullishCoalescingOperator` in `binaryExpression/index.ts` is also potentially wrong (and still uses an IIFE).

Right now the changes to expressions are very limited, all logic for combining and forwarding preceding statements is now spread over the statement transformers, which is already more than I would ideally like. Doing this change would have a very large impact because all expression transformers would need to be touched, and all of them would require to have some logic about how to forward preceding statements from its child expressions.

If we can make this PR change without doing this refactor, I would be in favor of that. I am interested in this alternative, but would prefer to judge it as an independent change.

tomblind:
The issue with `transformBinaryOperator` is fixed and tests have been added for it.

> transformNullishCoalescingOperator in binaryExpression/index.ts is also potentially wrong (and still uses an IIFE).

Thanks for reminding me on this. I had meant to put in a bug for it but forgot. Since it's not directly related to preceding statements, I think it would be best to fix this in a follow up PR.

src/transformation/visitors/switch.ts
```diff
@@ -18,30 +18,50 @@ const containsBreakOrReturn = (nodes: Iterable<ts.Node>): boolean => {
...
-    condition: lua.Expression | undefined,
-    switchVariable: lua.Identifier,
-    expression: ts.Expression,
-    context: TransformationContext
-): lua.Expression => {
+    condition: ts.Expression | undefined,
+    switchVariable: ts.Identifier,
+    expression: ts.Expression
+): ts.Expression => {
+    const comparison = ts.factory.createBinaryExpression(
```
Perryvw, line 25: This looks nice and simple, but don't all these ts.factory.create* methods interfere with source maps and potentially type checking that needs tree traversal?

tomblind: Good point. Wasn't fully considering that. I've reverted to using lua expressions here and manually handled preceding statements.

GlassBricks:
To comment on the above, I think you can use `ts.setOriginalNode` on a generated TS node to retain type information and source map position. This might be useful to reuse functionality of other complex transforms,

src/transformation/visitors/binary-expression/index.ts
```diff
@@ -76,6 +87,86 @@ export function transformBinaryOperation(
     return lua.createBinaryExpression(left, right, luaOperator, node);
 }
 
+export function createShortCircuitBinaryExpression(
```
Perryvw, line 90:
```suggestion
export function createShortCircuitBinaryExpressionPrecedingStatements(
```

src/transformation/visitors/binary-expression/compound.ts
```diff
@@ -85,74 +70,137 @@ export function transformCompoundAssignment(
...
         // local ____tmp = ${left};
         // ${left} = ____tmp ${replacementOperator} ${right};
         // return ____tmp
-        const tmpIdentifier = lua.createIdentifier("____tmp");
+        const tmpIdentifier = context.createTempNameForLuaExpression(left);
         const tmpDeclaration = lua.createVariableDeclarationStatement(tmpIdentifier, left);
-        const operatorExpression = transformBinaryOperation(context, tmpIdentifier, right, operator, expression);
-        const assignStatements = transformAssignment(context, lhs, operatorExpression);
+        let operatorExpression: lua.Expression;
+        [rightPrecedingStatements, operatorExpression] = transformBinaryOperation(
```
Perryvw, line 132: Personal opinion: I think reassigning existing variables here instead of just making consts in the scope where they are relevant is pretty ugly.

Perryvw (commented):
image

Perryvw (approved):

PR #1123, opened 2021-09-05
Use `ts.transformJsx` instead for jsx. 

GlassBricks:
Partially reverts commit 987899c8acdaf14668d234a8470c9f29a33f7459 "JSX support (#1052)".

This is an experiment with using typescript's own transformer instead of our own for JSX support. I think this could be better than what we are currently doing.

All JSX tests still pass, however, there is a backwards incompatibility: React.createElement now needs a `self` parameter, where it didn't previously. I couldn't find a good way to control this.

Advantages:
 - Low maintenance
 - Inherits all features
 - "reactJsx" option might be easy to add in the future
 - Integrates better with rest of features. e.g. this automatically fixes preceding statements (#1116), which was previously broken with JSX.

Disadvantages:
 - Less control over transformation
 - Slightly hacky


GlassBricks:
There are some unnecessary diff due to unnecessary formatting by my IDE, fixing...

src/transpilation/transformers.ts
```diff
@@ -86,40 +86,51 @@ function loadTransformersFromOptions(program: ts.Program, diagnostics: ts.Diagno
...
         }
     }
-
+    if (options.jsx === ts.JsxEmit.React) {
+        customTransformers.before.push(context => {
+            const patchedContext: ts.TransformationContext = {
+                ...context,
+                getCompilerOptions: () => ({
+                    ...context.getCompilerOptions(),
+                    target: ts.ScriptTarget.ESNext,
```
Perryvw, line 128: why is it required to overwrite compiler options here? could you add a comment explaining the reasoning?

Perryvw (commented):
Very nice, I'll gladly get rid of our custom jsx handling, this seems like the better solution to me.

src/transpilation/transformers.ts
GlassBricks: If the script target is < ES.2017, typescript does some additional unnecessary transformations relating to object spreading. I override it here so we always get the latest ES version features

GlassBricks:
Just remember to put the breaking change into the release notes:

`React.createElement` now **should** have a self parameter -- remove `@noSelf` or `this: void`, or add `this: unknown` if necessary.

PR #1125, opened 2021-09-06
Hoisting switch fix

### LINKED ISSUE #1099: Crash when hoisting inside a switch statement
```ts
let x = 0;
switch (x) {
    case 1:
        hoisted();
        break;
    case 2:
        function hoisted() {
            console.log("hoisted");
        }
        break;
}
```

It looks like hoisting inside a switch statement causes a crash inside TSTL:
```
Exception has occurred: AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:

  nativeAssert(value, message)
   at Object.assert (utils.js:67:5)
    at hoistVariableDeclarations (transformation\utils\scope.js:157:21)
    at Object.performHoisting (transformation\utils\scope.js:113:14)
    at Object.transformSwitchStatement [as transform] (transformation\visitors\switch.js:60:26)
    at TransformationContext.transformNode (transformation\context\context.js:44:62)
    at transformation\context\context.js:70:58
    at Array.flatMap (<anonymous>)
    at TransformationContext.transformStatements (transformation\context\context.js:70:40)
    at Object.transformSourceFileNode [as transform] (transformation\visitors\sourceFile.js:29:63)
    at TransformationContext.transformNode (transformation\context\context.js:44:62)
```

tomblind:
fixes #1099 

To fix hoisting in switch statements, I've added `separateHoistedStatements` as an intermediate step to `performHoisting`. This keeps the results separate, which `transformSwitchStatement` uses directly so it can properly order statements after all clauses are transformed.

Previously, switch cases had some custom code to handle hoisting all locals to the top. I've refactored this to use the standard hoisting code path and just force-hoist everything that is immediately in a switch's scope.

Additionally, there was some very hard to follow 'magic' which handled function hoisting so that it could be output as `local function foo()` in some cases and `local foo; function foo()` in others when needed. This was incredibly hard to follow in the codebase, so I've changed it to always use the later form which works in all cases and keeps the implementation a bit more straightforward.

### DIFF
--- a/src/transformation/utils/scope.ts
+++ b/src/transformation/utils/scope.ts
@@ -33,6 +33,11 @@ export interface Scope {
     functionReturned?: boolean;
 }
 
+export interface HoistingResult {
+    statements: lua.Statement[];
+    hoistedStatements: lua.Statement[];
+}
+
 const scopeStacks = new WeakMap<TransformationContext, Scope[]>();
 function getScopeStack(context: TransformationContext): Scope[] {
     return getOrUpdate(scopeStacks, context, () => []);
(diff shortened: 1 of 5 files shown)


thejustinwalsh (approved):
I think this is covered by other tests, but passes with expected results. LGTM!

```ts
let x = 1;
let y = 2;
let result = [];
switch (x) {
    case 1: {
        let y = 100;
        result.push(y); 
    }
    case 2: {
        let y  = 500;
        result.push(y);
    }
    case 3: {
        y = 50;
        result.push(y);
    } 
}
return result
 ```

File: src/transformation/utils/scope.ts [hunk 1]
```diff
@@ -179,69 +216,73 @@ function shouldHoistSymbol(context: TransformationContext, symbolId: lua.SymbolI
     return false;
 }
 
-function hoistVariableDeclarations(
-    context: TransformationContext,
-    scope: Scope,
-    statements: lua.Statement[]
-): lua.Statement[] {
+function hoistVariableDeclarations(context: TransformationContext, scope: Scope, statements: lua.Statement[]) {
```
Perryvw, line 219: Think I would prefer if you explicitly declared the return type of these hoist* functions for clarity - I guess it's just HoistingResult?

src/transformation/visitors/switch.ts
```diff
@@ -55,15 +55,25 @@ export const transformSwitchStatement: FunctionVisitor<ts.SwitchStatement> = (st
 
     // If the switch only has a default clause, wrap it in a single do.
     // Otherwise, we need to generate a set of if statements to emulate the switch.
-    let statements: lua.Statement[] = [];
+    const statements: lua.Statement[] = [];
+    const prefixStatements: lua.Statement[] = [];
```
Perryvw, line 59: Think it might be clearer if you just named this `hoistedStatements`? I think to avoid naming conflicts with your destructuring assignments is to just not destructure, I don't think that is really making things clearer when it's split over multiple lines like this anyway.

src/transformation/visitors/switch.ts
```diff
@@ -145,11 +166,24 @@ export const transformSwitchStatement: FunctionVisitor<ts.SwitchStatement> = (st
...
+            // Only push hoisted statements if this is the first time we're transforming the default clause
+            if (!defaultTransformed) {
+                prefixStatements.push(...hoistedStatements);
+                prefixIdentifiers.push(...hoistedIdentifiers);
+            }
+
+            // Combine the fallthrough statements
+            for (const clause of clauses.slice(start + 1, end >= 0 ? end + 1 : undefined)) {
+                let statements = context.transformStatements(clause.statements);
+                ({ statements } = separateHoistedStatements(context, statements));
```
Perryvw, line 184: what if there are hoisted statements in here, are they ignored? if so, where are those included?

tomblind: These are the duplicate clauses added to the end of the final default when it fell through into others. Re-adding their hoisted statements a 2nd time would generate bad lua. I've added a comment to explain this is intentional.

PR #1126, opened 2021-09-06
Runtime benchmarks

GlassBricks:
Adds basic runtime benchmarks, modeled off of existing memory benchmarks.

At least can be used for local benchmarks; Still need to test with github workflows.

### DIFF
--- a/benchmark/src/run.ts
+++ b/benchmark/src/run.ts
@@ -41,14 +41,20 @@ function benchmark(): void {
 }
 benchmark();
 
+function sortByName({ benchmarkName: a }: BenchmarkResult, { benchmarkName: b }: BenchmarkResult): number {
+    if (a < b) return -1;
+    if (a > b) return 1;
+    return 0;
+}
+
 function compareBenchmarks(oldResults: BenchmarkResult[], newResults: BenchmarkResult[]): ComparisonInfo {
-    const oldResultsMemory = oldResults.filter(isMemoryBenchmarkResult);
-    const newResultsMemory = newResults.filter(isMemoryBenchmarkResult);
+    const oldResultsMemory = oldResults.filter(isMemoryBenchmarkResult).sort(sortByName);
+    const newResultsMemory = newResults.filter(isMemoryBenchmarkResult).sort(sortByName);
 
     const memoryComparisonInfo = compareMemoryBenchmarks(oldResultsMemory, newResultsMemory);
 
-    const oldResultsRuntime = oldResults.filter(isRuntimeBenchmarkResult);
-    const newResultsRuntime = newResults.filter(isRuntimeBenchmarkResult);
+    const oldResultsRuntime = oldResults.filter(isRuntimeBenchmarkResult).sort(sortByName);
+    const newResultsRuntime = newResults.filter(isRuntimeBenchmarkResult).sort(sortByName);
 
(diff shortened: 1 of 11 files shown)


Perryvw:
Could you try uncommenting the benchmarks in: https://github.com/TypeScriptToLua/TypeScriptToLua/blob/master/.github/workflows/ci.yml? They are meant to be uncommented but I forgot it, and it would also be good to see what happens with your new benchmarks.

GlassBricks:
Are the benchmark output (markdown) supposed to be in a particular format for benchviz to work? In the CI, it displays a blank page.
Right now, I changed the formatting a bit on the benchmarks so my IDE views it better

benchmark/src/run.ts
```diff
@@ -32,13 +41,27 @@ function benchmark(): void {
...
 
     const memoryComparisonInfo = compareMemoryBenchmarks(oldResultsMemory, newResultsMemory);
 
-    return { summary: memoryComparisonInfo.summary, text: memoryComparisonInfo.text };
+    const oldResultsRuntime = oldResults.filter(isRuntimeBenchmarkResult).sort(sortByName);
+    const newResultsRuntime = newResults.filter(isRuntimeBenchmarkResult).sort(sortByName);
+
+    const runtimeComparisonInfo = compareRuntimeBenchmarks(oldResultsRuntime, newResultsRuntime);
+
+    return {
```
Perryvw, line 61: I believe this object is what is base64 encoded and sent to benchviz, it makes sense that just adding a newline breaks this. I'm okay with breaking benchviz for now, but we should make this a nice return type, maybe returning `ComparisonInfo[]` instead? (and add a benchmark name member), or maybe comparisonInfo should have some array internally? I want an array somewhere, no \n concatenation.

GlassBricks: That depends on how benchviz is meant to work. Could we just return the JSON result or something similar, and have benchviz format it? Although, the summary markdown text is useful for local viewing

Perryvw: This object as JSON is what is visualized by benchviz. Really this should just be the data, any visualization like benchviz or even printing it out as markdown should probably take this thing as input and display it however it sees fit.

GlassBricks: I'm thinking of creating an interface for "Benchmark result summary" as an intermediary, that gets both sent to benchviz and also is able to be formatted to markdown. Would that work?

GlassBricks: Also, I think the online benchviz is breaking for a different reason: I'm getting `ReferenceError: Buffer is not defined` in the console

Perryvw: Yeah I think so too, it seems broken for another reason, probably to do with our latest docusaurus update. I'm fine with keeping it broken, but like I said I don't want it to have to figure out what to show based on a \n somewhere in there. I propose something like this:
```ts
return {
    [BenchmarkKind.Memory]: {
        summary: memoryComparisonInfo.summary,
        text: memoryComparisonInfo.text,
    },
    [BenchmarkKind.Runtime]: {
        summary: runtimeComparisonInfo.summary,
        text: runtimeComparisonInfo.text,
    },
}
```
And adjusting return type to `Record<BenchmarkKind, ComparisonInfo>`, we can fix benchviz later.

GlassBricks: I made a separate PR on typescripttolua.github.io that fixes benchviz and adds support for runtime.

I'll make a separate commit with this proposal for the markdown output.

Issue #1121, opened 2021-09-05
Transpile String.replaceAll()


Zamiell:
I believe that the equivalent in Lua is:
```lua
local newString = string.gsub(oldString, "a", "b")
```

GlassBricks:
This might not work that well:
 - Regex in JS is _not_ the same as patterns in Lua (also see #543).
 - In the case of a string, gsub interprets the second arguments as a lua pattern, while replaceAll as a raw string.
 
This might be possible with a lualib function, that only works on replacing string with string, and "escapes" all characters in the match pattern. However, that sounds restrictive and nonperformant.

Would better to just use [lua-types](https://github.com/TypeScriptToLua/lua-types) and call string.gsub directly in your project

Cheatoid:
Nope, that's not appropriate equivalent.
Because `string.gsub` actually is aware of Lua patterns (with no option to disable it easily).
In case you don't know what I am talking about, here run this Lua code:
```lua
print( string.gsub( "Hello 1337", "%d+", "World" ) )
```
The result? `Hello World`

Zamiell:
I see. As for me, I don't care if the transpilation is implemented with `gsub` or something else, I just want the feature. :)

Cheatoid:
This could easily be added to lualib, assuming you only care about `String.replaceAll` operating in raw/plain string mode.
In the meantime, you can do this while waiting for this to be added into lualib; bundle/include the following Lua code into your environment (by any means necessary):
```lua
local string_len, string_sub, string_find, table_concat = string.len, string.sub, string.find, table.concat

local string_toTable = function(self)
  local tbl = {}
  for i = 1, string_len(self) do
    tbl[i] = string_sub(self, i, i)
  end
  return tbl
end
string.toTable = string_toTable

local string_explode = function(self, separator, withPattern)
  if string_len(separator) == 0 then return string_toTable(self) end
  local ret, currentPos = {}, 1
  for i = 1, string_len(self) do
    local startPos, endPos = string_find(self, separator, currentPos, not withPattern)
    if not startPos then break end
    ret[i] = string_sub(self, currentPos, startPos - 1)
    currentPos = endPos + 1
  end
  ret[#ret + 1] = string_sub(self, currentPos)
  return ret
end
string.explode = string_explode

string.replaceAll = function(self, searchValue, replaceValue)
  local tbl = string_explode(self, searchValue)
  return next(tbl) and table_concat(tbl, replaceValue) or self
[...]

PR #1129, opened 2021-09-08
Add `string.replaceAll` lualib and optimize/fix `string.replace`

GlassBricks:
Adds `string.replaceAll` to lualib #1121

Also optimizes `string.replace`.

This also fixes so that `string.replace(All)` with a function replacer gets called with the correct arguments.

### DIFF
--- a/src/lualib/StringReplaceAll.ts
+++ b/src/lualib/StringReplaceAll.ts
@@ -1,0 +1,40 @@
+function __TS__StringReplaceAll(
+    this: void,
+    source: string,
+    searchValue: string,
+    replaceValue: string | ((match: string, offset: number, string: string) => string)
+): string {
+    let replacer: (match: string, offset: number, string: string) => string;
+    if (typeof replaceValue === "string") {
+        replacer = () => replaceValue;
+    } else {
+        replacer = replaceValue;
+    }
+    const parts: string[] = [];
+    let partsIndex = 1;
+
+    const sub = string.sub;
+    if (searchValue.length === 0) {
+        parts[0] = replacer("", 0, source);
+        partsIndex = 2;
+        for (const i of $range(1, source.length)) {
+            parts[partsIndex - 1] = sub(source, i, i);
+            parts[partsIndex] = replacer("", i, source);
+            partsIndex += 2;
+        }
+    } else {
+        const find = string.find;
+        let currentPos = 1;
+        while (true) {
+            const [startPos, endPos] = find(source, searchValue, currentPos, true);
+            if (!startPos) break;
+            parts[partsIndex - 1] = sub(source, currentPos, startPos - 1);
+            parts[partsIndex] = replacer(searchValue, startPos - 1, source);
+            partsIndex += 2;
+
+            currentPos = endPos + 1;
+        }
+        parts[partsIndex - 1] = sub(source, currentPos);
+    }
+    return table.concat(parts);
+}
--- a/test/unit/builtins/string.spec.ts
+++ b/test/unit/builtins/string.spec.ts
@@ -54,8 +54,9 @@ test("string index (side effect)", () => {
     `.expectToMatchJsResult();
 });
 
-test.each([
+const replaceTestCases = [
     { inp: "hello test", searchValue: "", replaceValue: "" },
+    { inp: "hello test", searchValue: "", replaceValue: "_" },
     { inp: "hello test", searchValue: " ", replaceValue: "" },
     { inp: "hello test", searchValue: "hello", replaceValue: "" },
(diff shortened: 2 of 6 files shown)


test/unit/builtins/string.spec.ts
```diff
@@ -54,24 +55,46 @@ test("string index (side effect)", () => {
...
+        { inp: "hello test", searchValue: "test", replaceValue: "world" },
+        { inp: "hello test", searchValue: "test", replaceValue: "%world" },
+        { inp: "hello test", searchValue: "test", replaceValue: "." },
+        { inp: "hello %test", searchValue: "test", replaceValue: "world" },
+        { inp: "hello %test", searchValue: "%test", replaceValue: "world" },
+        { inp: "hello test.", searchValue: ".", replaceValue: "$" },
+        { inp: "aaa", searchValue: "a", replaceValue: "b" },
+    ];
+    describe("string replacer", () => {
+        test.each(testCases)("%p", ({ inp, searchValue, replaceValue }) => {
```
Perryvw, line 74:
```suggestion
    test.each(testCases)("string replacer %p", ({ inp, searchValue, replaceValue }) => {
```
You shouldn't nest describes, and in this case I don't see why it is needed in the first place as there is only one test inside.

test/unit/builtins/string.spec.ts
```diff
@@ -54,24 +55,46 @@ test("string index (side effect)", () => {
...
+        { inp: "hello test", searchValue: "hello", replaceValue: "" },
+        { inp: "hello test", searchValue: "test", replaceValue: "" },
+        { inp: "hello test", searchValue: "test", replaceValue: "world" },
+        { inp: "hello test", searchValue: "test", replaceValue: "%world" },
+        { inp: "hello test", searchValue: "test", replaceValue: "." },
+        { inp: "hello %test", searchValue: "test", replaceValue: "world" },
+        { inp: "hello %test", searchValue: "%test", replaceValue: "world" },
+        { inp: "hello test.", searchValue: ".", replaceValue: "$" },
+        { inp: "aaa", searchValue: "a", replaceValue: "b" },
+    ];
```
Perryvw, line 72: please add whitelines before and after tests like we have done in other .spec.ts files

test/unit/builtins/string.spec.ts
```diff
@@ -1,5 +1,6 @@
 import { LuaLibImportKind } from "../../../src";
 import * as util from "../../util";
+import { describe } from "jest-circus";
```
Perryvw, line 3: This should not be required, I don't think we import describe in any other file.

src/lualib/StringReplaceAll.ts
```diff
@@ -0,0 +1,40 @@
...
+    let replacer: (match: string, offset: number, string: string) => string;
+    if (typeof replaceValue === "string") {
+        replacer = () => replaceValue;
+    } else {
+        replacer = replaceValue;
+    }
+    const parts: string[] = [];
+    let partsIndex = 1;
+
+    const sub = string.sub;
```
Perryvw, line 16: Why are you aliasing this function?

Cheatoid: It is for localization (optimization technique), but this is *doing it wrong*... The localization should be done at top-level scope, not within the function. This is one of the reasons why I wanted locals at top-level, btw.

Cheatoid: As seen in [my snippet](https://github.com/TypeScriptToLua/TypeScriptToLua/issues/1121#issuecomment-913183505), localization is done with a workaround, that is by wrapping "top-level" code inside `{}`.
If you are not familiar with this technique, here're some links I collected (since I didn't make a proposal yet):

For more details, click [here](https://stackoverflow.com/q/32945039), [here](http://lua-users.org/wiki/OptimisingUsingLocalVariables), and [here for benchmarks](https://gitspartv.github.io/LuaJIT-Benchmarks/#test1) (particularly relevant tests are: 1st & 2nd). This is optimization is very common and it is universal.

Perryvw: fair enough, seems like a valid optimization

test/unit/builtins/string.spec.ts
GlassBricks: I'm going to blame auto-import for this

src/lualib/StringReplaceAll.ts
GlassBricks: I wanted to do the localization at the top level as well, but since lualib is all global this was hard to do. I figured putting it at the start of the function was the next best thing. There might be a way to do it using a code block like Cheatoid mentioned but it makes everything a bit ugly.

If we ever make lualib not all global, then we can move this to top level

PR #1136, opened 2021-09-12
validating assignment of functions to object literal method types

tomblind:
Adds correct validation for cases where a function is being indirectly assigned to a type defined as an object literal method.

Example:
```ts
function out(this: void, val: string) {
  print(val);
}

function bug(input: { consumer: (val: string) => void }) {
    input.consumer("test");
}

bug({ consumer: out }); // Should fail function assignment validation
```
[Playground](https://typescripttolua.github.io/play/#code/GYVwdgxgLglg9mABHEUAUUAWMDOAuRANzhgBMAaIgQwBsCcoAnGMAcwEpEBvAKEUQAOzMOkK12Abh4BfHj1CRYCRACMQrNCwGoCXRBAQ4QAWwCmjAmjF1EDYR0QBeAHxESpRNM69+-LagA6AzAjM0Y0ACIoUwYIyRk5NQ09YNDzAhQoT0lEAHpcxABlTBQaD2AqGBpEBWh4JCocHBhWMDMRahoyKiUwHiA)

### DIFF
--- a/test/unit/functions/validation/invalidFunctionAssignments.spec.ts
+++ b/test/unit/functions/validation/invalidFunctionAssignments.spec.ts
@@ -79,10 +79,10 @@ test.each(invalidTestFunctionCasts)(
     "Invalid object with method assignment with cast (%p, %p, %p)",
     (testFunction, castedFunction, isSelfConversion) => {
         util.testModule`
-        ${testFunction.definition ?? ""}
-        let obj: { fn: typeof ${testFunction.value} };
-        obj = {fn: ${castedFunction}};
-    `.expectDiagnosticsToMatchSnapshot(
+            ${testFunction.definition ?? ""}
+            let obj: { fn: typeof ${testFunction.value} };
+            obj = {fn: ${castedFunction}};
+        `.expectDiagnosticsToMatchSnapshot(
             isSelfConversion
                 ? [unsupportedSelfFunctionConversion.code, unsupportedNoSelfFunctionConversion.code]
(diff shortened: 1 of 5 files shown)


test/unit/functions/validation/invalidFunctionAssignments.spec.ts
```diff
@@ -44,6 +75,22 @@ test.each(invalidTestFunctionCasts)("Invalid function assignment with cast (%p)"
     );
 });
 
+test.each(invalidTestFunctionCasts)(
+    "Invalid object with method assignment with cast (%p, %p, %p)",
+    (testFunction, castedFunction, isSelfConversion) => {
+        util.testModule`
+        ${testFunction.definition ?? ""}
```
Perryvw, line 82: this indenting seems weird, it's different from the other tests in the diff

test/unit/functions/validation/validFunctionAssignments.spec.ts
```diff
@@ -43,6 +54,19 @@ test.each(validTestFunctionCasts)("Valid function assignment with cast (%p)", (t
         .expectToMatchJsResult();
 });
 
+test.each(validTestMethodCasts)(
+    "Valid object with method assignment with cast (%p, %p)",
+    (testFunction, castedFunction) => {
+        util.testFunction`
+        let obj: { fn: typeof ${testFunction.value} };
```
Perryvw, line 61: this indenting is weird too

PR #1155, opened 2021-11-02
Void expressions using preceding statements

GlassBricks:
Uses preceding statements to implement void expressions.

It uses `transformSingleExpressionToStatement`, extracted from `transformExpressionStatement`.

Will undraft this PR when preceding-statements is merged.


GlassBricks:
This seems to be already implemented in the preceding statements PR, however this implementation uses the same logic as `transformExpressionStatement` to avoid creating an anonymous identifier where possible.
If that's not desirable/worth it I'll close this PR

tomblind:
Ah, sorry - missed that!

If you don't mind, would you update your PR to add that additional functionality?

GlassBricks:
This PR already has that functionality.
You could cherry pick the one commit of this PR into the preceding statements PR (https://github.com/TypeScriptToLua/TypeScriptToLua/pull/1155/commits/1d6ecae7df09b15efc8fb4032ead36ad21a552d7).

PR #1160, opened 2021-11-07
Fix delete operator

### LINKED ISSUE #1159: Incorrect delete expression behavior
According to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete, in strict mode, the delete operator _always_ returns true, and throws a _TypeError_ if trying to delete an own nonconfigurable property.

The current implementation sometimes returns `false` and doesn't throw a TypeError.

GlassBricks:
Fixes #1159

### DIFF
--- a/test/unit/builtins/object.spec.ts
+++ b/test/unit/builtins/object.spec.ts
@@ -222,11 +222,15 @@ describe("delete from object", () => {
     test("delete from object", () => {
         util.testFunction`
             const obj = { foo: "bar", bar: "baz" };
-            delete obj["foo"];
-            return obj;
-        `
-            .setTsHeader("declare function setmetatable<T extends object>(this: void, table: T, metatable: any): T;")
-            .expectToEqual({ bar: "baz" });
+            return [delete obj["foo"], obj];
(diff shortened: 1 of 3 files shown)


test/unit/builtins/object.spec.ts
```diff
@@ -222,22 +222,42 @@ describe("delete from object", () => {
...
+        const obj = {};
+        Object.defineProperty(obj,"foo", {
+            configurable: false,
+            value: true
+        })
+        try {
+            delete obj["foo"]
+            return "deleted"
+        } catch (e) {
+            return e instanceof TypeError
```
Perryvw, line 259: Looks like this branch is not actually hit in this test, could you add a test where the TypeError is thrown? Also indenting seems worng here, it should be increased by 1 indent on line 250 to 261, empty line 249 should be removed.

GlassBricks: The test _does_ throw TypeError (the return value is true), at least when I run on my computer. Did you recompile lualib, or maybe is strict mode off somehow?

I added strict mode explicitly to the test just in case.

PR #1168, opened 2021-11-19
Fix __TS__AsyncAwaiter error handling

dicarlo2:
Similar to line 64, this should use `errorOrErrorHandler` in order to capture the thrown error.

Happy to add tests if you can give me some pointers on how to do so. Here is an example where the current code breaks:
```
const success = async () => {};
const error = async () => {
  await success();
  throw new Error('Hello World');
}
error().catch((err) => {
  // prints true prior to change in this PR 
  print(err === undefined); 
});
```

### DIFF
--- a/src/lualib/Await.ts
+++ b/src/lualib/Await.ts
@@ -30,7 +30,7 @@ function __TS__AsyncAwaiter(this: void, generator: (this: void) => void) {
             if (success) {
                 step(resultOrError, errorOrErrorHandler);
             } else {
-                reject(resultOrError);
+                reject(errorOrErrorHandler);
             }
         }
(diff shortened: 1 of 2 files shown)


Perryvw:
Thanks, would indeed be nice if you can add a test here: https://github.com/TypeScriptToLua/TypeScriptToLua/blob/master/test/unit/builtins/async-await.spec.ts#L297

Since all of this should be able to run synchonously, you can make use of the promiseTestLib log the error which you return from the test, similar to how this test works: https://github.com/TypeScriptToLua/TypeScriptToLua/blob/master/test/unit/builtins/async-await.spec.ts#L50

To save some time you can use `npm test async-await` to only run the tests in that file

Issue #1162, opened 2021-11-10
Conditional chaining doesn't produce nil-checks


powertomato:
The following input:
```.ts
function func(this: void, test?: []) {
    test?.forEach(x => print(x));
}
```

Emits this lua code:
```.lua
function func(test)
    __TS__ArrayForEach(
        test,
        function(____, x) return print(x) end
    )
end
```

According to the semantics of the `?.` operator, I'd expect a `nil`-check surrounding the `__TS__ArrayForEach`-call:
```.lua
function func(test)
    if test ~= nil then
        __TS__ArrayForEach(
            test,
            function(____, x) return print(x) end
        )
    end
end
```

Perryvw:
Fixed with #1171

PR #1171, opened 2021-11-20
Optional chaining using preceding statements

GlassBricks:
This implements optional chaining using preceding statements. It also fixes a number of problem with optional chaining interacting with other language features, e.g. #1162. This also supports delete expressions.

This implementation is roughly based on Typescript's transformES2020, with changes to integrate with existing TSTL logic.
Some implementation choices were made with compatibility with existing TSTL features in mind.

### Implementation overview
Given a chain (a)?.b.c.d, where (a) is any expression (possibly another optional chain), the result is transformed roughly into
```lua
local temp = (a)
if temp ~= nil then
    temp = temp.b.c.d
end
return temp
```
This ensures proper short-circuiting behavior.
The same temp variable is used for both the left-hand side and result to reduce the number of temps.
The chain is built as _typescript_ nodes first before transforming, to maximize compatibility with existing transformations. `ts.setOriginalNode` is used to retain type information and source-map position.

### Capturing this-value

If the optional chain is a contextual call, e.g `a.b?.(c)`, then the `this` parameter may need to be captured to be passed in the call, e.g `temp=a.b; result = temp(a, c)`. The need to capture `this` is detected by marking the leftmost `temp` in the chain, and having `transformCallExpression` mark it when it results in a contextual call -- see `OptionalContinuation`.
The this-value is captured by functions that return `ExpressionWithThisValue`. (This is analogous to `SyntheticExpression` in typescript's implementation, except that SyntheticExpression is an internal AST node).
The this-value is captured if first expression on the left-hand side is an appropriate property-access or element-access expression.

### Builtins
This implementation disallows optional calls on built-ins, e.g. `array?.forEach()` is allowed, but `array.forEach?.()`, `console?.log()`, `Number?.()` is not allowed. It might be possible to detect these situations and treat them as non-optional calls. However, the detection logic will be complicated and these optional calls almost never make sense, so we report a diagnostic on them instead.

`Enum?.member` also reports a diagnostic if the enum is marked with @compileMembersOnly.


src/transformation/visitors/optional-chaining.ts
```diff
@@ -0,0 +1,187 @@
...
+
+function transformExpressionWithThisValueCapture(
+    context: TransformationContext,
+    node: ts.Expression
+): ExpressionWithThisValue {
+    if (ts.isParenthesizedExpression(node)) {
+        return transformExpressionWithThisValueCapture(context, node.expression);
+    }
+    if (ts.isPropertyAccessExpression(node)) {
+        return transformPropertyAccessExpressionWithCapture(context, node, true);
```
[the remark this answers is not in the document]
GlassBricks: The main issue is that optional chaining cannot be transformed "normally", as the result of the optional chain will in the end be inside an if statement. If `transformCallExpression` were to handle this logic, it needs to somehow be able to tell the nodes higher in the AST (the rest of the optional chain) to put the result in the if statement.
The way I handled it here is largely based on `transformES2020` in typescript, which includes passing flags around. However, transformES2020 assumes all call expressions can take thisArg, which is not the case in TSTL, hence the state.

In hindsight, it might be possible to abandon `transformES2020`'s model and instead handle most of the logic within `transformCallExpression`, if some state were added in `TransformationContext` instead. This might be a better solution; I might try implementing this within the next few days and open another pull request.

GlassBricks:
See https://github.com/TypeScriptToLua/TypeScriptToLua/pull/1171#discussion_r755356138, I'll likely be submitting an alternative implementation soon

GlassBricks:
Update: I'm trying an implementation with details described above. It involves less weird state than the current implementation, however runs into multiple other problems, so I think the current approach is better.

Problems with the alternative approach:

There will be repeated logic in all of `transformCallExpression`, `transformPropertyExpression`, `transformElementAcessExpression`.

It will require modifying every builtin/"custom" transformations; they will also need to take an arbitrary `lua.Expression` for its left-hand-side, in case it is an optional chain continuation.
The existing implementation works around these problems by creating a typescript nodes that are given back to the existing transformation.

It gets complicated to handle separate preceding statements for each segment an optional chain correctly.

Additionally, any plugin that does not also take into account optional chaining explicitly will be broken.
The existing implementation has somewhat better plugin interoperability; one can simply delegate to the existing transform when an optional chain is received, and the plugin will eventually be called again with a non-optional chain.

src/transformation/visitors/optional-chaining.ts
```diff
@@ -0,0 +1,187 @@
...
+// the assignment is "manually hoisted" in transformOptionalChain so that the variable can be accessed in a later scope.
+export function captureThisValue(
+    context: TransformationContext,
+    expression: lua.Expression,
+    tsOriginal: ts.Node
+): { thisValue: lua.Expression; isNewTemp?: boolean } {
+    if (!shouldMoveToTemp(context, expression, tsOriginal) && !isOptionalContinuation(tsOriginal)) {
+        return { thisValue: expression };
+    }
+    const tempIdentifier = context.createTempNameForLuaExpression(expression);
```
Perryvw, line 53: I REALLY dislike the fact that tempIdentifier is generated from the expression here and just assigned. This means that this code has a bunch of invisible dependencies on other parts of our code that will be extremely  hard to maintain.

I would propose an alternative: Rename this function to `assignToPrecedingTemp` and pass in the identifier that should be assigned to. You could use this identifier (or undefined) instead of these 'shouldCaptureValue' boolean parameters (call it something like `assignToPrecedingIdentifier?: ts.Identifier`). I think that would make the flow of the code overall much clearer.

Perryvw: This would also eliminate the need for these ugly `transformElementAccessExpression() => transformElementAccessExpressionWithCapture` one liners

src/transformation/builtins/index.ts
```diff
@@ -68,55 +70,75 @@ export function transformBuiltinCallExpression(
...
+                if (isOptionalCall || isOptionalAccess) return unsupportedOptionalCall();
                 return transformNumberConstructorCall(context, node);
             case "PromiseConstructor":
+                if (isOptionalCall || isOptionalAccess) return unsupportedOptionalCall();
                 return transformPromiseConstructorCall(context, node);
         }
     }
 
     if (isStringType(context, ownerType) && hasStandardLibrarySignature(context, node)) {
+        if (isOptionalCall) return unsupportedOptionalCall();
```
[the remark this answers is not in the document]
GlassBricks: isOptionalCall is when the call itself is optional,`value.builtin?.()`.
isOptionalAccess is when the _left hand side_ of the call is itself an optional access. e.g. `console?.log`.

checking for isOptionalAccess is done when the builtlin also transforms the property access, e.g. `console?.log()` is not allowed, but `array?.forEach(...)` is. 

If it helps we could comment/rename so this is clearer.

src/transformation/visitors/delete.ts
```diff
@@ -5,21 +5,27 @@ import { transformLuaLibFunction, LuaLibFeature } from "../utils/lualib";
 import { unsupportedProperty } from "../utils/diagnostics";
 import { isArrayType, isNumberType } from "../utils/typescript";
 import { addToNumericExpression } from "../utils/lua-ast";
+import { transformOptionalDeleteExpression } from "./optional-chaining";
 
 export const transformDeleteExpression: FunctionVisitor<ts.DeleteExpression> = (node, context) => {
+    const innerExpression = ts.skipParentheses(node.expression);
```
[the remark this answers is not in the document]
GlassBricks: Oh, didn't know that.

Depending on how this works, though, this could result in incorrect behavior as parenthesis change the behavior of optional chaining, (a?.b)() is valid (if strict null checks is off). I'll investigate/test this.

src/transformation/builtins/object.ts
```diff
@@ -36,9 +37,9 @@ export function transformObjectConstructorCall(
 
 export function transformObjectPrototypeCall(
     context: TransformationContext,
-    node: PropertyCallExpression
+    node: ts.CallExpression,
+    expression: ts.PropertyAccessExpression
```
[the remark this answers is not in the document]
GlassBricks: See `transformBuiltinCallExpression`, where this called; the node is no longer always a `PropertyCallExpression`, as the left hand side may be an optional temp variable. So, the _original_ property access expression `ts.getOriginalNode(node.expression)` is passed in. This is only for checking if a builtin call is was used; if the function returns anything and if the builtin is an optional call, a diagnostic is emitted.

src/transformation/visitors/spread.ts
```diff
@@ -17,7 +17,7 @@ import { annotationRemoved } from "../utils/diagnostics";
 import { isGlobalVarargConstant } from "./language-extensions/vararg";
 
 export function isOptimizedVarArgSpread(context: TransformationContext, symbol: ts.Symbol, identifier: ts.Identifier) {
-    if (!ts.isSpreadElement(identifier.parent)) {
+    if (identifier.parent && !ts.isSpreadElement(identifier.parent)) {
```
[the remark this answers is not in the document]
GlassBricks: Actually this change was made before `identifier.ts` was changed, and no longer necessary (nodes created with `ts.factory.createXX` have undefined as parent, and was causing errors). Will revert.

src/transformation/visitors/optional-chaining.ts
GlassBricks: I considered this, but the problem is that there might not be a thisArg to capture, and the temp variable is not always needed.

However, I like your proposal that the temp identifier is passed in; and the temp identifier generated only if necessary. However, info on if the temp was used still needs to be returned, so the variable declaration is only generated/used if needed. I'll implement this in the next commit.

As for the one-liners/extra functions, typescript's own implementation uses an internal AST node called `SyntheticExpression` (`ExpressionWithThisValue` is analgous to this), and has a bunch of assertions that this node doesn't end up outside of `transformES2020`. I opted to go for an explicitly different interface instead, as it would be difficult to ensure an internal node never makes into the result AST, across multiple functions and different files.

Let me know if you would prefer something like typescript's approach (introducing an internal node) or other possible alternatives.

src/transformation/visitors/access.ts
```diff
@@ -89,49 +115,54 @@ export const transformPropertyAccessExpression: FunctionVisitor<ts.PropertyAcces
...
+            return { expression: lua.createIdentifier(property, node) };
         }
     }
 
-    if (ts.isOptionalChain(node)) {
-        // Only handle full optional chains separately, not partial ones
-        return transformOptionalChain(context, node);
+    const builtinResult = transformBuiltinPropertyAccessExpression(context, node);
+    if (builtinResult) {
+        // Ignore shouldCaptureThisValue.
```
Perryvw, line 149: This comment is out of sync now after updating the parameter name.

GlassBricks:
There is a problem with non-null assertion chains. Typescript's implementation already assumes they have been pre-transformed by a previous transformer

tomblind:
A very minor issue I noticed is that calls with args that generate preceding statements cause an erroneous temp for `_G` to be generated.
```ts
const a = { f: (arg: any) => {} };
let i = 0;
return a.f?.(i++);
```
```lua
local ____a_f_result_0 = a.f
if ____a_f_result_0 ~= nil then
    local ____a_f_result_0_3 = ____a_f_result_0
    local ____G_2 = _G -- ???
    local ____i_1 = i
    i = ____i_1 + 1
    ____a_f_result_0 = ____a_f_result_0_3(a, ____i_1)
end
return ____a_f_result_0
```

GlassBricks:
That can probably be fixed, actually, expanding on Perry's original suggestion for handling captureThisValue.

GlassBricks:
Actually, unless we want to _always_ create a temp variable for context, fixing this could be  complicated.

Instead, would it be preferable to occasionally have an unnecessary temp in the above case (keep as is), or always generate a new temp variable for context (which transformCallExpression always uses, even if `nil` or `_G`)?

src/transformation/visitors/expression-list.ts
```diff
@@ -5,12 +5,13 @@ import { TransformationContext, tempSymbolId } from "../context";
...
 import { transformInPrecedingStatementScope } from "../utils/preceding-statements";
 import { isConstIdentifier } from "../utils/typescript";
+import { isOptionalContinuation } from "./optional-chaining";
 
 export function shouldMoveToTemp(context: TransformationContext, expression: lua.Expression, tsOriginal?: ts.Node) {
     return (
         !lua.isLiteral(expression) &&
         !(lua.isIdentifier(expression) && expression.symbolId === tempSymbolId) && // Treat generated temps as consts
-        !(tsOriginal && isConstIdentifier(context, tsOriginal))
+        !(tsOriginal && (isConstIdentifier(context, tsOriginal) || isOptionalContinuation(tsOriginal)))
```
tomblind, line 14: What is the purpose of this?

GlassBricks: If an identifier is an optional continuation (ts temp variable), it doesn't need not be moved to temp. Noticed this when debugging the newly added tests.

tomblind: ah, ok

src/transformation/visitors/call.ts
```diff
@@ -322,7 +322,9 @@ export const transformCallExpression: FunctionVisitor<ts.CallExpression> = (node
     if (!isContextualCall) {
         [callPath, parameters] = transformCallAndArguments(context, node.expression, node.arguments, signature);
     } else {
-        const callContext = context.isStrict ? ts.factory.createNull() : ts.factory.createIdentifier("_G");
+        // if is optionalContinuation, context will be handled by transformOptionalChain.
+        const callContext =
+            optionalContinuation || context.isStrict ? ts.factory.createNull() : ts.factory.createIdentifier("_G");
```
Perryvw, line 327: Shouldn't this be a `??`

tomblind: It's a boolean check, so if anything, the proper way to do it would be `optionalContinuation !== undefined || context.isStrict ? ...`

GlassBricks: Its equivalent to || in this case, and its only used as a condition of ternary. Should we still change it?

tomblind (approved):

PR #1180, opened 2021-11-29
Implemented support for properties on functions

### LINKED ISSUE #388: Assigning properties on functions
Currently, this is allowed:
```ts
function foo() {}
foo.bar = "bar";
```
But it produces bad code since functions are not tables in lua.

A variant of this is namespace merging into a function:
```ts
function foo() {}
namespace foo {
    export let bar = "bar";
}
```

I'm not sure if there's a reasonable way to support this. It's probably just something that should be an error.

### LINKED ISSUE #1174: Function call call breaks when using a Function type branded using type alias.
See this example:
https://typescripttolua.github.io/play/#code/FAFwngDgpgBAQgJwIYDsAmU0B4AqA+GAXhhxgDIYBvAfQCNl0AuGVMAX2GAGMB7FAZxAwAtmABiAVxRciMABRyAlITyU2ilv3gMM2JSoBuPAJZo8wUZOkA6LkgA29uQAYNAejcwoAD2hcQmDA8EiAQIcyWUlwuipzAHjAgABZQ-LAA7jwIANYANDBJPOlQBlAIoJCwiKi6AIy4BMSknLwCQpHStbIKyqrqmto1mPX6eEam5h1ctbYOTq5xrYIi4lEATN2jahpIWj2GJmjkVHQ6zKzqFqvSa7OOMUA

I would expect `myFunc(0)` in the output, not `myFunc.call(0)`

This doesn't happen if the brand type is removed, or "inlining" the type alias

### LINKED ISSUE #1179: Async/Await libs don't have Promise libs as dependency
[Playground](https://typescripttolua.github.io/play#code/IYZwngdgxgBAZgV2gFwJYHsI2QUxMgCgEoYBvAXwCgA3dVAE2z0KIG5Kg)

tomblind:
Closes #388

Functions with properties are now supported (including via declaration merging with namespaces). This is done by wrapping the function definition in a "callable" table (using the `__call` metamethod).

Example:
```ts
function foo(s: string) { return s; }
foo.bar = "bar";
```
```lua
local foo = setmetatable(
    {},
    {__call = function(____, self, s)
        return s
    end}
)
foo.bar = "bar"
```

Additionally, this fixes #1174 and fixes #1179 as I needed those working for tests.

### DIFF
--- a/src/transformation/builtins/function.ts
+++ b/src/transformation/builtins/function.ts
@@ -32,7 +32,7 @@ export function transformFunctionPrototypeCall(
             return transformLuaLibFunction(context, LuaLibFeature.FunctionBind, node, caller, ...params);
         case "call":
             return lua.createCallExpression(caller, params, node);
-        default:
+        case "toString":
             context.diagnostics.push(unsupportedProperty(expression.name, "function", expressionName));
     }
 }
@@ -60,7 +60,10 @@ export function transformFunctionProperty(
                 ? lua.createBinaryExpression(nparams, lua.createNumericLiteral(1), lua.SyntaxKind.SubtractionOperator)
                 : nparams;
 
-        default:
+        case "arguments":
+        case "caller":
+        case "displayName":
+        case "name":
             context.diagnostics.push(unsupportedProperty(node.name, "function", node.name.text));
     }
 }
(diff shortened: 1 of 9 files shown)


src/transformation/builtins/function.ts
```diff
@@ -32,7 +32,7 @@ export function transformFunctionPrototypeCall(
             return transformLuaLibFunction(context, LuaLibFeature.FunctionBind, node, caller, ...params);
         case "call":
             return lua.createCallExpression(caller, params, node);
-        default:
+        case "toString":
```
Perryvw, line 35: This seems like a mistake, this should be default to catch everything we have not handled

tomblind: Since functions can have any properties, we don't want to throw errors on them being used. So I changed this to only error on JS lib functions that we don't handle.

PR #1182, opened 2021-12-01
Binding fixes

### LINKED ISSUE #1177: Local variable for parameter destructuring appears in wrong location, if variable is used as spread parameter
See the following:
https://typescripttolua.github.io/play#code/JYOwLgpgTgZghgYwgAgGIHt3IN4ChkHJwBcRIAngNoC6uAvrowPRPIA26CcbyAbnFGBwARmxRwiAB0kQBAZ2ShkYABYoA7lHQgA5skltEEXAm1ywyALZxJyALzIAFNjh1SGdAEp7APhz5CKAgwAFcoECdPOx9TEDl0MQA6Dh1HRPS4bxZkCAAPOEsDFHQYZE1tPWEIFTheYHQoABpFUq5JUKDFCIQOOTDjBhMzCxgQkAQARnsrG2cSZEoJ6jpPIbiRsYQAJmnrSTnSSi3l1dHxicdTza3LoA

In this specific scenario that I actually encountered in one of my projects:
- A parameter in a function is destructured
- The destructured variable is used as a spread parameter

Then the variable declaration for the destructured parameter is declared in the wrong scope.

This occurs since v1.2.0, not in v1.1.0.

tomblind:
This fixes a couple of issues with binding patterns introduced by preceding statements.

Fixes #1177 

Also fixes this case: [Playground](https://typescripttolua.github.io/play/#code/MYewdgzgLgBARgGwIYwLwwNoYIwF0A0GATARgMy64DcAUKJLMEggnEsANYRqbU00AzEACcYACnrRMYXDBAD4yAJQwA3jRiaYTFm04QAdAAcArhAAWYsStQA+GEeEBLMFDFglS2gF9+O1uxcBkLCAKLslsBwaPZR1l5AA)
In this instance, aside from the issue with the binding variable not being generated inside the loop's scope, an additional bug was found that has apparently always existed where the `local` would be stripped off from the loop variable because it was falsely seen as being global in scope (because it was transformed before entering the loops body).

### DIFF
--- a/src/transformation/utils/scope.ts
+++ b/src/transformation/utils/scope.ts
@@ -14,6 +14,7 @@ export enum ScopeType {
     Block = 1 << 5,
     Try = 1 << 6,
     Catch = 1 << 7,
+    LoopInitializer = 1 << 8,
 }
 
 interface FunctionDefinitionInfo {
--- a/src/transformation/visitors/function.ts
+++ b/src/transformation/visitors/function.ts
@@ -85,7 +85,11 @@ export function transformFunctionBodyHeader(
             }
 
             // Binding pattern
-            bindingPatternDeclarations.push(...transformBindingPattern(context, declaration.name, identifier));
+            const name = declaration.name;
+            const [precedingStatements, bindings] = transformInPrecedingStatementScope(context, () =>
+                transformBindingPattern(context, name, identifier)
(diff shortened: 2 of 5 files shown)


Zamiell:
thank you tom!

PR #1191, opened 2021-12-19
Fix nullish coalescing and conditionals when using generics

### LINKED ISSUE #1184: Wrong not-falsy assumption when using generics
See here:

https://typescripttolua.github.io/play#code/FDBmFcDsGMBcEsD2kAEAHe0DWAeAKgHwAUAhgFwoBGiiANgKYmQA0VFer07AlOwN7AUQgE71Y4YahIoA-FRQVowAL4hoyAM516AOlqIA5kQzYisYeHqsoAE3qh4kejdYAiSIjT1X3bsCA

When a generic type, ternary expressions and null coalescing (and possibly others) are generated incorrectly (assuming that the left argument cannot be falsy).

GlassBricks:
Fixes #1184

### DIFF
--- a/test/unit/conditionals.spec.ts
+++ b/test/unit/conditionals.spec.ts
@@ -99,3 +99,16 @@ test.each(["true", "false", "a < 4", "a == 8"])("Ternary Conditional Delayed (%p
         return delay();
     `.expectToMatchJsResult();
 });
+
+test.each([false, true, null])("Ternary conditional with generic whenTrue branch (%p)", trueVal => {
+    util.testFunction`
+        function ternary<B, C>(a: boolean, b: B, c: C) {
+            return a ? b : c
+        }
+        return ternary(true, ${trueVal}, "wasFalse")
+    `
+        .setOptions({
+            strict: true,
+        })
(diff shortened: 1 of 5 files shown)


test/unit/conditionals.spec.ts
```diff
@@ -99,3 +99,16 @@ test.each(["true", "false", "a < 4", "a == 8"])("Ternary Conditional Delayed (%p
...
+
+test.each([false, true, null])("Ternary conditional with generic whenTrue branch (%p)", trueVal => {
+    util.testFunction`
+        function ternary<B, C>(a: boolean, b: B, c: C) {
+            return a ? b : c
+        }
+        return ternary(true, ${trueVal}, "wasFalse")
+    `
+        .setOptions({
+            strict: true,
```
Perryvw, line 111: What is the reason for adding this strict here?

GlassBricks: if its not strict, in the old logic, due to any object possibly being null, it will always use an if statement.
I could change this to `strictNullChecks` to be more clear

PR #1193, opened 2021-12-19
Fix optimized vararg spread with casts

### LINKED ISSUE #1143: no table.unpack optimization when using cast to any[]
There probably shouldn't be a table.unpack when you cast parameters to any[] (or another array type). For instance:

https://typescripttolua.github.io/play/#code/FDBmFcDsGMBcEsD2kAEoCGMCeBhdAbfAHgDEUBTAD1nMgBMBnFACgDp30AnAcwYC4UmLAEoAvAD4h45hBgCSAGhTtWXXgIAKXdAFtyNTg1LjhAgEr7wnSABUsAB3LGA3sBTuUnS9bRRobdmY1JnQQyCwAbQBdYWFgAF8gA

Current workaround: make the function definition an overload and the implementation take `any[]` type as parameters

GlassBricks:
Fixes #1143.

Also fixes for other outer expressions.

### DIFF
--- a/src/transformation/visitors/spread.ts
+++ b/src/transformation/visitors/spread.ts
@@ -16,8 +16,15 @@ import { isMultiReturnCall } from "./language-extensions/multi";
 import { annotationRemoved } from "../utils/diagnostics";
 import { isGlobalVarargConstant } from "./language-extensions/vararg";
 
+function skipOuterExpressionParents(node: ts.Node) {
+    while (ts.isOuterExpression(node)) {
+        node = node.parent;
+    }
+    return node;
+}
+
 export function isOptimizedVarArgSpread(context: TransformationContext, symbol: ts.Symbol, identifier: ts.Identifier) {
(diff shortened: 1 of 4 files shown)


src/transformation/visitors/errors.ts
```diff
@@ -44,31 +44,38 @@ export const transformTryStatement: FunctionVisitor<ts.TryStatement> = (statemen
         const catchParameter = statement.catchClause.variableDeclaration
             ? transformIdentifier(context, statement.catchClause.variableDeclaration.name as ts.Identifier)
             : undefined;
-        const catchParameters = () => (catchParameter ? [lua.cloneIdentifier(catchParameter)] : []);
```
Perryvw, line 47: What happened here, this seems like the change from the other branch? Guess you need to merge master into this branch.

GlassBricks: Yeah, somehow those changes made it here.
I just force pushed a new commit without those changes

PR #1201, opened 2022-01-01
Improve expression statement handling and language extensions calls

GlassBricks:
These changes go together:

With preceding statements, table function language extensions (`TableGet`, `TableSet`, etc.) can now be used as expressions.
`TableDelete` functions are typed to return `boolean`, but previously the return value could never be used; these now return `true` to be similar to delete expressions. Other methods return `nil`.
Additionally, I fixed table function extensions that use preceding statements or spread arguments. This required some refactoring; The result is based on `transformBuiltinCallExpression`. This additionally avoids unnecessary calls to `getTableExtensionKindForCall` and fixes some other minor bugs that I found.

transformExpressionStatement now omits the statement entirely if the inner expression result is "synthetic" with no side effects (a temp variable, or a literal without a source map position). This removes unnecessary `local ____ = ...` in several cases (e.g. `a?.b()`). This also decouples transformExpressionStatement from other language features: now table call language extensions can be transformed as an expression, and its return value (`nil` or `true`) is omitted when in an expression statement (not used).
- `void` expressions now also share some of this logic to omit unecessary `local ____ = ...`.

### DIFF
--- a/src/transformation/visitors/expression-statement.ts
+++ b/src/transformation/visitors/expression-statement.ts
@@ -21,19 +21,21 @@ export const transformExpressionStatement: FunctionVisitor<ts.ExpressionStatemen
 export function transformExpressionToStatement(
     context: TransformationContext,
     expression: ts.Expression
-): lua.Statement[] | lua.Statement | undefined {
+): lua.Statement | undefined {
     const result = context.transformExpression(expression);
 
-    // omit temp identifiers, and non-side effect expressions without source map position
-    if (
-        (lua.isIdentifier(result) && result.symbolId === tempSymbolId) ||
-        ((lua.isIdentifier(result) || lua.isLiteral(result)) && result.line === undefined)
-    ) {
+    const isTempVariable = lua.isIdentifier(result) && result.symbolId === tempSymbolId;
+    if (isTempVariable) {
         return undefined;
     }
-
-    return lua.isCallExpression(result) || lua.isMethodCallExpression(result)
-        ? lua.createExpressionStatement(result)
-        : // Assign expression statements to dummy to make sure they're legal Lua
-          lua.createVariableDeclarationStatement(lua.createAnonymousIdentifier(), result);
+    // "synthetic": no side effects and no original source
+    const isSyntheticExpression = (lua.isIdentifier(result) || lua.isLiteral(result)) && result.line === undefined;
+    if (isSyntheticExpression) {
+        return undefined;
+    }
+    if (lua.isCallExpression(result) || lua.isMethodCallExpression(result)) {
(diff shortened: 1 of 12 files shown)


test/unit/language-extensions/table.spec.ts
```diff
@@ -212,18 +212,14 @@ describe("LuaTableDelete extension", () => {
             .expectToEqual({ table: { bar: 12 } });
     });
 
-    test.each([
-        'const foo = tableDelete({}, "foo");',
-        'const foo = `${tableDelete({}, "foo")}`;',
-        'declare function foo(arg: any): void; foo(tableDelete({}, "foo"));',
-        'const foo = [tableDelete({}, "foo")];',
-    ])("LuaTableDelete invalid use as expression (%p)", expression => {
+    test("LUaTableDelete use as expression", () => {
```
lolleko, line 215:
```suggestion
    test("LuaTableDelete use as expression", () => {
```

src/transformation/visitors/expression-statement.ts
```diff
@@ -36,9 +15,25 @@ export const transformExpressionStatement: FunctionVisitor<ts.ExpressionStatemen
...
+
+export function transformExpressionToStatement(
+    context: TransformationContext,
+    expression: ts.Expression
+): lua.Statement[] | lua.Statement | undefined {
     const result = context.transformExpression(expression);
+
+    // omit temp identifiers, and non-side effect expressions without source map position
+    if (
+        (lua.isIdentifier(result) && result.symbolId === tempSymbolId) ||
```
lolleko, line 29: could you extract the condition into variable(s) with descriptive names?

lolleko (changes requested):

test/unit/language-extensions/table.spec.ts
```diff
@@ -212,18 +212,14 @@ describe("LuaTableDelete extension", () => {
...
-        'const foo = tableDelete({}, "foo");',
-        'const foo = `${tableDelete({}, "foo")}`;',
-        'declare function foo(arg: any): void; foo(tableDelete({}, "foo"));',
-        'const foo = [tableDelete({}, "foo")];',
-    ])("LuaTableDelete invalid use as expression (%p)", expression => {
+    test("LuaTableDelete use as expression", () => {
         util.testModule`
             declare const tableDelete: LuaTableDelete<{}, string>;
-            ${expression}
+            export const result = tableDelete({}, "foo")
```
lolleko, line 218: An additional test with a more complex expression would be nice (maybe just re-add the tests in the `.each block`).
Also If I understood correctly all TableExtensions can now be used as expression, so testing these as well would also be good.

lolleko (approved):

Issue #1197, opened 2021-12-23
Casting/assigning a normal function to a function type with-properties results in different references


GlassBricks:
See example: https://typescripttolua.github.io/play#code/JYOwLgpgTgZghgYwgAgGIFcQICoE8AOKA3gFDLnIAUAlAFzIBuA9sACZkX5RP4D89cELhIBfEiQRMQAZzDIYyALxVqigHxExkmXIBGcKAEZ6GLHkJL5EqbOT6oAJhOYcBFMpjiS9wwDouPJYAzNYyTAA2EL7hTADmlD7+3PjUodIRUTHx9g5JPNRAA

GlassBricks:
This may not have an elegant solution; If so, it might be best to add this to the caveats in docs, and/or detect and provide a warning in such cases.

tomblind:
Probably the best we can do is disallow assignments of functions without properties to types that have them. We already have assignment validations for all of the self vs noself stuff which covers all the various types of assignments (passing as arguments, etc...), so we can likely add this check there.

GlassBricks:
Actually, here's a potential solution: instead of creating a table with a `__call` metamethod, give the function a metatable using `debug.setmetatble` with `__index` and `__newindex` set to the same table. Also add a check in case the function already has a metatable; this can be done with lualib.

PR #1202, opened 2022-01-02
Fix function with properties assigned to variable declaration

### LINKED ISSUE #1196: Incorrect behavior when a function type with property is assigned to a variable
If a variable declaration is made where the right-hand side is a function type with properties, it wraps the result inside a metatable instead of just assigning the value.

Easier to explain with example:
https://typescripttolua.github.io/play#code/JYOwLgpgTgZghgYwgAgGIFcQICoE8AOKA3gFDLnIAUAlAFzIBuA9sACZkX5RP71wi4SAXxIkYmBGGBMQyAOYQwGLHkI16ynAWIdyCGQGcmAGwgA6Y0zmUARArA3qu5PpAGwyGMgC8Vat4A+UgoQl0MTc0trGwQ4Y1NWR2cREJgzLh4fZABmZyhFdChZGGFRV3dGOPQUX3tNVQgaEgB6ZuQuUDADZDtFG2QwAHdgJBIGKsanceNqptbkYG6QYGMSDvBKaer07nwnFra4o2RB7hA5EnKPLYgAJiybkiA

GlassBricks:
Fixes #1196: Only function expressions/arrow functions are wrapped in variable declarations.
Still leaves #1197 open.

Also fixed a bug I found, where a named, _not_ self-recursive function is assigned to a variable wasn't wrapped.

### DIFF
--- a/src/transformation/visitors/function.ts
+++ b/src/transformation/visitors/function.ts
@@ -277,8 +277,9 @@ export function transformFunctionLikeDeclaration(
 
     const [functionExpression, functionScope] = transformFunctionToExpression(context, node);
 
+    const isNamedFunctionExpression = ts.isFunctionExpression(node) && node.name;
     // Handle named function expressions which reference themselves
-    if (ts.isFunctionExpression(node) && node.name && functionScope.referencedSymbols) {
+    if (isNamedFunctionExpression && functionScope.referencedSymbols) {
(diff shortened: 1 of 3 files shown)


GlassBricks:
Also, this broke my existing code; could we get this in a patch release?

PR #1213, opened 2022-02-03
Lualib as module

GlassBricks:
This implements lualib as modules (lualib source files _and_ lualib_bundle are now modules), removes use of global variables.

 - Creates a new "internal" compiler option `lualibCompilation` for compiling lualib.
 - In this mode, lualib source files are modules, with the following restrictions:
   - No `export default` or `export =`
   - ltems can only be imported from other lualib source files
  As this mode is meant to be "internal", these restrictions are currently not enforced (no checking or diagnostics).
- LuaLib feature dependencies are now automatically determined (instead of manually defined), and written to a file called `lualib_modules_info.json`.
- Fixes a number of issues with lualib features not being loaded.
 
## Additional details
Each lualib module compilations's output is such that it declares local variables that correspond to the module's exports,  potentially wrapping module statements in a `do` block if nessecary to avoid leaking variables out of scope. See sourceFile.ts for more info.
This allows lualib modules to have non-exported members, i.e. be written like a normal module.
The result can be directly inlined for `lualibImport: "inline"`, or for assembling `lualib_bundle`.

## Possible future changes
 - Change polyfilled global objects to actual objects, so that `const func = Object.assign` or `const foo = console` works.
 - Change lualib/builtin handling to be more declarative; Allow plugins to supply lualib functions or to modify lualib/builtin behavior.
 - Assemble a "custom" lualib_bundle for projects, which contains only used features.
 - Make it so that multiple tstl compiled projects (bundles) running in the same environment use the same lualib.

EDIT: reflect latest commits.
Thanks to @Zamiell for testing these changes in an existing project.


src/lualib/ArrayConcat.ts
```diff
@@ -1,4 +1,4 @@
-function __TS__ArrayConcat(this: void, arr1: any[], ...args: any[]): any[] {
+export function __TS__ArrayConcat(this: void, arr1: any[], ...args: any[]): any[] {
```
lolleko, line 1: Is it possible to make Array a class like set/map as well?
To give people the ability to overwrite array built ins.

EDIT: same applies to Object

GlassBricks: that might cause problems when interacting with existing lua code. Could be done in separate PR, see #262

Zamiell: i think it would be annoying for end users to have to mix Array and LuaArray, for example my game API returns lots of LuaArray

Zamiell:
I tried out this PR today with my mod, and everything seems to be working great so far.
There are no more global variables, as promised!
I will update this if I find any issues in the coming days.

src/transformation/visitors/class/index.ts
```diff
@@ -86,10 +85,6 @@ function transformClassLikeDeclaration(
     const superInfo = getOrUpdate(classSuperInfos, context, () => []);
     superInfo.push({ className, extendedTypeNode });
 
-    if (extendedType) {
```
Perryvw, line 89: Are you sure this can be removed?  Why does this not break promises construction?

src/transformation/visitors/sourceFile.ts
```diff
@@ -28,11 +32,39 @@ export const transformSourceFileNode: FunctionVisitor<ts.SourceFile> = (node, co
         }
     } else {
         pushScope(context, ScopeType.File);
-
         statements = performHoisting(context, context.transformStatements(node.statements));
         popScope(context);
 
-        if (context.isModule) {
+        if (context.options.luaLibCompilation) {
```
Perryvw, line 38: Why are you treating lualib statements different from regular statements?

src/transpilation/bundle.ts
```diff
@@ -104,7 +104,8 @@ export function getBundleResult(program: ts.Program, files: ProcessedFile[]): [t
     const footers: string[] = [];
     if (options.sourceMapTraceback) {
         // Generates SourceMapTraceback for the entire file
-        footers.push('require("lualib_bundle")\n');
+        // TODO: remove
+        footers.push('local __TS__SourceMapTraceBack = require("lualib_bundle").__TS__SourceMapTraceBack\n');
```
Perryvw, line 108: Should be removed?

src/transpilation/transpiler.ts
```diff
@@ -99,17 +100,21 @@ export class Transpiler {
...
 
         return { emitPlan };
     }
 }
 
-export function getEmitPath(file: string, program: ts.Program): string {
-    const relativeOutputPath = getEmitPathRelativeToOutDir(file, program);
+export function getEmitPath(file: string | ProcessedFile, program: ts.Program): string {
     const outDir = getEmitOutDir(program);
+    if (typeof file !== "string" && file.isRawFile) {
```
Perryvw, line 113: ew, why

src/transpilation/utils.ts
```diff
@@ -19,6 +19,7 @@ interface BaseFile {
     code: string;
     sourceMap?: string;
     sourceFiles?: ts.SourceFile[];
+    isRawFile?: boolean;
```
Perryvw, line 22: why is this even required?

test/transpile/module-resolution.spec.ts
```diff
@@ -490,7 +490,7 @@ test("includes lualib_bundle when external lua requests it", () => {
         .addExtraFile(
             "lualibuser.lua",
             `
-                require("lualib_bundle")
+                local __TS__ArrayPush = require("lualib_bundle").__TS__ArrayPush
```
Perryvw, line 493: why is this optimized into one statement here, but not in the cases in the snapshot below?

src/lualib/tsconfig.json
```diff
@@ -5,12 +5,12 @@
...
         "types": ["lua-types/5.4"],
         "skipLibCheck": true,
-
         "noUnusedLocals": true,
         "noUnusedParameters": true
     },
     "tstl": {
         "luaLibImport": "none",
-        "noHeader": true
+        "noHeader": true,
```
Perryvw, line 13: why not?

Perryvw (changes requested):

src/cli/parse.ts
```diff
@@ -83,6 +84,12 @@ export const optionDeclarations: CommandLineOption[] = [
         description: "An array of paths that tstl should not resolve and keep as-is.",
         type: "array",
     },
+    {
```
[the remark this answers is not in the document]
GlassBricks: I couldn't find a great way to introduce different compiler behavior otherwise. Do you have any suggestions?

src/LuaPrinter.ts
```diff
@@ -174,7 +173,7 @@ export class LuaPrinter {
 
     public print(file: lua.File): PrintResult {
         // Add traceback lualib if sourcemap traceback option is enabled
-        if (this.options.sourceMapTraceback) {
+        if (this.options.sourceMapTraceback && !isBundleEnabled(this.options)) {
```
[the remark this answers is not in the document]
GlassBricks: Individual files in bundles don't actually call sourcemaptraceback, instead one big call is done at the end of entire bundle, so this would import unnessecarily.

src/transformation/visitors/class/index.ts
GlassBricks: The check for lualib type has been moved to transformIdentifier. For a class extends, transformIdentifier is eventually called in createClassSetup.

src/transpilation/utils.ts
GlassBricks: This is for the json file. Before `getEmitPath` expects absolute paths and turns filenames into lua, and the json file name is not absolute. This allows treating the json file differently. 
If you can think of a better way to handle this let me know.

src/cli/parse.ts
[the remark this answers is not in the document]
GlassBricks: I changed the logic such that it does not touch the cli/external api, but still uses a field in `CompilerOptions`.

GlassBricks:
Closing in favor of #1224

PR #1217, opened 2022-02-25
Optimize lualib functions

### LINKED ISSUE #1130: Array.prototype methods are missing `thisArg`
array methods such as `array.map, array.findIndex`, etc. currently don't work with `thisArg`.

I might fix this as part of #1128

GlassBricks:
Optimizes lualib functions.
Also fixes #1130.
~~Depends on #1213, however it can be made independent.~~

Array.push with only 1 argument gets even more optimization, as it is very common:
 - Pushing a single element results in `array[#array + 1] = el`
 - Pushing a spread array of elements uses a new function `ArrayPushArray` to avoid unpacking/repacking the array.

Benchmarks on my machine show 10-70% time reduction for most functions on lua5.3 and luajit. As can be expected, lua5.3 has better relative improvement for many functions compared to luajit.

### Some optimization methods
 - caching global functions in local variables (when used more than once)
 - using a lua `for i in range` loop to iterate arrays


Zamiell:
Super excited for this! Well done glass!

src/lualib/ArrayUnshift.ts
```diff
@@ -1,6 +1,12 @@
-export function __TS__ArrayUnshift<T>(this: void, arr: T[], ...items: T[]): number {
-    for (let i = items.length - 1; i >= 0; --i) {
-        table.insert(arr, 1, items[i]);
+export function __TS__ArrayUnshift<T>(this: T[], ...items: T[]): number {
+    const length = items.length;
```
Perryvw, line 2: Length is confusing here because it's not `this.length`, suggest  to rename to `numItemsToInsert`

src/lualib/ObjectAssign.ts
```diff
@@ -1,15 +1,12 @@
 // https://tc39.github.io/ecma262/#sec-object.assign
 // eslint-disable-next-line @typescript-eslint/ban-types
-export function __TS__ObjectAssign<T extends object>(this: void, to: T, ...sources: object[]): T {
-    if (to === undefined) {
```
Perryvw, line 4: Can we just get rid of this check? Wouldn't this cause unexpected errors if `strictNullChecks` is disabled?

src/lualib/ArraySplice.ts
```diff
@@ -23,56 +27,62 @@ export function __TS__ArraySplice<T>(this: void, list: T[], ...args: T[]): T[] {
         actualDeleteCount = 0;
     } else if (actualArgumentCount === 1) {
         // ECMA-spec line 6: if number of actual arguments is 1
-        actualDeleteCount = len - actualStart;
+        actualDeleteCount = len - start;
     } else {
-        actualDeleteCount = Math.min(Math.max(deleteCount || 0, 0), len - actualStart);
+        actualDeleteCount = deleteCount || 0;
```
Perryvw, line 32: Should be `??` I guess

Perryvw (approved):

PR #1222, opened 2022-03-05
Upgrade TypeScript to 4.6

Perryvw:


.eslintrc.js
```diff
@@ -74,10 +74,8 @@ module.exports = {
         "jest/expect-expect": "off",
         "jest/consistent-test-it": ["error", { fn: "test", withinDescribe: "test" }],
         "jest/no-disabled-tests": "error",
-        "jest/no-expect-resolves": "error",
         "jest/no-identical-title": "off",
         "jest/no-test-return-statement": "error",
-        "jest/no-truthy-falsy": "error",
```
lolleko, line 80: Why do we need to remove this?

Perryvw: I updated eslint/jest and apparently this rule no longer exists

lolleko (approved):

PR #1225, opened 2022-03-06
Add warning diagnostic when using condition variable that will never be false

### LINKED ISSUE #1212: Add warning diagnostic when trying to use if statement expression that always evaluates to true in Lua
Maybe i missed something in the documentation or issues, but i cannot find anything about the difference in falsy values. The most obvious one being "0", and the transpilation of this code:
image
As far as i can tell, this should not print the value because 0 is false in typescript. But it will print it, because 0 is true in lua!

Is this intended? Am i wrong and 0 is actually true in typescript as well? I am quite confused right now

Perryvw:
Adds a warning diagnostic when using a type in a conditional which cannot be falsy in Lua (exception being `true`, that is still allowed).

Closes #1212


Perryvw:
Leads to issues in tests and lualib compilation, closing again for now.

Issue #1178, opened 2021-11-29
Ability to overwrite the bundle template / Ability to customize the comment header


Zamiell:
I want to automatically add a comment to the top of a "main.lua" file that is transpiled with TSTL's `luabundle` flag.
The idea is to explain to readers that this file was generated with a tool and is not intended to be read by humans, and that the actual source code for the file is available on GitHub and people should go there to read the source code.

Unfortunately, it sounds like performing this simple task is currently impossible, because:

1) TSTL does not support overwriting the bundle template.
2) TSTL does not support customizing the comment header that says: `--[[ Generated with https://github.com/TypeScriptToLua/TypeScriptToLua ]]`

I would love the ability to do one of these two things (preferably both!), either through a TSTL plugin, or through a TSTL compiler flag, or some other mechanism.

Zamiell:
I think that with the new features in TSTL 1.4.0, this should be possible now, so I'll close

Zamiell:
According to Perry, the `afterPrint` hook won't work with bundles, so this issue is still not resolved; I'll re-open.

Perryvw:
This is now somewhat possible with a `beforeEmit` plugin ([see docs](https://typescripttolua.github.io/docs/api/plugins#beforeemit)).

PR #1227, opened 2022-03-08
Lualib modules using a plugin to generate lualib

Perryvw:
This PR contains two big changes:

# Lualib files are now modules

- Lualib sources now use import/export
- Importing lualib functions now works the same as importing modules (in generated lua)
- Inlining lualib files is unchanged, except the imported functionality is now local instead of global
- The lualib dependency tree is no longer manually maintained, instead it is automatically generated during the lualib build.
  - The lualib dependency tree is distributed in our package as a json file.
- Lualib output is generated with a tstl plugin to minimize lualib-generation-specific code in the main transpiler. 

# Plugins now provide `beforeTransform` and `afterEmit` hooks

To generate the required lualib files we used a plugin that required some extra hooks. Because of this we introduced two extra TSTL plugin functions:

## beforeTansform

```ts
{
beforeTransform?: (
    program: ts.Program,
    options: CompilerOptions,
    emitHost: EmitHost
) => ts.Diagnostic[] | void;
}
```

This function on the plugin is called _after_ gathering the TypeScript program and options, but _before_ starting the transformation process. This can be used to set up stuff in the plugin, or to modify the program/options that the transpiler will act on.

## afterPrint

```ts
{
afterPrint?: (
    program: ts.Program,
    options: CompilerOptions,
    emitHost: EmitHost,
    result: ProcessedFile[]
) => ts.Diagnostic[] | void;
}
```

This function on the plugin is run _after_ the printing of Lua has completed. It provides the output array as one of its parameters, which can be used to modify the final output before finishing the emit.


Zamiell:
Question: Is it possible to use the new `afterPrint` hook to put a comment at the beginning of the file? e.g. #1178

Perryvw:
> Question: Is it possible to use the new `afterPrint` hook to put a comment at the beginning of the file? e.g. #1178

Yes, that is exaclty what [the added test for afterPrint](https://github.com/TypeScriptToLua/TypeScriptToLua/pull/1227/files#diff-0316d6848ad02ca629f4786e849d73fab2f2108ed839d4fd6d79beb5a419aca1) does.

Zamiell:
Woot! 🎉

build-lualib.js
```diff
@@ -1,17 +1,11 @@
 require("ts-node/register/transpile-only");
```
lolleko, line 1: I think we can actually get rid of this file. The reason for using this was because we couldn't just use command line tstl because we needed to write the bundle. That part is now gone, so we can just use tstl CLI in our build script.

src/lualib/Decorator.d.ts
```diff
@@ -0,0 +1,5 @@
+export type Decorator<TTarget extends AnyTable, TKey extends keyof TTarget> = (
```
lolleko, line 1: Decorator sounds extremely generic, what is this used for?

src/lualib-build/plugin.ts
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    protected createLuaLibModulesInfo(): { result: LuaLibModulesInfo; diagnostics: ts.Diagnostic[] } {
+        const result: Partial<LuaLibModulesInfo> = {};
+        const diagnostics: ts.Diagnostic[] = [];
+        for (const feature of Object.values(tstl.LuaLibFeature)) {
+            const exports = this.featureExports.get(feature);
+            if (!exports) {
+                diagnostics.push(lualibDiagnostic(`Missing file for lualib feature: ${feature}`));
+                console.error(`Missing file for lualib feature: ${feature}`);
```
lolleko, line 145: Isn't the diagnostic enough?

GlassBricks: Yes, this looks like a commit merge oversight

src/lualib/Decorator.d.ts
GlassBricks: It's used in both Decorate.ts and DecorateParam.ts, which are independent lualib features.

src/lualib-build/plugin.ts
Perryvw: Removed

build-lualib.js
Perryvw: Done, now invoking dist/tstl.js from package.json

lolleko (approved):

build-lualib.js
lolleko: Currently not possible due to unknown issue tracked in  #1232

Perryvw:
Trying to re-trigger checks

PR #1229, opened 2022-03-12
Allow configuring output file extension in config

Perryvw:
Got a request from a modding environment that does not use .lua but instead .scar files (with Lua content). Since there is no good way of changing extension otherwise, besides using some kind of post-build step or custom file watcher, I added it as a tsconfig option.

### DIFF
--- a/src/transpilation/transpiler.ts
+++ b/src/transpilation/transpiler.ts
@@ -127,8 +127,11 @@ export function getEmitPathRelativeToOutDir(fileName: string, program: ts.Progra
         emitPathSplits[0] = "lua_modules";
     }
 
-    // Make extension lua
-    emitPathSplits[emitPathSplits.length - 1] = trimExtension(emitPathSplits[emitPathSplits.length - 1]) + ".lua";
+    // Set extension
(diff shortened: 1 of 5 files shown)


src/transpilation/transpiler.ts
```diff
@@ -127,8 +127,11 @@ export function getEmitPathRelativeToOutDir(fileName: string, program: ts.Progra
         emitPathSplits[0] = "lua_modules";
     }
 
-    // Make extension lua
-    emitPathSplits[emitPathSplits.length - 1] = trimExtension(emitPathSplits[emitPathSplits.length - 1]) + ".lua";
+    // Set extension
+    const extension = ((program.getCompilerOptions() as CompilerOptions).extension ?? "lua").trim();
+    const trimmedExtension = extension.startsWith(".") ? extension.substring(1) : extension;
```
lolleko, line 132: Should we really strip the leading dot?
Maybe there are cases where it's intended, although i can't think of one.

Perryvw: I couldn't come up with a situation where doing it this way could be problematic, but I could see a situation where not specifying the dot would be nicer (ie specifying dot in CLI feels weird to me)

lolleko (approved):

PR #1230, opened 2022-03-13
Fix prototype call check for nullable types

### LINKED ISSUE #1218: Null-safe `?.toUpperCase()` on Interface property is converted to `:toUpperCase()` (a nil value)
[Playground example](https://typescripttolua.github.io/play/#code/FASwdgLgpgTgZgQwMZQAQFkCeARBEGoDewqpqYCAtlAFyoDOEM4A5qgD6oCuYAJlHHBRewAL7BgSAPZhGqSjjwEAvKgA8WXPgB8REmQrVUdHv0FhhYiQq0IAdIagB+OxCkBVAA6fYAYQT0UAAUAJQS0rIQ5FRodIzMYGycpgJCvKiqNkoOMcCOLm5ePjD+gaHhMnIWAO4AcjFxTKwc3HypFumqAEQARgA2CAAWqP1DIwODXXlQdTEFHt5+AcFhQA)

Given some TypeScript code that calls `toUpperCase` on the property of an interface:

```typescript
interface MyData {
    name: string | undefined
}

const myData = <MyData> {
    name : undefined
}

myData.name?.toUpperCase()
```

This produces Lua that fails with an error: `attempt to call method 'toUpperCase' (a nil value)`

```lua
myData = {name = nil}
local ____myData_name_toUpperCase_result_0 = myData.name
if ____myData_name_toUpperCase_result_0 ~= nil then
    ____myData_name_toUpperCase_result_0 = ____myData_name_toUpperCase_result_0:toUpperCase()
end
```

Extracting the property to a new value also produces invalid code.

```typescript
const name : string | undefined = myData.name
name?.toUpperCase()
```

```lua
name = myData.name
local ____name_toUpperCase_result_2 = name
if ____name_toUpperCase_result_2 ~= nil then
    ____name_toUpperCase_result_2 = ____name_toUpperCase_result_2:toUpperCase()
end
```

But otherwise, it works

```typescript
const newName: string | undefined = "blah blah blah"
newName?.toUpperCase()
```
```lua
-- correct
newName = "blah blah blah"
local ____newName_toUpperCase_result_4 = newName
if ____newName_toUpperCase_result_4 ~= nil then
    ____newName_toUpperCase_result_4 = string.upper(____newName_toUpperCase_result_4)
end
```

Perryvw:
Turns out that when we check for built-in prototype methods (like `string.prototype.substring` for example), we were not taking optional chaining into account. Therefore `strOrUndefined?.substring()` would lead to invalid Lua. This PR fixes the check for all prototype calls we detect, so they also work in case of optional chaining.

Fixes #1218

### DIFF
--- a/test/unit/builtins/string.spec.ts
+++ b/test/unit/builtins/string.spec.ts
@@ -367,3 +367,15 @@ test("string intersected method", () => {
         return ({ abc: () => "a" } as Vector).abc();
     `.expectToMatchJsResult();
 });
+
+// Issue #1218: https://github.com/TypeScriptToLua/TypeScriptToLua/issues/1218
+test.each(['"foo"', "undefined"])("prototype call on nullable string (%p)", value => {
+    util.testFunction`
+        function toUpper(str?: string) {
+            return str?.toUpperCase();
+        }
+        return toUpper(${value});
+    `
+        .setOptions({ strictNullChecks: true })
+        .expectToMatchJsResult();
+});
--- a/test/unit/builtins/string.spec.ts
+++ b/test/unit/builtins/string.spec.ts
@@ -379,3 +379,18 @@ test.each(['"foo"', "undefined"])("prototype call on nullable string (%p)", valu
         .setOptions({ strictNullChecks: true })
         .expectToMatchJsResult();
 });
+
+// Issue #1218: https://github.com/TypeScriptToLua/TypeScriptToLua/issues/1218
+test.each(["string | undefined", "string | null", "null | string", "null | undefined | string"])(
+    "prototype call on nullable string type (%p)",
+    type => {
+        util.testFunction`
+        function toUpper(str: ${type}) {
+            return str?.toUpperCase();
+        }
+        return toUpper("foo");
+    `
+            .setOptions({ strictNullChecks: true })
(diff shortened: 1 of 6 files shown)


test/unit/builtins/string.spec.ts
```diff
@@ -367,3 +367,15 @@ test("string intersected method", () => {
         return ({ abc: () => "a" } as Vector).abc();
     `.expectToMatchJsResult();
 });
+
+// Issue #1218: https://github.com/TypeScriptToLua/TypeScriptToLua/issues/1218
+test.each(['"foo"', "undefined"])("prototype call on nullable string (%p)", value => {
+    util.testFunction`
+        function toUpper(str?: string) {
```
aSemy, line 374: `str?: string` is one way of triggering the bug in #1218 - Is it also worth having tests for the other potential triggers, `str: string | undefined` and `str: string | null`? I only see tests for `str?: string`.

Perryvw: They should all resolve to the same type (kinda), but you are right, adding some more tests wouldn't hurt. I added some more.

lolleko (approved):

PR #1234, opened 2022-03-16
Fix preceding statements with ternary operator

### LINKED ISSUE #1231: Error in ternary operator
`const a = false ? (console.log("hello"), 4) : 5;`

Expected output:
```
a = (false and (function()
    console:log("hello")
    return 4
end)()) or 5
```
Actual output:
```
print("hello")
a = false and 4 or 5
```

GlassBricks:
Fixes #1231

### DIFF
--- a/test/unit/conditionals.spec.ts
+++ b/test/unit/conditionals.spec.ts
@@ -112,3 +112,29 @@ test.each([false, true, null])("Ternary conditional with generic whenTrue branch
         })
         .expectToMatchJsResult();
 });
+
+test.each([false, true])("Ternary conditional with preceding statements in true branch (%p)", trueVal => {
+    // language=TypeScript
+    util.testFunction`
+        let i = 0;
+        const result = ${trueVal} ? i += 1 : i;
+        return { result, i };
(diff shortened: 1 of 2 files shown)


test/unit/conditionals.spec.ts
```diff
@@ -112,3 +112,29 @@ test.each([false, true, null])("Ternary conditional with generic whenTrue branch
...
+
+test.each([false, true])("Ternary conditional with preceding statements in true branch (%p)", trueVal => {
+    // language=TypeScript
+    util.testFunction`
+        let i = 0;
+        const result = ${trueVal} ? i += 1 : i;
+        return { result, i };
+    `
+        .setOptions({
+            strictNullChecks: true,
```
Perryvw, line 124: Why is this option required here, I don't see any undefineds/nulls in the tests?

GlassBricks: Without it, the test doesn't start as failing and test the issue:
The type `number` with no strict null checks may be null and therefore falsy, so the if statement is used.

PR #1239, opened 2022-03-19
Added BeforeEmit plugin hook

### LINKED ISSUE #1236: afterPrint from the latest patch doesn't work
repro: https://github.com/Zamiell/tstl-bug

Perryvw:
Closes #1236

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -8,4 +8,4 @@ export { LuaLibFeature } from "./LuaLib";
 export * from "./LuaPrinter";
 export * from "./transformation/context";
 export * from "./transpilation";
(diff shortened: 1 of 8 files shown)


GlassBricks:
Just tested, this does _not_ fix #1236 yet.

Perryvw:
> @Perryvw Just tested, this does _not_ fix #1236 yet. I commented on that issue with more details.

Although I agree with your comments on that issue, that is a different problem from what #1236 is about. The question is how to place a comment at the top of a bundle file/make modifications to the final bundle, and `beforeEmit` allows you to do exactly that, although it could indeed break the source map.

PR #1240, opened 2022-03-19
Implement Array.from and Array.of

### LINKED ISSUE #1183: Feature request: Array.from
It's frustrating to get the red squiggly line for this.

I'd like to request that it gets added!

And failing that, please update the error message from `Array.from is unsupported.` to `Array.from is unsupported. Please use the spread operator instead.` Actionable advice makes a big difference here.

GlassBricks:
Closes #1183, alternative to #1235.

This allows constructs like `Array.from({ length: 3 }, (_, index) => index * 2)` (which works in js).

Array.of simply transforms into a table expression.


Zamiell:
much better than my PR ><
thank you!

PR #1247, opened 2022-04-01
Add new option to bypass 5.1 try/catch in async/await precaution

### LINKED ISSUE #1158: Ability to re-enable `try`/`catch` in `async`/`await` for Lua 5.1 as a config setting?
I use TypeScriptToLua with the Minecraft mod ComputerCraft, which runs on a version of Lua 5.1 with various 5.2/5.3 library features backported. I noticed that TSTL 1.1.0 removes support for `try`/`catch` inside `async`/`await` functions on Lua 5.1, which makes sense because of the fact that `pcall` can't yield; however, both ComputerCraft, which uses the [Cobalt](https://github.com/SquidDev/Cobalt) VM (based on LuaJ) in Java, and my emulator CraftOS-PC, which uses [a heavily patched version of Lua 5.1](https://github.com/MCJack123/craftos2-lua), do not suffer from this issue.

Would it be possible to add some sort of override inside `tsconfig.json` to allow exception handling inside `async` functions on 5.1? I do agree that the default behavior should be to disallow it, since standard Lua 5.1 is not compatible, but there are multiple 5.1 implementations that *do* support it, so it would be handy to be able to override that without having to switch to another version (which would then break `switch`) or edit the compiler manually.

hazzard993:
Closes #1158

Adds a very specific option to allow TSTL to proceed with transforming try/catch in async/await functions 

`lua51AllowTryCatchInAsyncAwait` (boolean) (false by default)

I figured `lua51` in the name to come first because this only has an effect on Lua 5.1 and isn't needed in any other targets

I can add more to this PR if needed just wanted to get an idea if this is the kind of solution expected 🤔 

Also planning to add some more details to the wiki about what specifically in the language we're looking for so that someone can figure out if this is a switch they should enable

### DIFF
--- a/src/CompilerOptions.ts
+++ b/src/CompilerOptions.ts
@@ -21,7 +21,7 @@ export interface LuaPluginImport {
     [option: string]: any;
 }
 
-export type CompilerOptions = OmitIndexSignature<ts.CompilerOptions> & {
+export interface TypeScriptToLuaOptions {
     buildMode?: BuildMode;
     extension?: string;
     luaBundle?: string;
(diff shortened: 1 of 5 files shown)


Perryvw:
Looks fine to me

Issue #1259, opened 2022-04-16
debug.traceback of loadstring get unknown file


pilaoda:
My traceback:
```
Script Runtime Error: unknown:743: attempt to perform arithmetic on local 'c' (a string value)
stack traceback:
    unknown:743: in function 'handler'
    unknown:189: in function <unknown:186>
```

My code:
```
local encryptedCode = "XXXXXX"
local decryptedCode = decrypt(encryptedCode)
loadstring(decryptedCode , "sys/Cheat.lua")()
```

What I have done is encrypt the generated lua file, and use `loadstring` after decryption.
I already set the file path as the second parameter `chunkname` of loadstring, but the traceback still be replaced by this code: https://github.com/TypeScriptToLua/TypeScriptToLua/blob/288ab8cd1c7e1736578db0b31fd23f62c0ccfce6/src/lualib/SourceMapTraceBack.ts#L46

What I expect is using the sourcemap with the `chunkname` I set in loadstring.
I have two ideas to achieve this:
1.  Assume the chunkname is the lua path, and replace the traceback substring `[string "xx/xx.lua"]:22` to `xx/xx.ts:10` or `[string "xx/xx.ts"]:10`
2.  Provide a **custom replacer** of `loadstring` for users to implement by themselves.

pilaoda:
If my idea is accepted, I would like to write a PR for this.

Perryvw:
Option 1 sounds fine to me, I expected that to work already, PR to correct the behaviour is welcome.

pilaoda:
PR to fix this with Option1: #1260

PR #1260, opened 2022-04-17
fix: debug.traceback of loadstring get unknown file #1259

pilaoda:
Assume the chunkname is the lua path, and replace the traceback substring [string "xx/xx.lua"]:22 to xx/xx.ts:10

### DIFF
--- a/test/unit/printer/sourcemaps.spec.ts
+++ b/test/unit/printer/sourcemaps.spec.ts
@@ -1,5 +1,6 @@
 import { SourceMapConsumer } from "source-map";
 import * as tstl from "../../../src";
+import { LuaTarget } from "../../../src";
 import { couldNotResolveRequire } from "../../../src/transpilation/diagnostics";
 import * as util from "../../util";
 import { lineAndColumnOf } from "./utils";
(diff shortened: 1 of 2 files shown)


Perryvw (changes requested):
Looks alright, could you see if you can add a test that fails without your fix and fix the linter errors by running `npm run fix:prettier` ?

Perryvw:
Can you add a test that would fail without your changes, but passes on your branch? Maybe something throwing an error in loadstring like your issue mentions. Add it here: https://github.com/TypeScriptToLua/TypeScriptToLua/blob/master/test/unit/printer/sourcemaps.spec.ts#L286

pilaoda:
@Perryvw Hope I done right this time.

pilaoda:
> Can you add a test that would fail without your changes, but passes on your branch? Maybe something throwing an error in loadstring like your issue mentions. Add it here: https://github.com/TypeScriptToLua/TypeScriptToLua/blob/master/test/unit/printer/sourcemaps.spec.ts#L286

This test is really complicated. I tried my best.

test/unit/printer/sourcemaps.spec.ts
```diff
@@ -326,3 +327,28 @@ test("Inline sourcemaps", () => {
...
+    `;
+    const builder = util.testModule`
+        return (require as (this: void, modname: string) => any)("foo");
+    `
+        .addExtraFile("foo.ts", file)
+        .setOptions({ sourceMapTraceback: true, luaTarget: LuaTarget.Lua51 });
+
+    const traceback = builder.getLuaExecutionResult();
+    expect(traceback).toContain("foo.ts");
+});
```
Perryvw, line 354: This looks way too complicated, couldn't you just do something like:
```ts
    const builder = util.testFunction`
        return loadstring(
            "return debug.traceback();",
            "foo.lua"
        )();
    `
        .setTsHeader("declare function loadstring(this: void, str: string, chunkName: string): () => unknown;")
        .setOptions({ sourceMapTraceback: true, luaTarget: LuaTarget.Lua51 });

    const traceback = builder.getLuaExecutionResult();
    expect(traceback).toContain("foo.ts");
```

pilaoda: > This looks way too complicated, couldn't you just do something like:
> 
> ```ts
>     const builder = util.testFunction`
>         return loadstring(
>             "return debug.traceback();",
>             "foo.lua"
>         )();
>     `
>         .setTsHeader("declare function loadstring(this: void, str: string, chunkName: string): () => unknown;")
>         .setOptions({ sourceMapTraceback: true, luaTarget: LuaTarget.Lua51 });
> 
>     const traceback = builder.getLuaExecutionResult();
>     expect(traceback).toContain("foo.ts");
> ```

This is intentional.
The compiled lua code and sourcemap info should be passed into the loadstring together , otherelse the foo.lua could not be mapped back to foo.ts .

Perryvw: I played around with it a bit, a much nicer test seems to be:
```ts
const loadStrCode = transpileString(
        `function bar() {
            const trace = (debug.traceback as (this: void)=>string)();
            return trace;
        }
        return bar();`,
        { sourceMapTraceback: true, luaTarget: LuaTarget.Lua51 }
    ).file?.lua;

        const luaCode = \`${loadStrCode}\`;
        return loadstring(luaCode, "foo.lua")();
    `
        .setTsHeader("declare function loadstring(this: void, string: string, chunkname: string): () => unknown")
        .setOptions({ sourceMapTraceback: true, luaTarget: LuaTarget.Lua51 });

    const traceback = builder.getLuaExecutionResult();
    expect(traceback).toContain("foo.ts");
```
Seems to work and is much neater.

pilaoda: Cool. I will commit with your nicer test.

PR #1263, opened 2022-04-20
Lua 5.0 support

YoRyan:
This PR is a first attempt at introducing support for Lua 5.0, a version that some older games remain stuck on. It adds a 5.0 LuaTarget that triggers the appropriate syntax for variadic arguments and length-checking and modulo operations.

Lua 5.0, while mostly similar to 5.1, is not forwards-compatible; it is not possible to write a transpiler that is simultaneously compatible with both 5.0 and 5.1 syntax. Hence, if you choose to accept support for 5.0, you might also need to rename the "Universal" target to something more descriptive, like "5.1+".

Next steps: Adding tests, adding the appropriate type annotations for the 5.0 standard library to the lua-types package, and updating lualib to be 5.0-compatible.


lolleko:
I guess for this PR it is not yet required. But in the future we will need to add 5.0 to our WASM bindings, that way we can run tests for 5.0 properly.

src/transformation/builtins/function.ts
```diff
@@ -40,7 +40,11 @@ export function transformFunctionProperty(
 ): lua.Expression | undefined {
     switch (node.name.text) {
         case "length":
-            if (context.luaTarget === LuaTarget.Lua51 || context.luaTarget === LuaTarget.Universal) {
+            if (
+                context.luaTarget === LuaTarget.Lua50 ||
+                context.luaTarget === LuaTarget.Lua51 ||
+                context.luaTarget === LuaTarget.Universal
+            ) {
                 context.diagnostics.push(unsupportedForTarget(node, "function.length", LuaTarget.Lua51));
```
Perryvw, line 48: Noticed a small bug in our code, could you take it along: I think this diagnostic should actually get `context.luaTarget` instead of hardcoding `LuaTarget.Lua51`

src/transformation/visitors/break-continue.ts
```diff
@@ -11,7 +11,11 @@ export const transformBreakStatement: FunctionVisitor<ts.BreakStatement> = (brea
 };
 
 export const transformContinueStatement: FunctionVisitor<ts.ContinueStatement> = (statement, context) => {
-    if (context.luaTarget === LuaTarget.Universal || context.luaTarget === LuaTarget.Lua51) {
+    if (
+        context.luaTarget === LuaTarget.Universal ||
+        context.luaTarget === LuaTarget.Lua50 ||
+        context.luaTarget === LuaTarget.Lua51
+    ) {
         context.diagnostics.push(unsupportedForTarget(statement, "Continue statement", LuaTarget.Lua51));
```
Perryvw, line 19: Same here, this diagnostic should get context.luaTarget instead of hardcoding Lua51

src/transformation/visitors/function.ts
```diff
@@ -142,7 +143,12 @@ export function transformFunctionBodyHeader(
 
     // Push spread operator here
     if (spreadIdentifier && isRestParameterReferenced(spreadIdentifier, bodyScope)) {
-        const spreadTable = wrapInTable(lua.createDotsLiteral());
+        let spreadTable: lua.Expression;
+        if (context.luaTarget === LuaTarget.Lua50) {
+            spreadTable = lua.createArgLiteral();
+        } else {
+            spreadTable = wrapInTable(lua.createDotsLiteral());
+        }
```
Perryvw, line 151:
```suggestion
        let spreadTable = context.luaTarget === LuaTarget.Lua50
            ? lua.createArgLiteral()
            : wrapInTable(lua.createDotsLiteral());
```

test/unit/__snapshots__/expressions.spec.ts.snap
```diff
@@ -36,6 +36,116 @@ ____exports.__result = 10 - (4 + 5)
 return ____exports"
 `;
 
+exports[`Bitop [5.0] ("~a"): code 1`] = `
+"local ____exports = {}
+____exports.__result = bit.bnot(a)
+return ____exports"
+`;
+
+exports[`Bitop [5.0] ("~a"): diagnostics 1`] = `"main.ts(1,25): error TSTL: Bitwise operations is/are not supported for target Lua 5.1."`;
```
Perryvw, line 45: Yeah this message is wrong, should be fixed by addressing my previous comments.

test/unit/expressions.spec.ts
```diff
@@ -48,6 +48,14 @@ test.each(["a+=b", "a-=b", "a*=b", "a/=b", "a%=b", "a**=b"])("Binary expressions
 const supportedInAll = ["~a", "a&b", "a&=b", "a|b", "a|=b", "a^b", "a^=b", "a<<b", "a<<=b", "a>>>b", "a>>>=b"];
 const unsupportedIn53And54 = ["a>>b", "a>>=b"];
 const allBinaryOperators = [...supportedInAll, ...unsupportedIn53And54];
+test.each(allBinaryOperators)("Bitop [5.0] (%p)", input => {
```
Perryvw, line 51: This test is good, but I'm still missing a `testEachVersion` test in this file testing `%` modulo, since that is now version-specific.

test/unit/functions/functions.spec.ts
```diff
@@ -216,14 +216,17 @@ test.each([
...
 
-test.each([tstl.LuaTarget.Lua51, tstl.LuaTarget.Universal])("function.length unsupported (%p)", luaTarget => {
-    util.testFunction`
+test.each([tstl.LuaTarget.Lua50, tstl.LuaTarget.Lua51, tstl.LuaTarget.Universal])(
+    "function.length unsupported (%p)",
+    luaTarget => {
+        util.testFunction`
         function fn() {}
         return fn.length;
     `
```
Perryvw, line 225: indenting seems a bit wonky here, though that might be github struggling to show whitespace diffs. Normally we indent code in our test functions 1 more than the util.testFunction line they are for:
```suggestion
        util.testFunction`
            function fn() {}
            return fn.length;
        `
```

test/unit/spread.spec.ts
```diff
@@ -76,6 +76,7 @@ describe("in function call", () => {
         {
             [tstl.LuaTarget.Universal]: builder => builder.tap(expectLualibUnpack).expectToMatchJsResult(),
             [tstl.LuaTarget.LuaJIT]: builder => builder.tap(expectUnpack),
+            [tstl.LuaTarget.Lua50]: builder => builder.tap(expectUnpack).expectToMatchJsResult(),
```
Perryvw, line 79: Maybe I'm overlooking something, but do any of these existing tests cover the arg literal you added?

Perryvw (changes requested):
Changes look good to me, just had some comments related to error messages and tests.

src/transformation/visitors/function.ts
lolleko: shouldnt it be `const spreadTable`

Perryvw: Yes it should, although I'm sure eslint would complained either way 😅

src/transformation/builtins/function.ts
YoRyan: Fixed with 635d271841dd637f8ef44f4088c36fdd9df2ed7f!

src/transformation/visitors/function.ts
YoRyan: Fixed with 2836859476950fb158ffb0dc4bff9fdefc5eeb24 (using `const`!).

test/unit/expressions.spec.ts
YoRyan: Added a special test for `%` with fac16d27576b541b3f5d3b248999b995e373b8fa. (Did I do that right?)

test/unit/spread.spec.ts
YoRyan: No, I think you're onto something. These tests have to be split into 5.0 and non-5.0 cases and the snapshots updated to expect `arg` for 5.0.

test/unit/expressions.spec.ts
```diff
@@ -18,8 +18,56 @@ test.each([
...
+        .setOptions({ luaTarget: tstl.LuaTarget.Lua53, luaLibImport: tstl.LuaLibImportKind.None })
+        .disableSemanticCheck()
+        .expectToMatchJsResult();
+});
+
+test.each(allNumericOperators)("Binary expressions basic numeric [5.4] (%p)", input => {
+    util.testExpression(input)
+        .setOptions({ luaTarget: tstl.LuaTarget.Lua54, luaLibImport: tstl.LuaLibImportKind.None })
+        .disableSemanticCheck()
+        .expectToMatchJsResult();
```
Perryvw, line 70: This seems way too complicated, seems to me these 50 lines could be replaced with just:
```ts
for (const expression of ["3+4", "5-2", "6*3", "6**3", "20/5", "15/10", "15%3"]) {
    util.testEachVersion(
        `Binary expressions basic numeric (${expression})`,
        () => util.testExpression(expression),
        util.expectEachVersionExceptJit(b => b.expectToMatchJsResult())
    );
}
```
All of these should be supported in all versions, and we want to minimize the use of snapshots so that can just be left out.

test/unit/__snapshots__/expressions.spec.ts.snap
```diff
@@ -36,6 +36,116 @@ ____exports.__result = 10 - (4 + 5)
...
+
+exports[`Bitop [5.0] ("a&=b"): code 1`] = `
+"local ____exports = {}
+a = bit.band(a, b)
+____exports.__result = a
+return ____exports"
+`;
+
+exports[`Bitop [5.0] ("a&=b"): diagnostics 1`] = `"main.ts(1,25): error TSTL: Bitwise operations is/are not supported for target Lua 5.1."`;
```
Perryvw, line 54: This snapshot or test seems to not be updated yet

Perryvw (changes requested):

test/unit/expressions.spec.ts
YoRyan: Very well. Although I assume we can't execute any Lua 5.0 code until we add the wasm bindings for that version.

Perryvw: Correct, unfortunately we cannot test that now, but a snapshot still wouldn't really test if that actually worked in 5.0 either

test/unit/spread.spec.ts
Perryvw, line 79: You cannot expect to match JS result here (as the test failures show) because we cannot execute Lua 5.0 yet

test/unit/__snapshots__/expressions.spec.ts.snap
YoRyan: Fixed with 64e5f6d9f572ab04cd31895e5486bacaf354152e.

test/unit/spread.spec.ts
YoRyan: Since @lolleko is working on bindings for 5.0, I'll keep the match JS result here.

YoRyan: Refactored the spread tests with [a5f39f6](https://github.com/TypeScriptToLua/TypeScriptToLua/pull/1263/commits/a5f39f617bae32741ded2f6e08c580f103ae4b94). Since @lolleko is working on bindings for 5.0, I'll keep the match JS result calls.

test/unit/expressions.spec.ts
YoRyan: Okay, I've adopted your approach with 555e2f855719e7830d446e857fc58d48dd57b00f.

YoRyan:
I've submitted the type declarations with pull request https://github.com/TypeScriptToLua/lua-types/pull/59.

YoRyan:
Okay, so here's a status update on this porting effort:

- [X] Added a 5.0 mode to TSTL.
- [X] Added type declarations for the 5.0 standard library (https://github.com/TypeScriptToLua/lua-types/pull/59).
- [X] Made lualib compatible with 5.0 (see my branch [here](https://github.com/YoRyan/TypeScriptToLua/tree/lua5.0-lualib); file another PR after merging this one?).

Still to do:

- [ ] Lualib is currently only built for the `universal` target, which is not compatible with 5.0's syntax. We need to ship a second copy of lualib built for the `5.0` target.

YoRyan:
Okay, so here's a status update on this porting effort:

- [X] Added a 5.0 mode to TSTL.
- [X] Added type declarations for the 5.0 standard library (https://github.com/TypeScriptToLua/lua-types/pull/59).
- [X] Made lualib compatible with 5.0 (see my branch [here](https://github.com/YoRyan/TypeScriptToLua/tree/lua5.0-lualib); file another PR after merging this one?).

Still to do:

- [ ] Lualib is currently only built for the `universal` target, which is not compatible with 5.0's syntax. We need to ship a second copy of lualib built for the `5.0` target.
- [ ] Add bindings to test against the 5.0 interpreter (with @lolleko).
- [ ] Update the documentation.

lolleko:
FYI: I added bindings for 5.0 a while back, you can bump the `lua-wasm-bindings` package in this repo to [0.2.3-rc.0](https://github.com/TypeScriptToLua/lua-wasm-bindings/releases/tag/v0.2.3-rc.0). You have to adapt our test framework a little to also run tests for 5.0. Let me know if you need help with that.

YoRyan:
Ok, so I've integrated my lualib changes and added a second lualib build specifically for 5.0. I've also made the necessary changes to the testing framework to run against the new 5.0 bindings.  With https://github.com/TypeScriptToLua/lua-wasm-bindings/pull/3 applied, most of these tests pass! And you can now run TSTL to transpile to 5.0-compatible code.

That means we have:

- [x] Working syntax transpilation for 5.0
- [x] Type declarations for 5.0
- [x] Lualib ported to 5.0
- [x] Lua bindings for 5.0 (including the ability to inject transpiled files, which in 5.0 has to happen via the `_LOADED` table rather than `package.preload`)
- [x] The ability to output code for 5.0

So, we're almost at the finish line. All that's left to do is:

- [ ] Investigate the remaining failing tests
- [ ] Update the project documentation

Some of the failures seem to have something to do with a failing lualib import:

```
[string "local function __TS__ArrayIsArray(value)..."]:2598: `}' expected (to close `{' at line 2490) near `<eof>'
```

It's as if the last line of the lualib_bundle.lua file is cut off, because it loads just fine when fed to the real interpreter. Is the end of the string being corrupted when passed through the C API, somehow? I have spent some hours investigating to no avail.

YoRyan:
All tests should pass with the latest commits.

src/LuaLib.ts
```diff
@@ -122,8 +128,9 @@ export function getLuaLibModulesInfo(emitHost: EmitHost): LuaLibModulesInfo {
     return luaLibModulesInfo;
 }
 
-export function readLuaLibFeature(feature: LuaLibFeature, emitHost: EmitHost): string {
-    const featurePath = path.resolve(__dirname, `../dist/lualib/${feature}.lua`);
+export function readLuaLibFeature(feature: LuaLibFeature, luaTarget: LuaTarget, emitHost: EmitHost): string {
+    const lualibDir = `../dist/${luaTarget === LuaTarget.Lua50 ? "lualib-lua50" : "lualib"}`;
```
Perryvw, line 132: Duplicate code with line 119, also we should probably be using path.join, proposal:
```ts
function resolveLuaLibDir(target: LuaTarget) {
    const lualibDir = luaTarget === LuaTarget.Lua50 ? "lualib-lua50" : "lualib";
    return path.resolve(__dirname, "dist", lualibDir);
}
```

src/lualib-lua50/tsconfig.json
```diff
@@ -0,0 +1,21 @@
...
+        "noUnusedLocals": true,
+        "noUnusedParameters": true
+    },
+    "tstl": {
+        "luaLibImport": "none",
+        "noHeader": true,
+        "luaTarget": "5.0",
+        "luaPlugins": [{ "name": "../../dist/lualib-build/plugin.js" }]
+    },
+    "include": ["../lualib", "../../language-extensions/index.d.ts"]
```
Perryvw, line 20: Could you just add this file as tsconfig.lua50.json to `src/lualib` instead? I don't like this new directory

src/lualib/CountVarargs.ts
```diff
@@ -0,0 +1,11 @@
+/** @noSelfInFile */
+
+import { __TS__IsLua50 } from "./IsLua50";
+
+export function __TS__CountVarargs<T>(...args: T[]): number {
+    // select() is not available in Lua 5.0. In that version, the arg table
+    // includes trailing nils.
+    // It is important that the select() branch come first as we need vararg
+    // optimization for this call.
+    return !__TS__IsLua50 ? select("#", ...args) : args.length;
```
Perryvw, line 10: probably better if you just check for the existence of `select`  and get rid of IsLua50

src/lualib/IsLua50.ts
```diff
@@ -0,0 +1,4 @@
+/** @noSelfInFile */
+
+export const __TS__IsLua50 =
```
Perryvw, line 3: This looks like a bad idea, please remove.

src/lualib/Match.ts
```diff
@@ -0,0 +1,18 @@
+/** @noSelfInFile */
+
+import { __TS__IsLua50 } from "./IsLua50";
+
+export function __TS__Match(s: string, pattern: string, init?: number): LuaMultiReturn<string[]> {
+    if (__TS__IsLua50) {
```
Perryvw, line 6: use `if (string.find)` instead

src/lualib/NumberToString.ts
```diff
@@ -1,5 +1,16 @@
+import { __TS__IsLua50 } from "./IsLua50";
+
 const radixChars = "0123456789abcdefghijklmnopqrstuvwxyz";
 
+function modf(this: void, x: number): LuaMultiReturn<[number, number]> {
+    if (__TS__IsLua50) {
```
Perryvw, line 6: use `if math.modf` instead

src/transformation/builtins/array.ts
```diff
@@ -29,6 +30,8 @@ export function transformArrayConstructorCall(
     }
 }
 
+const lua50TableLength = lua.createTableIndexExpression(lua.createIdentifier("table"), lua.createStringLiteral("getn"));
```
Perryvw, line 33: you should probably only make this when it is requested, and add the last parameter (tsOriginal node)

src/transpilation/bundle.ts
```diff
@@ -22,7 +24,9 @@ local function require(file, ...)
     end
     if ____modules[file] then
         local module = ____modules[file]
-        ____moduleCache[file] = { value = (select("#", ...) > 0) and module(...) or module(file) }
+        ____moduleCache[file] = { value = (${isLua50 ? "table.getn(arg)" : 'select("#", ...)'} > 0) and module(${
```
Perryvw, line 27: This is unreadable, please pre-format this entire table literal beforehand

test/unit/spread.spec.ts
```diff
@@ -4,6 +4,7 @@ import { formatCode } from "../util";
 
 // TODO: Make some utils for testing other targets
 const expectUnpack: util.TapCallback = builder => expect(builder.getMainLuaCodeChunk()).toMatch(/[^.]unpack\(/);
+const expectArg: util.TapCallback = builder => expect(builder.getMainLuaCodeChunk()).toMatch(/ arg/);
```
Perryvw, line 7: what's with this space before arg in the pattern? That doesn't seem right, also this pattern seems way too easy to accidentally pass, you should probably use more specific patterns, like 'unpack(arg)' for example

test/util.ts
```diff
@@ -9,16 +9,27 @@ import * as ts from "typescript";
...
 import * as tstl from "../src";
 import { createEmitOutputCollector } from "../src/transpilation/output-collector";
-import { EmitHost, getEmitOutDir, transpileProject } from "../src";
+import { EmitHost, getEmitOutDir, LuaTarget, transpileProject } from "../src";
 import { formatPathToLuaPath, normalizeSlashes } from "../src/utils";
 
-const jsonLib = fs.readFileSync(path.join(__dirname, "json.lua"), "utf8");
-const luaLib = fs.readFileSync(path.resolve(__dirname, "../dist/lualib/lualib_bundle.lua"), "utf8");
+function readLuaLib(target: tstl.LuaTarget) {
+    const luaLibDir = `../dist/${target === LuaTarget.Lua50 ? "lualib-lua50" : "lualib"}`;
```
Perryvw, line 16: Also duplicated with an earlier comment I made, you can export that function and import & use it here

test/util.ts
```diff
@@ -57,7 +68,7 @@ export function testEachVersion<T extends TestBuilder>(
         const testName = name === undefined ? version : `${name} [${version}]`;
         defineTest(testName, () => {
             const builder = common();
-            builder.setOptions({ luaTarget: version });
+            builder.setOptions({ luaTarget: version, luaLibImport: tstl.LuaLibImportKind.Inline });
```
Perryvw, line 71: Why, this was require for performance reasons

Perryvw (changes requested):
I do not agree with this 'one lualib for all versions' setup. I much prefer the way it was set up before, where there was a separate src/lualib/lua50 directory with 5.0 specific lualib functions (if required). The transpiler can then decide which files (or which bundle) to emit based on the target.

The current setup with branches and if statements is messy, hard to read, and imposes a runtime penalty that can simply be avoided.

src/lualib-lua50/tsconfig.json
YoRyan: Good idea. Implemented with https://github.com/TypeScriptToLua/TypeScriptToLua/pull/1263/commits/ced3532e7e16e9f13ce62e5f0a1b73960d1be308.

src/LuaLib.ts
YoRyan: Implemented with https://github.com/TypeScriptToLua/TypeScriptToLua/pull/1263/commits/e775d2b45557ed89c5f3a31c709f71d0e49b3ed4.

YoRyan: Implemented with https://github.com/TypeScriptToLua/TypeScriptToLua/commit/28cc8cdd5be3ef523c9bc8905676fc02c51755e0.

test/util.ts
YoRyan: Implemented with https://github.com/TypeScriptToLua/TypeScriptToLua/commit/28cc8cdd5be3ef523c9bc8905676fc02c51755e0.

src/transformation/builtins/array.ts
YoRyan: Changed to a function with https://github.com/TypeScriptToLua/TypeScriptToLua/pull/1263/commits/d1bf66a73f8ec284d93cd6de239847d699c2c566. I don't believe `node` needs to be passed to the first usage, as it's passed to the assignment expression instead.

src/transpilation/bundle.ts
YoRyan: Done with https://github.com/TypeScriptToLua/TypeScriptToLua/pull/1263/commits/f2563cc76ddd7a638a71e437448faec4669f1aed.

test/unit/spread.spec.ts
YoRyan: It looks like I used this pattern for testing (before we had bindings to execute 5.0) and forgot to remove it. It was supposed to check for translation to `= arg`, which is correct for 5.0, rather than `= {...}`, as is used for the other versions. It was also extremely crude and, clearly, should never have been committed...

I've reinstated the unpack test because verifying the translation is redundant; if it was wrong, it wouldn't execute or match the JS result. See https://github.com/TypeScriptToLua/TypeScriptToLua/pull/1263/commits/4ab5891c38f84773ef3e70c7b35985581267c7c1.

YoRyan:
> I do not agree with this 'one lualib for all versions' setup. I much prefer the way it was set up before, where there was a separate src/lualib/lua50 directory with 5.0 specific lualib functions (if required). The transpiler can then decide which files (or which bundle) to emit based on the target.

I agree that ideally we wouldn't need to bother with any runtime checks, but I can't identify an easy way to switch between non-5.0 and 5.0 source code, while keeping import paths and other metadata intact. Lualib just isn't structured that way; I'll need more time to study the issue. In the meantime, I've replaced the exported constant with checks that use the built-ins with https://github.com/TypeScriptToLua/TypeScriptToLua/pull/1263/commits/fdcea36ba61659580a8f53c8734b97c391ba2104.

test/util.ts
YoRyan: Three tests are failing to import lualib if that option is removed, as I noted earlier. Something to do with the large size of the bundle, as best as I can tell? So switching to inline imports was my solution. Guess I'll have to come up with another answer...

YoRyan:
I managed to achieve different source files for different Lua targets with https://github.com/TypeScriptToLua/TypeScriptToLua/pull/1263/commits/f390622c5b7d50376cb5d121a957162c55f31657! I used tsconfig.json's `rootDirs` feature while also stripping the sub-directories from the output paths as they are passed through the build plugin. This way, we have no version-specific branches or runtime checks.

test/util.ts
YoRyan: These failures are fixed with https://github.com/TypeScriptToLua/TypeScriptToLua/pull/1263/commits/e0d6206ba44de81a8455103999b55236dac30371 and https://github.com/TypeScriptToLua/lua-wasm-bindings/pull/5.

YoRyan: Those failures are fixed with https://github.com/TypeScriptToLua/TypeScriptToLua/pull/1263/commits/e0d6206ba44de81a8455103999b55236dac30371 and https://github.com/TypeScriptToLua/lua-wasm-bindings/pull/5. I have reverted this change.

YoRyan:
I've addressed all feedback and believe this is ready for a re-review.

You need https://github.com/TypeScriptToLua/lua-types/pull/75 and https://github.com/TypeScriptToLua/lua-wasm-bindings/pull/5 to get the tests to pass.

src/transpilation/bundle.ts
```diff
@@ -12,7 +12,31 @@ const createModulePath = (pathToResolve: string, program: ts.Program) =>
...
+local ____modules = {}
+local ____moduleCache = {}
+local ____originalRequire = require
+local function require(file, ...)
+    if ____moduleCache[file] then
+        return ____moduleCache[file].value
+    end
+    if ____modules[file] then
+        local module = ____modules[file]
+        ____moduleCache[file] = { value = (table.getn(arg) > 0) and module(unpack(arg)) or module(file) }
```
Perryvw, line 27: Oh I didnt mean this entire string, this is just code duplication. I meant you could just store this one line that is different in a variable beforehand, then format it in later:
```ts
const setModuleCacheLine = options.luaTarget == LuaTarget.Lua50
    ? "{ value = (table.getn(arg) > 0) and module(unpack(arg)) or module(file) }"
    : "{ value = (select("#", ...) > 0) and module(...) or module(file) }"
```

src/LuaLib.ts
```diff
@@ -46,8 +48,10 @@ export enum LuaLibFeature {
     Iterator = "Iterator",
     LuaIteratorSpread = "LuaIteratorSpread",
     Map = "Map",
+    Match = "Match",
     MathAtan2 = "MathAtan2",
     MathSign = "MathSign",
+    Modulo = "Modulo",
```
Perryvw, line 54: If this modulo is only required for 50, maybe we should call this `Modulo50`

src/LuaLib.ts
```diff
@@ -46,8 +48,10 @@ export enum LuaLibFeature {
     Iterator = "Iterator",
     LuaIteratorSpread = "LuaIteratorSpread",
     Map = "Map",
+    Match = "Match",
```
Perryvw, line 51: Since this and CountVarargs are only used by the lualib internally and not by the transpiler, could we leave them out of this enum? Or would that somehow mess up the lualib compilation?

Perryvw (changes requested):

src/LuaLib.ts
```diff
@@ -122,8 +131,8 @@ export function getLuaLibModulesInfo(emitHost: EmitHost): LuaLibModulesInfo {
     return luaLibModulesInfo;
 }
 
-export function readLuaLibFeature(feature: LuaLibFeature, emitHost: EmitHost): string {
-    const featurePath = path.resolve(__dirname, `../dist/lualib/${feature}.lua`);
+export function readLuaLibFeature(feature: LuaLibFeature, luaTarget: LuaTarget, emitHost: EmitHost): string {
+    const featurePath = path.join(resolveLuaLibDir(luaTarget), `${feature}.lua`);
```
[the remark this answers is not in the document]
YoRyan: rootDirs flattens the import paths, so we can seamlessly combine the common files in lualib/ with the target-specific files in lualib/universal/ and lualib/5.0/ without affecting the imports. However, rootDirs doesn't affect output: tsc always reproduces the source directory structure in the outdir. Therefore, we also [strip the unwanted subdirectories from the emitted file paths](https://github.com/YoRyan/TypeScriptToLua/blob/2dc07a824c6424f25fd74529343f212cb16b5efc/src/lualib-build/plugin.ts#L46).

src/transpilation/bundle.ts
YoRyan: Yes, of course. Done with https://github.com/TypeScriptToLua/TypeScriptToLua/pull/1263/commits/471554074a6defbb96fb27011a9c28e508e34e34.

src/LuaLib.ts
YoRyan: Currently, any file in lualib/ [has to be named](https://github.com/YoRyan/TypeScriptToLua/blob/471554074a6defbb96fb27011a9c28e508e34e34/src/lualib-build/plugin.ts#L78) in this enum or the compilation will fail. The lualib builder strips the import statements from every file and constructs its own dependency graph.

src/lualib/universal/Match.ts
```diff
@@ -0,0 +1,5 @@
+/** @noSelfInFile */
+
+export function __TS__Match(s: string, pattern: string, init?: number): LuaMultiReturn<string[]> {
```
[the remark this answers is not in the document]
YoRyan: Yes. Done with https://github.com/TypeScriptToLua/TypeScriptToLua/pull/1263/commits/3bc745ced91aab80034f6b33d7b2f88def2be213.

src/lualib/Modulo.ts
```diff
@@ -0,0 +1,5 @@
+/** @noSelfInFile */
+
+export function __TS__Modulo(a: number, b: number): number {
```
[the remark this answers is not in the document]
YoRyan: Yes, only for 5.0 The other alternative would be to perform this transformation inline in the AST, going from `a % b` to `a - Math.floor(a / b) * b`. How difficult would this be? We'd need temporary variables to hold `a` and `b`, so we don't evaluate them twice...

src/LuaLib.ts
YoRyan: If we still need it (see above), I would agree with calling it `Modulo50`.

src/lualib/universal/CountVarargs.ts
```diff
@@ -0,0 +1,7 @@
+/** @noSelfInFile */
+
+export function __TS__CountVarargs<T>(...args: T[]): number {
```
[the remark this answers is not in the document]
YoRyan: Sadly, no, exclude doesn't work:

```json
{
    "compilerOptions": {
        "rootDirs": ["5.0", "."],
    },
    "include": ["*.ts", "5.0/*.ts", "declarations", "../../language-extensions/index.d.ts"],
    "exclude": [
        "CountVarargs.ts",
        "Error.ts",
        "Generator.ts",
        "Match.ts",
        "NumberToString.ts",
        "SourceMapTraceback.ts",
        "SparseArraySpread.ts",
        "Unpack.ts"
    ]
}
```

It breaks the 5.0 build. tsc just imports the universal version of the file, regardless of the presence of rootDirs. I think having the desired filename in the same directory is just too tempting for the compiler.

test/json.50.lua
```diff
@@ -0,0 +1,147 @@
+--
```
lolleko, line 1: Out of scope for this PR, but we should thing about, rewriting this lib in TS and compiling it for the different version before injecting into the test runner.

src/lualib/5.0/Error.ts
```diff
@@ -0,0 +1,71 @@
+interface ErrorType {
```
lolleko, line 1: Not a huge fan of the code duplication. As far as I can tell this is almost identical with the existing Error.ts, except (wrapErrorToString). Do you think it would be possible to use the existing Error.ts and work around the difference by using `_VERSION`? I think it's fine  to add an additional condition since errors are not performance critical.

src/lualib/5.0/Generator.ts
```diff
@@ -0,0 +1,28 @@
+import { __TS__CountVarargs } from "./CountVarargs";
```
lolleko, line 1: Again don't really like the code duplication. But also no real idea how to work around it, except refactoring into a common base, which might be even messier. Would be cool if we could come up with something, otherwise I am also fine with keeping it this way.

src/lualib/5.0/SourceMapTraceBack.ts
```diff
@@ -0,0 +1,69 @@
+// TODO: In the future, change this to __TS__RegisterFileInfo and provide tstl interface to
```
lolleko, line 1: (See Error.ts comment) Maybe use `_Version` in existing SourceMapTraceBack.ts to work around differences. Since it is not performance critical.

lolleko (changes requested):

src/lualib/Modulo.ts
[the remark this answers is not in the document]
Perryvw: We could do it with preceding statements, but I'm fine with keeping this a lualib module

Perryvw (approved):

src/LuaLib.ts
YoRyan: Renamed with https://github.com/TypeScriptToLua/TypeScriptToLua/pull/1263/commits/236bd774ece164145eeb334143f2ade012e91ad4.

src/lualib/5.0/Generator.ts
YoRyan: With the help of the unpack helper, it turns out we don't have to duplicate this file at all! https://github.com/TypeScriptToLua/TypeScriptToLua/pull/1263/commits/9820e42473c8b8617d62a1a5887fe4a54359423c

src/lualib/5.0/SourceMapTraceBack.ts
YoRyan: Implemented with https://github.com/TypeScriptToLua/TypeScriptToLua/pull/1263/commits/581485960ed7c053ba1e2e20a68f4dacb5726629.

src/lualib/5.0/Error.ts
YoRyan: Yes, good idea! Implemented with https://github.com/TypeScriptToLua/TypeScriptToLua/pull/1263/commits/5549a1c83e6374b1831e3dec77895c8e2b953769.

lolleko (approved):

PR #1269, opened 2022-05-07
noImplicitGlobalVariables compiler option

### LINKED ISSUE #1119: Feature request - new compiler option - noImplicitGlobalVariables
Edit - I renamed this issue; see discussion below.

In Lua land, regardless of what file you are working in, you always have to prepend your variables with `local` in order to avoid them from becoming a global variable. This is annoying, because you almost never want to create a global variable.

```lua
-- in file: foo.lua
local foo = "bar";
```

Most of the time, TSTL does not have this problem, which is great.
In TSTL land, you just declare normal variables, and they are almost always local.

```ts
// in file: foo.ts
const foo = "bar"; // Will transpile to the previous code snippet
```

However, there are two exceptions:
1) when `luaBundle` mode is turned off, any variables/functions in a file called "main.ts" or "init.ts" will automatically get turned into globals
2) when `luaBundle` mode is turned on, any variables/functions in the specified bundle entry point file will automatically get turned into globals

This is a hidden "gotcha" for TSTL users. I've been happily using TSTL for almost an entire year now, and only discovered this behavior recently. Unbeknownst to me, global variables were being created and causing run-time errors in production. ><

In a way, this behavior does make sense, because if I typed `const foo = "bar";` into a `<script>` tag on a webpage, that would obviously create a global variable. So, I think that TSTL is trying to mimic that behavior.

With that said, I think there is room for improvement here. If TSTL only creates global variables in some situations, and doesn't create global variables in other situations, this is confusing for end users. A better default might be for TSTL to just always output local variables, even in the above two situations. This helps prevents users from shooting themselves in the foot.

Consider that if an end-user is in a file called "foo.ts" and they want to create a global variable, it would look like this:

```ts
declare let foo: string;
foo = "bar";
```

Essentially, I am requesting that this global-variable-declaration-format is necessary across the board, for all TSTL files.

I think it's a good default behavior to have because using a global variable is the exception, rather than the norm. One typically only creates a global variable if they are explicitly exposing functionality to the outside world. And any particular program will only need to have a few global variables in total.

Zamiell:
Closes #1119.

Pretty self explanatory, tiny PR.

I guess I need to do website PR too?

### DIFF
--- a/test/unit/functions/noImplicitGlobalVariables.spec.ts
+++ b/test/unit/functions/noImplicitGlobalVariables.spec.ts
@@ -1,5 +1,19 @@
 import * as util from "../../util";
 
+test("normal TSTL creates global variables", () => {
+    const fooBuilder = util.testModule`
+        function foo() {}
+        const bar = 123;
+    `
+        .expectToHaveNoDiagnostics();
+
+    const transpiledFile = fooBuilder.getLuaResult().transpiledFiles[0];
+    util.assert(transpiledFile !== undefined);
+    const { lua } =  transpiledFile;
+    util.assert(lua !== undefined);
+    expect(lua).not.toContain("local");
+});
+
 test("noImplicitGlobalVariables does not create any global variables", () => {
     const fooBuilder = util.testModule`
         function foo() {}
--- a/test/unit/functions/noImplicitGlobalVariables.spec.ts
+++ b/test/unit/functions/noImplicitGlobalVariables.spec.ts
@@ -1,0 +1,17 @@
+import * as util from "../../util";
+
+test("noImplicitGlobalVariables does not create any global variables", () => {
+    const fooBuilder = util.testModule`
+        function foo() {}
+        const bar = 123;
+    `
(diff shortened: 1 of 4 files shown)


test/unit/functions/noImplicitGlobalVariables.spec.ts
```diff
@@ -0,0 +1,30 @@
+import * as util from "../../util";
+
+test("normal TSTL creates global variables", () => {
+    const builder = util.testModule`
+        function foo() {}
+        const bar = 123;
+    `.expectToHaveNoDiagnostics();
+
+    const transpiledFile = builder.getLuaResult().transpiledFiles[0];
+    util.assert(transpiledFile !== undefined);
```
Perryvw, line 10: These asserts should just be `expect.toBeDefined()`

Perryvw, line 10: These asserts should just be `expect.toBeDefined()`

PR #1270, opened 2022-05-07
Apply JSDoc Comments to Lua code

### LINKED ISSUE #815: Preserve comments
https://github.com/Microsoft/TypeScript/wiki/Architectural-Overview#trivia

Zamiell:
Fixes #815, sort-of-but-not-really.

According to [this StackOverlow page](https://stackoverflow.com/questions/47429792/is-it-possible-to-get-comments-as-nodes-in-the-ast-using-the-typescript-compiler), it is not possible to get single-line comments in the AST. However, it is possible to extract JSDoc comments that are specifically attached to functions and variables.

So, for this PR, I simply extract JSDoc where I can, and then stick it on the "leadingComments" property that Perry added in 0fba69b06767811f554f21fa8f3d20cb51d5b25c.

This works, and seems good for a first iteration of the feature.

A better solution would be to:
1) Wait for the Lua to be finished generating.
2) Grab every possible comment in the entire TS source file via the method described in the StackOverflow post.
3) Then, massage the comments into the already-generated-Lua.

However, how do we find the correct spot in the generated Lua that corresponds with this spot in the source code? Is that possible?

### DIFF
--- a/test/unit/comments.spec.ts
+++ b/test/unit/comments.spec.ts
@@ -7,7 +7,7 @@ test("JSDoc is copied on a function", () => {
          * It has multiple lines.
          */
         function foo() {}
-    `.expectToHaveNoDiagnostics();
+    `.expectToHaveNoDiagnostics().expectDiagnosticsToMatchSnapshot();
 
     const transpiledFile = builder.getLuaResult().transpiledFiles[0];
     expect(transpiledFile).toBeDefined();
@@ -23,7 +23,7 @@ test("JSDoc is copied on a variable", () => {
          * It has multiple lines.
          */
         const foo = 123;
-    `.expectToHaveNoDiagnostics();
+    `.expectToHaveNoDiagnostics().expectDiagnosticsToMatchSnapshot();
 
     const transpiledFile = builder.getLuaResult().transpiledFiles[0];
     expect(transpiledFile).toBeDefined();
--- a/test/unit/comments.spec.ts
+++ b/test/unit/comments.spec.ts
@@ -7,7 +7,9 @@ test("JSDoc is copied on a function", () => {
          * It has multiple lines.
          */
         function foo() {}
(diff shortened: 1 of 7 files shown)


test/unit/comments.spec.ts
```diff
@@ -0,0 +1,33 @@
+import * as util from "../util";
+
+test("JSDoc is copied on a function", () => {
+    const builder = util.testModule`
+        /**
+         * This is a function comment.
+         * It has multiple lines.
+         */
+        function foo() {}
+    `.expectToHaveNoDiagnostics();
```
lolleko, line 10: Can we have a snapshot test for at least one of these or a similar snippet? Usually, we try to avoid them, but in this context a snapshot test seems useful to me.

Perryvw:
Could you make this respect [the removeComments tsconfig configuration value](https://www.typescriptlang.org/tsconfig#removeComments)

test/unit/comments.spec.ts
Zamiell: im not sure how to do that, can you help?

Zamiell:
done

test/unit/comments.spec.ts
Perryvw: Simply add `.expectToMatchSnapshot()` to the test builder here

Zamiell: ok, its committed, let me know if i did it right

test/unit/comments.spec.ts
```diff
@@ -0,0 +1,37 @@
+import * as util from "../util";
+
+test("JSDoc is copied on a function", () => {
+    const builder = util.testModule`
+        /**
+         * This is a function comment.
+         * It has multiple lines.
```
Perryvw, line 7: Could you also add a test (with snapshot) for a jsdoc comment with jsdoc tags like `@param` and `@returns`

PR #1279, opened 2022-05-29
adding settings.json file

Zamiell:


Perryvw:
These are user settings and do not belong in the repo

Zamiell:
Perry, these are not per-user settings, they are specific to this TSTL repo.

Imagine that:
- I submit a pull request to TSTL, and then start working on some other project.
- One hour later, you respond to my PR with some comments. So now, I need to do a context switch, and open up the TSTL repository in VSCode again.
- At this point, I am forced to go manually change my ruler settings in my personal settings.json file to match what TSTL declares is the right setting (which happens to be 120, which is different from what Prettier recommends and is different from what the rest of the TypeScript ecosystem uses).
- After that, I make my changes to the repo in accordance with your PR comments, make a commit, and then context switch back to the project I was originally working on.
- But now I'm forced yet again manually change my ruler setting in my settings.json file back to what it was before.

This is bad DX - I expect the ruler to be per-project in the same way that I expect ESLint rules to be per-project. Imagine if you specified what the ESLint rules for the project were, but then you told your developers to put those ESLint rules in their personal settings.json file. =p

PR #1291, opened 2022-06-22
Reduce use of temp variables

GlassBricks:
This reduces the use of temp variables by using `moveToPrecedingTemp` instead of always using a temp. This was done for some cases of assigment and in destructuring declarations.

Additionally, the `this` keyword (which is read-only) is no longer moved to a temp variable.


Perryvw:
Could you give an example of a case that is not translated differently?

lolleko:
I know the validity of these changes is probably already ensured by the existing tests. But I would feel better if we had some explicit functional tests, for your optimization cases, if possible.

GlassBricks:
Sure, I just added some tests.

Issue #1262, opened 2022-04-18
Custom JSX factory incorrectly produces `nil` self parameter


andrei-markeev:
Hi,

I am trying to compile JSX with custom `jsxFactory` pointing to `createElement` function.
**TypeScriptToLua** generates extra `nil` parameter in the output:

**test_jsx.ts**:
```ts
function createElement(tag: string | Function, props: { [key: string]: string | boolean }, ...children: any[]) {
    return "whatever, just testing";
}

function testJsx() {
    return <div></div>
}
```

**test_jsx.lua**:
```lua
--[[ Generated with https://github.com/TypeScriptToLua/TypeScriptToLua ]]
function createElement()
    return "whatever"
end
function testJsx()
    return createElement(nil, "div", nil)
end
```

I would expect that it would generate this:
```lua
    return createElement("div", nil);
```

I have `noImplicitSelf` in tsconfig, and I also tried setting `/** @noSelf */` attribute, but both of these are ignored.

**tsconfig.json**:
```json
{
    "compilerOptions": {
        "target": "esnext",
        "lib": [
            "esnext"
        ],
        "jsx": "react",
        "jsxFactory": "createElement",
        "moduleResolution": "node",
        "types": ["lua-types/jit"],
        "strict": true,
        "outDir": "build"
    },
    "tstl": {
        "luaTarget": "JIT",
        "noImplicitSelf": true
    }
}
```

PR #1294, opened 2022-06-23
Fix call context

### LINKED ISSUE #1292: Function calls with this parameter break when using noImplicitSelf
```ts
function foo(this: unknown, val: unknown) {
  print(this, val)
}

foo(3)
```

In TSTL v1.6.0, when using noImplicitSelf, the `foo(3)` call gets translated to `foo(3)` when it should be translated to `foo(nil, 3)` (the this parameter is explicitly specified).

This doesn't happen in v1.5.0, so I suspect this related to the recent "fix" for noImplicitSelf.

GlassBricks:
Fixes #1292, which was broken by #1282 

To still keep #1262, I couldn't find an elegant way to detect jsx createElement calls specifically. Instead I found a condition that tells if a call is _probably_ jsx createElement (see changes in call.ts).

Maybe the best thing to do is to actually revert #1282, and document that the `this` parameter is _always_ required for jsx `createElement`.

### DIFF
--- a/src/transformation/visitors/call.ts
+++ b/src/transformation/visitors/call.ts
@@ -271,12 +271,9 @@ export const transformCallExpression: FunctionVisitor<ts.CallExpression> = (node
         );
     }
 
-    const signatureDeclaration = signature?.getDeclaration();
-
     let callPath: lua.Expression;
     let parameters: lua.Expression[];
-    const isContextualCall =
-        !signatureDeclaration || getDeclarationContextType(context, signatureDeclaration) !== ContextType.Void;
+    const isContextualCall = isContextualCallExpression(context, node, signature);
     if (!isContextualCall) {
         [callPath, parameters] = transformCallAndArguments(context, calledExpression, node.arguments, signature);
     } else {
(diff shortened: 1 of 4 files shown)


src/transformation/visitors/call.ts
```diff
@@ -300,6 +296,23 @@ export const transformCallExpression: FunctionVisitor<ts.CallExpression> = (node
...
+    context: TransformationContext,
+    node: ts.CallExpression,
+    signature: ts.Signature | undefined
+): boolean {
+    const declaration = signature?.getDeclaration();
+    if (!declaration) {
+        if (ts.nodeIsSynthesized(node.expression)) {
+            // this includes JSX createElement calls
+            // For this we use noImplicitSelf option
+            return !context.options.noImplicitSelf;
```
Perryvw, line 309: Can't we just use this return for any node without declaration? In what case does this fail?

GlassBricks: I just checked and it works, will change.

I somehow recall that it broke some tests, I guess that is (no longer?) the case.

PR #1303, opened 2022-07-12
Add LuaMap and LuaSet language extensions

GlassBricks:
Adds `LuaMap` and `LuaSet` language extensions. Readonly variants are also declared. They somewhat mimic ES maps and sets.

`LuaMap` is like a table, except that the `get` method may return `nil`. This more closely matches the ES map interface.

For `LuaSet`, two language extensions are added:
- `LuaTableAdd` which translates calls of `table.add(foo)` to `table[foo] = true`
- `LuaPairsKeyIterable`, which is like `LuaPairsIterable` except only the key value is used (and multi-return is not used).


src/transformation/visitors/language-extensions/table.ts
```diff
@@ -172,3 +181,29 @@ function transformTableSetExpression(
...
+    // arg0[arg1] = true
+    const [table, value] = transformExpressionList(context, args);
+    context.addPrecedingStatements(
+        lua.createAssignmentStatement(
+            lua.createTableIndexExpression(table, value),
+            lua.createBooleanLiteral(true),
+            node
+        )
+    );
+    return lua.createNilLiteral();
```
Perryvw, line 208: Any thoughts on returning `table` instead? This is how sets work in TS:
```ts
const myset = new Set<number>().add(1).add(2).add(3);
```
Though I guess that could get us in an annoying spot where we have to worry about side effects evaluating `table`, so I'd be fine leaving this as it is.

test/unit/language-extensions/table.spec.ts
```diff
@@ -332,6 +332,16 @@ describe("LuaTable extension interface", () => {
             .expectToEqual({ baz: 5 });
     });
 
+    test("table add", () => {
```
Perryvw, line 335: Missing test for LuaTableAdd, this only tests LuaTableAddMethod.

Perryvw (changes requested):

language-extensions/index.d.ts
```diff
@@ -495,11 +502,28 @@ declare type LuaTableSet<TTable extends AnyTable, TKey extends AnyNotNil, TValue
...
     LuaExtension<"__luaTableSetMethodBrand">;
 
+/**
+ * Calls to functions with this type are translated to `table[key] = true`.
+ * For more information see: https://typescripttolua.github.io/docs/advanced/language-extensions
+ *
+ * @param TTable The type to access as a Lua table.
+ * @param TKey The type of the key to use to access the table.
+ */
+declare type LuaTableAdd<TTable extends AnyTable, TKey extends AnyNotNil> = ((table: TTable, key: TKey) => void) &
```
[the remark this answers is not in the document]
GlassBricks: I've renamed it to `LuaTableAddKey`.

src/transformation/visitors/language-extensions/table.ts
GlassBricks: This could be possible to implement, however it would be inconsistent with existing lua table extension functions.
Perhaps this could be done in another PR.

test/unit/language-extensions/table.spec.ts
GlassBricks: I added more tests similar to existing lua table extensions tests.

PR #1308, opened 2022-07-16
Significant performance improvements

GlassBricks:
When benchmarking with isaccscript-common using cli, nearly halves the total time, and improves transpilation time (time spent in `transformSourceFile` by a factor of 5.

The most improvement was from changing checking for language-extensions from O(total number of extensions) to O(1) (which is checked at _every_ identifier and call expression!) To do this, language extensions were separated into separate categories (`__tstlExtension`, `__tstlIterable`, and `__tstlMultiReturn`), and extension type is checked by reading the string literal value of the extension property (except for `__tstlMultiReturn`).

This rewrites/re-structures language extension calls and builtin calls, and cleans up some code/bugs found along the way.

Annotations that were deprecated have now been completely removed from being checked.


Zamiell:
Amazing!!!

src/transformation/visitors/class/index.ts
```diff
@@ -220,13 +207,14 @@ function transformClassLikeDeclaration(
...
 
-    superInfo.pop();
+    context.classSuperInfos.pop();
 
     return { statements: result, name: className };
 }
 
 export const transformSuperExpression: FunctionVisitor<ts.SuperExpression> = (expression, context) => {
-    const superInfos = getOrUpdate(classSuperInfos, context, () => []);
+    // const superInfos = getOrUpdate(classSuperInfos, context, () => []);
```
Perryvw, line 216: Remove  commented code

src/transformation/visitors/language-extensions/multi.ts
```diff
@@ -85,29 +88,9 @@ export function shouldMultiReturnCallBeWrapped(context: TransformationContext, n
     }
 
     // LuaIterable in for...of
-    if (ts.isForOfStatement(node.parent) && isIterableExpression(context, node)) {
+    if (ts.isForOfStatement(node.parent) && getIterableExtensionKindForNode(context, node) === "Iterable") {
```
Perryvw, line 91: Should be comparing with the enum value here, not with a string literal

src/transformation/visitors/language-extensions/table.ts
```diff
@@ -114,41 +49,40 @@ function transformTableDeleteExpression(
...
-        args.unshift(node.expression.expression);
-    }
+    const args =
+        extensionKind === ExtensionKind.TableGetMethodType
+            ? transformWithTableArgument(context, node)
+            : transformExpressionList(context, node.arguments);
 
-    const [table, accessExpression] = transformExpressionList(context, args);
-    // arg0[arg1]
+    const [table, accessExpression] = args; // arg0[arg1]
```
Perryvw, line 70: Comment is misplaced now, it applies to the next line

src/transformation/visitors/loops/for-of.ts
```diff
@@ -49,19 +47,22 @@ export const transformForOfStatement: FunctionVisitor<ts.ForOfStatement> = (node
...
-        return transformForOfIterableStatement(context, node, body);
-    } else if (isPairsIterableExpression(context, node.expression)) {
-        return transformForOfPairsIterableStatement(context, node, body);
-    } else if (isPairsKeyIterableExpression(context, node.expression)) {
-        return transformForOfPairsKeyIterableStatement(context, node, body);
-    } else if (isLuaIteratorType(context, node.expression)) {
-        context.diagnostics.push(annotationRemoved(node.expression, AnnotationKind.LuaIterator));
-    } else if (isArrayType(context, context.checker.getTypeAtLocation(node.expression))) {
+    }
+    const iterableType = getIterableExtensionKindForNode(context, node.expression);
```
Perryvw, line 51:
```suggestion
    const iterableExtensionType = getIterableExtensionKindForNode(context, node.expression);
```

src/transformation/visitors/loops/for-of.ts
```diff
@@ -49,19 +47,22 @@ export const transformForOfStatement: FunctionVisitor<ts.ForOfStatement> = (node
...
+    if (iterableType) {
+        if (iterableType === "Iterable") {
```
Perryvw, line 53: Compare to the enum member instead of string literal

src/transformation/visitors/return.ts
```diff
@@ -93,7 +93,8 @@ export function createReturnStatement(
 
     if (isInTryCatch(context)) {
         // Bubble up explicit return flag and check if we're inside a try/catch block
-        results.unshift(lua.createBooleanLiteral(true));
+        // results.unshift(lua.createBooleanLiteral(true));
+        results = [lua.createBooleanLiteral(true), ...results];
```
Perryvw, line 97: I'm not sure this is really an improvement, commented code should be removed either way

src/utils.ts
```diff
@@ -8,6 +8,16 @@ export function castArray<T>(value: T | readonly T[]): readonly T[] {
     return Array.isArray(value) ? value : [value];
 }
 
+export function castArrayFlatMap<T, TResult>(
```
Perryvw, line 11: As commented before, I don't see the value in this function

Perryvw (changes requested):
Would have preferred if this had come in smaller incremental PRs. There are a lot of different improvement ideas mixed together now which makes it much harder to review...

I reviewed most of this but skipped some parts that seemed not finished yet. Also resolve needs a further very careful look.

language-extensions/index.d.ts
```diff
@@ -17,15 +29,17 @@ declare type LuaExtension<TBrand extends string> = { [T in TBrand]: { readonly _
...
 
 /**
  * Represents multiple return values as a tuple.
  * For more information see: https://typescripttolua.github.io/docs/advanced/language-extensions
  *
  * @param T A tuple type with each element type representing a return value's type.
  */
-declare type LuaMultiReturn<T extends any[]> = T & LuaExtension<"__luaMultiReturnBrand">;
+declare type LuaMultiReturn<T extends any[]> = T & {
+    readonly __tstlMultiReturn: any;
```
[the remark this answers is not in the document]
GlassBricks: The `LuaIterable` type is both MultiReturn and LuaIterable, which breaks the assumption that one type can only be one extension (this broke multiple tests). Making multireturn its own property fixed this.

src/transformation/builtins/string.ts
```diff
@@ -145,6 +145,9 @@ export function transformStringPrototypeCall(
             return transformLuaLibFunction(context, LuaLibFeature.StringPadStart, node, caller, ...params);
         case "padEnd":
             return transformLuaLibFunction(context, LuaLibFeature.StringPadEnd, node, caller, ...params);
+        case "toString":
+            const toStringIdentifier = lua.createIdentifier("tostring");
+            return lua.createCallExpression(toStringIdentifier, [caller, ...params]);
```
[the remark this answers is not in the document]
GlassBricks: Without this, there would be and "unsupported property" diagnostic. I suppose returning to caller could work here.

src/transformation/context/context.ts
```diff
@@ -49,80 +53,92 @@ export class TransformationContext {
...
 
     public superTransformStatements(node: StatementLikeNode | readonly StatementLikeNode[]): lua.Statement[] {
-        return castArray(node).flatMap(n => {
+        return castArrayFlatMap(node, n => {
             this.pushPrecedingStatements();
             const statements = this.superTransformNode(n) as lua.Statement[];
-            statements.unshift(...this.popPrecedingStatements());
-            return statements;
+            const result = this.popPrecedingStatements();
+            result.push(...statements);
```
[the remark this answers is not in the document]
GlassBricks: Push is much more performant than unshift, and this is an very commonly called method.

src/transformation/utils/language-extensions.ts
```diff
@@ -3,142 +3,111 @@ import { TransformationContext } from "../context";
...
-    );
+export function getExtensionKindForSymbol(
+    context: TransformationContext,
+    symbol: ts.Symbol
+): ExtensionKind | undefined {
+    const type = context.checker.getTypeOfSymbolAtLocation(symbol, context.sourceFile);
+    return getExtensionKindForType(context, type);
+}
+
+export enum IterableExtensionKind {
```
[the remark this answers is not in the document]
GlassBricks: See comment on MultiReturn, it's to avoid the possibility of a type having multiple extensions in different "categories".

src/transformation/context/context.ts
```diff
@@ -49,80 +53,92 @@ export class TransformationContext {
...
-        }
-
-        return result as lua.Expression;
+        const result = this.doSuperTransformNode(node);
+        return this.assertIsExpression(node, result);
     }
 
     public transformStatements(node: StatementLikeNode | readonly StatementLikeNode[]): lua.Statement[] {
-        return castArray(node).flatMap(n => {
+        return castArrayFlatMap(node, n => {
```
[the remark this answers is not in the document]
GlassBricks: I just did a remeasurment; as this method is commonly called, this change provide a consistently measurable (30ms over a 5500ms time) speedup. If the slight clarity is still prefered over the slight improvement, this could be reverted.

src/transformation/utils/annotations.ts
```diff
@@ -1,153 +1,116 @@
...
+                kind: tag.name as AnnotationKind,
+                args: tag.text?.map(p => p.text) ?? [],
+            };
+            annotationsMap.set(tagName, annotation);
         }
     }
 }
 
+interface SymbolWithAnnotationMap extends ts.Symbol {
+    tstlAnnotationMap?: AnnotationsMap;
```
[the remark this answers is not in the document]
GlassBricks: This is how typescript caches computations internally too; using a weak map instead would add a measurable overhead.

src/transformation/utils/language-extensions.ts
```diff
@@ -3,142 +3,111 @@ import { TransformationContext } from "../context";
...
+    TableGetType = "TableGet",
+    TableGetMethodType = "TableGetMethod",
+    TableHasType = "TableHas",
+    TableHasMethodType = "TableHasMethod",
+    TableSetType = "TableSet",
+    TableSetMethodType = "TableSetMethod",
+    TableAddKeyType = "TableAddKey",
+    TableAddKeyMethodType = "TableAddKeyMethod",
 }
+const extensionNames: Set<string> = new Set(Object.values(ExtensionKind));
```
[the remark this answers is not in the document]
GlassBricks: I'm not sure I understand what you mean; This explicitly a set of values, not keys (the enum keys and values are no longer the same).

language-extensions/index.d.ts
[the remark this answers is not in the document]
GlassBricks: Without this, the `LuaIterable` type broke because it is both a "MultiReturn" and a "LuaIterable", breaking the assumption a type can have at most one extensions. Putting extensions in multiple "categories", and making MultiReturn its own category, worked around this.

src/transformation/context/context.ts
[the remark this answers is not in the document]
Perryvw: I don't think this extra code is worth for 0.4% performance increase, please revert

src/transformation/visitors/identifier.ts
```diff
@@ -3,95 +3,114 @@ import * as lua from "../../LuaAST";
...
+        context.diagnostics.push(invalidRangeUse(identifier));
+    } else if (extensionKind === ExtensionKind.VarargConstant) {
+        context.diagnostics.push(invalidVarargUse(identifier));
     }
+}
 
-    if (isOperatorMapping(context, identifier)) {
-        context.diagnostics.push(invalidOperatorMappingUse(identifier));
-    }
+const extensionKindToValueName: { [T in ExtensionKind]?: string } = {
```
Perryvw, line 34: This should probably be located somewhere in transformation/language-extensions and imported

src/transformation/visitors/language-extensions/table.ts
```diff
@@ -1,109 +1,44 @@
...
-        extensionType === extensions.ExtensionKind.TableAddType ||
-        extensionType === extensions.ExtensionKind.TableAddMethodType
-    ) {
-        return transformTableAddExpression(context, node, extensionType);
-    }
+    return getExtensionKindForNode(context, node.expression) === ExtensionKind.TableNewType;
 }
+export const tableNewExtensions = [ExtensionKind.TableNewType];
+
+export const tableExtensionTransformers: { [P in ExtensionKind]?: LanguageExtensionCallTransformer } = {
```
Perryvw, line 13: This might also be nicer to locate somewhere in the language-extensions dir and then import

src/transformation/visitors/return.ts
Perryvw: Is allocating a new array and copying really faster than unshift?

src/transformation/visitors/return.ts
```diff
@@ -102,7 +102,9 @@ export function createReturnStatement(
 function isInTryCatch(context: TransformationContext): boolean {
     // Check if context is in a try or catch
     let insideTryCatch = false;
-    for (const scope of walkScopesUp(context)) {
```
Perryvw, line 105: I preferred this helper method, prevents a lot of error-prone mess, surely that was not that bad performance-wise

src/tstl.ts
```diff
@@ -95,20 +96,27 @@ function performCompilation(
     options: tstl.CompilerOptions,
     configFileParsingDiagnostics?: readonly ts.Diagnostic[]
 ): void {
+    if (options.measurePerformance) performance.enable();
```
Perryvw, line 99: I'd propose either renaming `performance` to `measurePerformance` (in all files), *or* to rename `enable` to `enableMeasurement` (and `disableMeasurement`). `performance.enable()` looks more like we're enabling some kind of performance mode, it's not clear.

src/transpilation/resolve.ts
```diff
@@ -34,206 +36,184 @@ class ResolutionContext {
...
+    private getFileFromPath(resolvedPath: string): string | undefined {
+        const existingFile = this.pathToFile.get(resolvedPath);
+        if (existingFile) return existingFile;
+        if (existingFile === false) return undefined;
+
+        const file = searchForFile(this);
+        this.pathToFile.set(resolvedPath, file ?? false);
+        return file;
+
+        function searchForFile(context: ResolutionContext) {
```
Perryvw, line 172: I don't really get the point of making this a local function that is called just once, why not just keep this function code inlined like it was. Also why do you pass in `this` if you could just capture from context?

src/transpilation/resolve.ts
```diff
@@ -34,206 +36,184 @@ class ResolutionContext {
...
     }
+}
 
-    // Not a TS file in our project sources, use resolver to check if we can find dependency
-    try {
-        const resolveResult = resolver.resolveSync({}, fileDirectory, dependency);
-        if (resolveResult) {
-            return resolveResult;
+export function resolveDependencies(program: ts.Program, files: ProcessedFile[], emitHost: EmitHost): ResolutionResult {
+    // const outFiles: ProcessedFileMap = new Map(files.map(file => [file.fileName, file]));
```
Perryvw, line 203: Commented out code, should be removed

Perryvw (changes requested):
Why is the benchmark failing, mismatch of language extensions between master and your branch?

GlassBricks:
> Why is the benchmark failing, mismatch of language extensions between master and your branch?

The benchmark is using an old version of language-extensions, so MultiReturn is not being detected. This is due to the lua-types package depending on a older version of tstl (for type definitions, not compilation). A somewhat hacky workaround I found was to add `"typescript-to-lua": "file:."` to devDependencies.

src/transformation/visitors/language-extensions/table.ts
GlassBricks: Isn't this already in the language-extensions dir?

src/transpilation/resolve.ts
GlassBricks: The function was helpful because you can `return` out of it early. I moved it to a member function.

src/transformation/context/context.ts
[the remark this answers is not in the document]
GlassBricks: Reverted.

src/transformation/utils/annotations.ts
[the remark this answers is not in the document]
GlassBricks: Performance degradation is slight but not too significant; a map has been used instead.

src/transformation/utils/function-context.ts
```diff
@@ -99,67 +109,63 @@ export function getDeclarationContextType(
...
         }
+    }
 
-        return signatureDeclaration;
-    });
+    return [signatureDeclaration];
+}
+
+interface TypeWithFunctionContext extends ts.Type {
+    tstlFunctionContextType?: ContextType;
```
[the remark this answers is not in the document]
Perryvw: Not yet resolved, I don't want to mix TS and TSTL types.

src/transformation/utils/function-context.ts
```diff
@@ -29,15 +29,28 @@ function hasNoSelfAncestor(declaration: ts.Declaration): boolean {
...
-        param => ts.isIdentifier(param.name) && param.name.originalKeywordKind === ts.SyntaxKind.ThisKeyword
-    );
+    const param = signatureDeclaration.parameters[0];
+    if (param && ts.isIdentifier(param.name) && param.name.originalKeywordKind === ts.SyntaxKind.ThisKeyword) {
+        return param;
+    }
+}
+
+interface SignatureDeclarationWithContextType extends ts.SignatureDeclarationBase {
+    tstlFunctionContextType?: ContextType;
```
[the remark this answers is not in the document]
Perryvw: Not yet resolved, I don't want to mix TS and TSTL types.

src/transformation/visitors/language-extensions/table.ts
Perryvw: Not yet resolved

src/transformation/visitors/return.ts
```diff
@@ -93,10 +91,10 @@ export function createReturnStatement(
 
     if (isInTryCatch(context)) {
         // Bubble up explicit return flag and check if we're inside a try/catch block
-        results.unshift(lua.createBooleanLiteral(true));
+        values.unshift(lua.createBooleanLiteral(true));
```
Perryvw, line 94: Oh I see now why I was confused before, my bad. We should not be modifying input parameters, please revert.

Perryvw (changes requested):

src/transformation/visitors/language-extensions/table.ts
GlassBricks: I somehow managed to fix every other comment in the file except that one.
Fixed now.

PR #1314, opened 2022-07-20
Add ObjectConstructor method overrides for LuaPairsIterable types

GlassBricks:
This allows users to use code such as `Object.values(luaMap)` with correct types.

Possible future related changes:
 - Update website documentation to mention this
 - Update the type for `pairs` in the `lua-types` package


.eslintrc.js
```diff
@@ -226,6 +226,7 @@ module.exports = {
             files: "language-extensions/index.d.ts",
             rules: {
                 "@typescript-eslint/naming-convention": "off",
+                "@typescript-eslint/array-type": "off",
```
Perryvw, line 229: Please revert this and just fix the location where it complains

PR #1344, opened 2022-09-03
Update VS Code development container configuration

YoRyan:
Bump to Node 18, delete the obsolete shell preference, fix npm error with the post-install command, and run as a non-root user.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -7,6 +7,6 @@
         "editorconfig.editorconfig",
         "esbenp.prettier-vscode"
     ],
-    "postCreateCommand": "npm ci --legacy-peer-deps",
+    "postCreateCommand": "npm ci",
     "remoteUser": "node"
 }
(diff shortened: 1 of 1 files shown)


.devcontainer/devcontainer.json
```diff
@@ -1,14 +1,12 @@
...
-    },
+    "image": "mcr.microsoft.com/vscode/devcontainers/javascript-node:18",
     "extensions": [
         "typescript-to-lua.vscode-typescript-to-lua",
         "dbaeumer.vscode-eslint",
         "editorconfig.editorconfig",
         "esbenp.prettier-vscode"
     ],
-    "postCreateCommand": "npm ci"
+    "postCreateCommand": "npm ci --legacy-peer-deps",
```
Perryvw, line 10: legacy-peer-deps should not be in here, this is something I'm looking to fix, not something we want to keep

YoRyan: Ok, no problem to remove it. The post-create command failing in the meantime is a very minor inconvenience.

PR #1370, opened 2022-11-25
5.0: Emulate % with native math.mod()

YoRyan:
It's come to my attention that this is a much better way of doing modulo.

Depends on https://github.com/TypeScriptToLua/lua-types/pull/78.

If you're wondering why the universal addition is there, our code still requires all targets to build the Modulo50 feature, so I've inserted a non-functional dummy.

### DIFF
--- a/src/lualib/universal/Modulo50.ts
+++ b/src/lualib/universal/Modulo50.ts
@@ -1,0 +1,5 @@
+/** @noSelfInFile */
+
+export function __TS__Modulo50(_a: number, _b: number) {
+    throw "Using a 5.0 shim for a non-5.0 target, what?";
+}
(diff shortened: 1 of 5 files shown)


src/lualib/universal/Modulo50.ts
```diff
@@ -0,0 +1,5 @@
+/** @noSelfInFile */
+
+export function __TS__Modulo50(_a: number, _b: number) {
+    throw "Using a 5.0 shim for a non-5.0 target, what?";
```
Perryvw, line 4: This file probably just should not exist

YoRyan: Done, should have thought of it first.

PR #1372, opened 2022-12-24
Add async iterator symbol to Symbol table.

aermicioi:
Added missing async iterator symbol, preventing proper implementation of custom async iterators.


Perryvw:
Please add some tests demonstrating implementation of an async iterator here: https://github.com/TypeScriptToLua/TypeScriptToLua/blob/master/test/unit/loops.spec.ts#L430

PR #1381, opened 2023-01-13
Simplify optional chaining expressions where possible

GlassBricks:
This PR simplifies/reduces the lua output of optional chains, in the most common cases.

- Uses an `and` statement (`a and a.b`) instead of an if statement where possible.
- Reuses identifiers on the left expression when possible

### DIFF
--- a/src/transformation/visitors/optional-chaining.ts
+++ b/src/transformation/visitors/optional-chaining.ts
@@ -1,10 +1,12 @@
 import * as ts from "typescript";
 import * as lua from "../../LuaAST";
-import { TransformationContext, tempSymbolId } from "../context";
+import { tempSymbolId, TransformationContext } from "../context";
 import { assert, assertNever } from "../../utils";
 import { transformInPrecedingStatementScope } from "../utils/preceding-statements";
-import { transformPropertyAccessExpressionWithCapture, transformElementAccessExpressionWithCapture } from "./access";
+import { transformElementAccessExpressionWithCapture, transformPropertyAccessExpressionWithCapture } from "./access";
 import { shouldMoveToTemp } from "./expression-list";
+import { expressionResultIsUsed } from "../utils/typescript";
+import { wrapInStatement } from "./expression-statement";
 
 type NormalOptionalChain = ts.PropertyAccessChain | ts.ElementAccessChain | ts.CallChain;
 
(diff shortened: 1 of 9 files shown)


src/transformation/visitors/optional-chaining.ts
```diff
@@ -90,16 +97,16 @@ export function transformOptionalChain(context: TransformationContext, node: ts.
 
 export function transformOptionalChainWithCapture(
     context: TransformationContext,
-    node: ts.OptionalChain,
+    tsNode: ts.OptionalChain,
     thisValueCapture: lua.Identifier | undefined,
     isDelete?: ts.DeleteExpression
 ): ExpressionWithThisValue {
-    const luaTemp = context.createTempNameForNode(node);
+    const luaTempName = context.createTempName("opt");
```
Perryvw, line 104: Why change this from what it was?

src/transformation/visitors/optional-chaining.ts
```diff
@@ -121,26 +128,27 @@ export function transformOptionalChainWithCapture(
...
     });
 
     // transform left expression, handle thisValue if needed by rightExpression
     const thisValueCaptureName = context.createTempName("this");
     const leftThisValueTemp = lua.createIdentifier(thisValueCaptureName, undefined, tempSymbolId);
     let capturedThisValue: lua.Expression | undefined;
 
-    const rightContextualCall = getOptionalContinuationData(tsTemp)?.contextualCall;
+    const optionalContinuationData = getOptionalContinuationData(tsTemp);
+    const rightContextualCall = optionalContinuationData?.contextualCall;
```
Perryvw, line 151: I don't understand the point of this change

src/transformation/visitors/optional-chaining.ts
GlassBricks: "node" was too generic and so confusing; there are a lot of variables, TS and Lua, in the body.

src/transformation/visitors/optional-chaining.ts
GlassBricks: optionalContinuationData is used later (line 193).

src/transformation/visitors/optional-chaining.ts
Perryvw: It wasn't actually "node" literal though, it was already deducing a very specific name based on what the node is, see the changes in the snapshots where it changed `____table_has_result_0` to `___opt_0`

GlassBricks: Sorry, earlier I thought the comment was addressing something else.

The reason for this is that long-ish optional chains, which are somewhat common, generate long names such as "____a_b_c_foo_result_c_1". "____opt" is shorter and still signifies what the original TS was.

Issue #1389, opened 2023-02-02
Support 'exports' field in package.json of published modules.


pilaoda:
There are two problems about this field.

First:
https://github.com/TypeScriptToLua/TypeScriptToLua/blob/93ff05d082340ccfd77942a602cac0c3e96fd67f/src/transpilation/resolve.ts#L13-L19
The resolver should add conditions to support export conditions in a published module's package.json like:
```
{
    "exports": {
        "./client": {
            "require": "./dist/client/index.lua",
            "types": "./src/index.ts"
        }
    },
}
```
I guess we should add `["node", "require"]`, and maybe `"tstl"` ?
Related doc: 
 - https://nodejs.org/api/packages.html#conditional-exports
 - https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/#esm-nodejs

Second:
https://github.com/TypeScriptToLua/TypeScriptToLua/blob/93ff05d082340ccfd77942a602cac0c3e96fd67f/src/transpilation/utils.ts#L54
`resolvePlugin` use `node-resolve` rather than `enhanced-resolve`, which not support this "exports" field.

PR #1392, opened 2023-02-04
Support 'exports' field in package.json of published modules.

pilaoda:
#1389 Fix problem 1


test/transpile/module-resolution.spec.ts
```diff
@@ -272,6 +273,49 @@ describe("module resolution project with dependencies built by tstl library mode
...
 
+describe("module resolution project with dependencies built by tstl library mode and has exports field", () => {
+    const projectPath = path.resolve(__dirname, "module-resolution", "project-with-tstl-library-has-exports-field");
+    const appPath = path.join(projectPath, "app");
+
+    // First compile dependencies into node_modules. NOTE: Actually writing to disk, very slow
+    tstl.transpileProject(path.join(projectPath, "dependency1-ts", "tsconfig.json"));
+
+    // Install dependencies. This will create node_modules folder with dependency1-ts in it.
+    execSync("npm i", { cwd: appPath });
```
Perryvw, line 284: I don't want to run npm during tests for all kind of reasons (dependency, security, flakyness etc). Please just put dependency1 in node_modules beforehand and compile there, or commit a symlink or something like that.

test/transpile/module-resolution/project-with-tstl-library-has-exports-field/app/package-lock.json
```diff
@@ -0,0 +1,28 @@
+{
```
Perryvw, line 1: Could this file be removed now that we no longer do a npm install?

PR #1401, opened 2023-02-13
Set lualib tsconfig to strict

Perryvw:
Enabled strict TS in lualib source so it can be fully checked and understood by TS, as well as eslint

As added bonus the extra types enabled tstl to optimize some lualib functions


src/lualib/tsconfig.lua50.json
```diff
@@ -7,14 +7,13 @@
...
+        "strict": true
     },
     "tstl": {
         "luaLibImport": "none",
         "noHeader": true,
         "luaTarget": "5.0",
         "luaPlugins": [{ "name": "../../dist/lualib-build/plugin.js" }]
     },
-    "include": ["*.ts", "5.0/*.ts", "declarations", "../../language-extensions/index.d.ts"]
+    "include": ["*.ts", "5.0/**/*.ts", "declarations/**/*.ts", "../../language-extensions/index.d.ts"]
```
Zamiell, line 18: can you simplify this to `"./**/.*.ts"`

Perryvw: no because I expicitly want ot include 5.0 and not universal

PR #1407, opened 2023-02-25
Fix crash missing arguments

### LINKED ISSUE #1406: Exception thrown when  calling string.indexOf without arguments
Offending code:
```
"".indexOf()
```

Perryvw:
Fixes #1406


hazzard993 (approved):
Looks good! Just some added caution

PR #1408, opened 2023-02-25
Fix exception when unwrapping parenthesised multireturn calls

### LINKED ISSUE #1404: Generation failed when using LuaMultiReturn in type assertion
see [playground example](https://typescripttolua.github.io/play/#code/DYUwLgBAZgrgdgYwFwQIZwJ4CgsApaJoDOEuYAFgJZEoBuA9pQCYA0EAdJ6gE4DmNaTAEoIAXgB8EADIxUAWRjAwlAErgY3OAB4A2kTDdKcXm32HjAXXFDcARjYAmNgGYhQA)

The code below will cause an error in generation.
```typescript
let func: any
let [a, b] = (func as (this: void, ...args: any) => LuaMultiReturn<[string, string]>)(1, 2, 3)
```

Perryvw:
Fixes #1404

### DIFF
--- a/test/unit/language-extensions/multi.spec.ts
+++ b/test/unit/language-extensions/multi.spec.ts
@@ -329,7 +329,7 @@ test("return LuaMultiReturn from catch", () => {
 });
 
 // https://github.com/TypeScriptToLua/TypeScriptToLua/issues/1404
-test("LuaMultiReturn applies after casting a functtion (#1404)", () => {
+test("LuaMultiReturn applies after casting a function (#1404)", () => {
     util.testFunction`
         let swap: any = (a: number, b: number) => $multi(b, a);
         let [a, b] = (swap as (...args: any) => LuaMultiReturn<[number, number]>)(4, 3);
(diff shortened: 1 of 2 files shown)


test/unit/language-extensions/multi.spec.ts
```diff
@@ -327,3 +327,14 @@ test("return LuaMultiReturn from catch", () => {
         .withLanguageExtensions()
         .expectToEqual(2);
 });
+
+// https://github.com/TypeScriptToLua/TypeScriptToLua/issues/1404
+test("LuaMultiReturn applies after casting a functtion (#1404)", () => {
```
hazzard993, line 332: functtion

hazzard993 (changes requested):

test/unit/language-extensions/multi.spec.ts
hazzard993, line 332:
```suggestion
test("LuaMultiReturn applies after casting a function (#1404)", () => {
```

I should use this

hazzard993 (approved):

PR #1409, opened 2023-02-25
Spread LuaPairsIterable and LuaPairsKeysIterable

### LINKED ISSUE #1384: LuaSet spread iterable operation leads to runtime error
This probably also applies to LuaMap:

```ts
const a: LuaSet<number> = new LuaSet();

a.add(10);
a.add(20);

[...a].map(a => 10);
```

https://typescripttolua.github.io/play/#code/MYewdgzgLgBAhgLhgGQK5wMoFMoB4yoC2ARlgE4B8MAvDGFgO4rrZQAUAlANwBQPcAOjgATYWwCMABm78hotgCZpvHgG0BGuAF0BhOAAc2cGlSncgA

Perryvw:
Fixes #1384

### DIFF
--- a/test/unit/spread.spec.ts
+++ b/test/unit/spread.spec.ts
@@ -524,8 +524,8 @@ test.each(["LuaTable", "LuaMap"])("can spread %s with pairs (#1244)", type => {
 
     // We don't know the order so match like this
     expect(result).toHaveLength(2);
-    expect(result.some(([k, v]) => k === "foo" && v === "bar")).toBe(true);
-    expect(result.some(([k, v]) => k === "fizz" && v === "buzz")).toBe(true);
+    expect(result).toContainEqual(["foo", "bar"]);
+    expect(result).toContainEqual(["fizz", "buzz"]);
 });
 
 // https://github.com/TypeScriptToLua/TypeScriptToLua/issues/1384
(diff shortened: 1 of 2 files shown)


test/unit/spread.spec.ts
```diff
@@ -527,3 +527,37 @@ test.each(["LuaTable", "LuaMap"])("can spread %s (#1244)", type => {
...
+        tbl.set("fizz", "buzz");
+        return [...tbl];
+    `
+        .withLanguageExtensions()
+        .getLuaExecutionResult();
+
+    // We don't know the order so match like this
+    expect(result).toHaveLength(2);
+    expect(result.some(([k, v]) => k === "foo" && v === "bar")).toBe(true);
+    expect(result.some(([k, v]) => k === "fizz" && v === "buzz")).toBe(true);
```
hazzard993, line 545: This matcher might help here, doesn't care about the order of the outer-most array but DOES care about the order of `["foo", "bar"]`

```suggestion
    expect(result).toContainEqual(["foo", "bar"]);
    expect(result).toContainEqual(["fizz", "buzz"]);
```

Perryvw: Thanks, didn't know about this one. That is much nicer indeed.

PR #1410, opened 2023-02-26
Fix element access exception caused by invalid TS

Perryvw:


test/unit/destructuring.spec.ts
```diff
@@ -225,3 +226,14 @@ describe("array destructuring optimization", () => {
             .expectToMatchJsResult();
     });
 });
+
+test("no exception from invalid TS syntax", () => {
```
hazzard993, line 230: We do consider these "sequence expressions" (`a, b` execute `a`, then use `b` as a result) valid syntax and the code below is just `let [a, b] = testFunc(5)[a, b] = testFunc(b)`

This test is more "can use sequence expressions inside an element access expression without fatal errors" which isn't related to destructuring

Don't want to cover all potential invalid TS code!

Perryvw: We want to at least not throw  as long as a valid AST is passed into tstl. I guess I should not call it invalid syntax because the syntax itself is valid, it's just that there are also semantic errors reported by TS, in which case we will not guarantee we produce correct output, but there should be no exception.

PR #1413, opened 2023-02-28
Implement Number.toFixed

Dimava:
Needs some throughts on if `Number.toFixed` has to follow JS rounding behaviour vs having simple implementation and possibly being more fast
Otherwise seems to work fine

### DIFF
--- a/src/lualib/NumberToFixed.ts
+++ b/src/lualib/NumberToFixed.ts
@@ -1,0 +1,14 @@
+/// https://www.ecma-international.org/ecma-262/10.0/index.html#sec-number.prototype.tofixed
+export function __TS__NumberToFixed(this: number, fractionDigits?: number): string {
+    if (Math.abs(this) >= 1e21 || this !== this) {
+        return this.toString();
+    }
+    const f = Math.floor(fractionDigits ?? 0);
+    // reduced to 99 as string.format only supports 2-digit numbers
+    if (f < 0 || f > 99) {
+        throw "toFixed() digits argument must be between 0 and 99";
+    }
+    // throws "invalid format (width or precision too long)" if strlen > 99
+    // if (f < 80) return fmt; else try {return fmt} catch(_) { throw "toFixed() digits argument..." }
+    return string.format(`%.${f}f`, this);
+}
--- a/test/unit/builtins/numbers.spec.ts
+++ b/test/unit/builtins/numbers.spec.ts
@@ -69,6 +69,26 @@ test.each([
     util.testExpressionTemplate`(${value}).toString(2)`.expectToEqual(luaNativeSpecialNumString);
 });
 
+const toFixedFractions = [undefined, 0, 1, 2, Math.PI, 5, 99];
+// 1.5, 1.25 and 1.125 fails as rounding differ
+const toFixedValues = [-1, 0, 1, Math.PI, -1.1234, -9.99e19, 1e22];
+const toFixedPairs = toFixedValues.flatMap(value => toFixedFractions.map(frac => [value, frac] as const));
+test.each(toFixedPairs)("(%p).toFixed(%p)", (value, frac) => {
+    util.testExpressionTemplate`(${value}).toFixed(${frac})`.expectToMatchJsResult();
+});
+
+test.each([
+    [NaN, "(0/0)"],
+    [Infinity, "(1/0)"],
+    [-Infinity, "(-(1/0))"],
+])("%p.toFixed(2)", (value, luaNativeSpecialNum) => {
+    // Need to get the actual lua tostring version of inf/nan
+    // this is platform dependent so we can/should not hardcode it
+    const luaNativeSpecialNumString = util.testExpression`${luaNativeSpecialNum}.toString()`.getLuaExecutionResult();
+    // Cannot use expectToMatchJsResult because this actually wont be the same in JS in Lua
+    util.testExpressionTemplate`(${value}).toFixed(2)`.expectToEqual(luaNativeSpecialNumString);
+});
+
 test.each(cases)("isNaN(%p)", value => {
     util.testExpressionTemplate`isNaN(${value} as any)`.expectToMatchJsResult();
 });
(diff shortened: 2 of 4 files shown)


src/lualib/NumberToFixed.ts
```diff
@@ -0,0 +1,13 @@
+/// https://www.ecma-international.org/ecma-262/10.0/index.html#sec-number.prototype.tofixed
+export function __TS__NumberToFixed(this: number, fractionDigits?: number): string {
+    if (Math.abs(this) >= 1e21 || this !== this) {
```
Perryvw, line 3: I'm guessing this `this !== this` is to handle `NaN`? If so, please also add a test for `NaN`

test/unit/builtins/numbers.spec.ts
```diff
@@ -69,6 +69,13 @@ test.each([
     util.testExpressionTemplate`(${value}).toString(2)`.expectToEqual(luaNativeSpecialNumString);
 });
 
+const toFixedRadixes = [undefined, 0, 1, 2, 5, 10, 80];
+const toFixedValues = [-1, 0, 1, /* 1.5 fails, */ 1024, 1.2, 1e19, 1e22];
```
Perryvw, line 73: Why does 1.5 fail?

test/unit/builtins/numbers.spec.ts
```diff
@@ -69,6 +69,13 @@ test.each([
     util.testExpressionTemplate`(${value}).toString(2)`.expectToEqual(luaNativeSpecialNumString);
 });
 
+const toFixedRadixes = [undefined, 0, 1, 2, 5, 10, 80];
```
Perryvw, line 72: Would be nice to add fractional radixes here since we support it with Math.floor

test/unit/builtins/numbers.spec.ts
Perryvw, line 73: Please add a value with 6+ decimal places here

Perryvw (changes requested):
I'm not sure what the 'simple' behavior vs the rounding behavior is, but this looks alright to me

test/unit/builtins/numbers.spec.ts
Dimava: Because JS is rounding it up to `2` but the used Lua implementation is rounding it down to `1`

Possibly https://stackoverflow.com/questions/64562060/why-does-formatting-a-number-round-differently-in-luajit-vs-lua

Dimava: > ```
> // Differences in math implementations between JS & Lua cause inaccuracies
> // therefore test input numbers are "carefully" selected to always match accuratly.

src/lualib/NumberToFixed.ts
```diff
@@ -0,0 +1,13 @@
...
+        return this.toString();
+    }
+    const f = Math.floor(fractionDigits ?? 0);
+    if (f < 0 || f > 100) {
```
Dimava, line 7: Actually it may make sence to limit the fraction on ~80 to avoid "invalid format (width or precision too long)"

test/unit/builtins/numbers.spec.ts
Dimava: Added a number and Math.PI

test/unit/builtins/numbers.spec.ts
```diff
@@ -69,6 +69,45 @@ test.each([
...
+    -34910.3,
+    1.45,
+    1.55,
+    1.444555,
+    1024,
+    1.2,
+    1e19,
+    1e22,
+];
+const toFixedPairs = toFixedValues.flatMap(value => toFixedFractions.map(frac => [value, frac] as const));
```
Perryvw, line 93: 9 * 16 seems a bit excessive, lots of these tests seem to be slight duplicates of each other, i,e if you just have 1.444555 and 1.2 in here you can probably remove the other fractionals

test/unit/builtins/numbers.spec.ts
```diff
@@ -69,6 +69,45 @@ test.each([
...
+test.each([
+    [NaN, "(0/0)"],
+    [Infinity, "(1/0)"],
+    [-Infinity, "(-(1/0))"],
+])("%p.toFixed(2)", (value, luaNativeSpecialNum) => {
+    // Need to get the actual lua tostring version of inf/nan
+    // this is platform dependent so we can/should not hardcode it
+    const luaNativeSpecialNumString = util.testExpression`${luaNativeSpecialNum}.toString()`.getLuaExecutionResult();
+    // Cannot use expectToMatchJsResult because this actually wont be the same in JS in Lua
+    // TODO fix this in lualib/NumberToString.ts
```
Perryvw, line 107: Either make this change or remove this TODO

test/unit/builtins/numbers.spec.ts
Dimava: 1.444555 I was trying to get another bad rounding error, but it seems only `N+0.5` can reliably cause those

src/lualib/NumberToFixed.ts
Perryvw: I don't really care, even leaving out the check would be an option. Do what you think is best.

src/lualib/NumberToFixed.ts
```diff
@@ -0,0 +1,14 @@
+/// https://www.ecma-international.org/ecma-262/10.0/index.html#sec-number.prototype.tofixed
+export function __TS__NumberToFixed(this: number, fractionDigits?: number): string {
+    if (Math.abs(this) >= 1e21 || this !== this) {
+        return this.toString();
+    }
+    const f = Math.floor(fractionDigits ?? 0);
+    // reduced to 99 as string.format only supports 2-digit numbers
+    if (f < 0 || f > 99) {
+        throw "toFixed() digits argument must be between 0 and 99";
```
Dimava, line 9: Should errors be exaclty like in JS (`between 0 and 100`) or this is fine?

Dimava:
> I'm not sure what the 'simple' behavior vs the rounding behavior is, but this looks alright to me

Nevermind, I've missed that __TS__NumberToString uses `string.format` on `radix === 10` instead of manual rounding

Perryvw (approved):

PR #1424, opened 2023-03-27
Don't output parameter initializer statement if default value is nil

### LINKED ISSUE #1422: [TSTL transformation] Default parameters with 'null' and 'undefined' get compiled
## Problem:

When assign default values for property parameters and assign 'null' or 'undefined' value, resulting code generate instructions that do not make sense.

## Example:

source:
image

result:
image

## What is expected:

Probably no code at all as result since it does assign + check, but logically it is invalid

### LINKED ISSUE #1423: Wrong output with preceding statements in function parameter initializer
Example:

```typescript
let i = 1
function foo(x = i++){
    return x
}
foo()
```
This puts the preceding statements for `i++` outside of the function.

GlassBricks:
Closes #1422
Also fixes #1423

### DIFF
--- a/src/transformation/visitors/function.ts
+++ b/src/transformation/visitors/function.ts
@@ -27,11 +27,10 @@ function transformParameterDefaultValueDeclaration(
     value: ts.Expression,
     tsOriginal?: ts.Node
 ): lua.Statement | undefined {
-    // const parameterValue = context.transformExpression(value);
-    const { precedingStatements, result: parameterValue } = transformInPrecedingStatementScope(context, () =>
-        context.transformExpression(value)
+    const { precedingStatements: statements, result: parameterValue } = transformInPrecedingStatementScope(
+        context,
+        () => context.transformExpression(value)
     );
-    const statements = [...precedingStatements];
     if (!lua.isNilLiteral(parameterValue)) {
         statements.push(lua.createAssignmentStatement(parameterName, parameterValue));
     }
--- a/src/transformation/visitors/function.ts
+++ b/src/transformation/visitors/function.ts
@@ -27,11 +27,15 @@ function transformParameterDefaultValueDeclaration(
     value: ts.Expression,
     tsOriginal?: ts.Node
 ): lua.Statement | undefined {
-    const parameterValue = context.transformExpression(value);
-    if (lua.isNilLiteral(parameterValue)) {
-        return undefined;
+    // const parameterValue = context.transformExpression(value);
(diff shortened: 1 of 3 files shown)


src/transformation/visitors/function.ts
```diff
@@ -24,19 +24,26 @@ import { transformBindingPattern } from "./variable-declaration";
...
-): lua.Statement {
-    const parameterValue = value ? context.transformExpression(value) : undefined;
-    const assignment = lua.createAssignmentStatement(parameterName, parameterValue);
+): lua.Statement | undefined {
+    // const parameterValue = context.transformExpression(value);
+    const { precedingStatements, result: parameterValue } = transformInPrecedingStatementScope(context, () =>
+        context.transformExpression(value)
+    );
+    const statements = [...precedingStatements];
+    if (!lua.isNilLiteral(parameterValue)) {
```
Perryvw, line 35: Couldn't we do that at the top of the function? Check ts.isUndefinedLiteral || ts.isNullLiteral? That way we can avoid a whole bunch of transformations.

test/unit/functions/functions.spec.ts
```diff
@@ -107,6 +107,48 @@ test("Function default binding parameter maintains order", () => {
...
+            return x ?? "nil";
+        }
+        return [a.get("foo"), foo(), a.get("foo")];
+    `
+        .withLanguageExtensions()
+        .tap(builder => {
+            const lua = builder.getMainLuaCodeChunk();
+            expect(lua).not.toMatch("    x = nil");
+        })
+        .expectToEqual(["bar", "nil", "baz"]);
```
Perryvw, line 149: Would prefer if you could just use `expectToMatchJsResult` here, using default TS (no language extensions) with an object and helper function for example.

Perryvw (changes requested):

src/transformation/visitors/function.ts
GlassBricks: Doing this makes this the most general; it applies to anything that results in a nil literal (casts/parenthesis/etc, language extensions, plugin output) and shouldn't have any performance impact.

test/unit/functions/functions.spec.ts
GlassBricks: Could you give a more detailed example what that would that look like?
Besides using language extensions, I can't think of a good method that transpiles to nil (not just returns undefined) that works both in JS and lua.

Perryvw: Oh I see what you are testing now, that seems maybe a bit too specific but okay

PR #1428, opened 2023-04-10
Fixed multi returning all values

### LINKED ISSUE #1411: LuaMultiReturn returns all values even when indexed with [0]
See [Playground sample](https://typescripttolua.github.io/play/#code/GYVwdgxgLglg9mABMOcAUUAWMDOAuRANzhgBMAaRHKAJwOppjAHMBKe2p5xAbwChEgxDQCmUEDSRoGXAHTMcIAEbTalAESz1GgFTrWAbQAMAXVYBuPgF8+oSLASIlAQxoZs+IiQpVaHRizsvgHc-EKIEAjUiAaiigA2UCaIALzBcgrKqjQaWrr6luGi4pLCIglQljZ20PBILgBe7rgExGSUDP5cQTIsvAJCkWDRcSCJqeks8ooqDLnaiOp6hqaFQsUSSKOJVXwADgFQaCjo6tj6FvuHaC5u6rcXlgdMR41o9yANDY98Q9EAjiAAB4TE7vQHAgpAA).

I expected that `foo` would have transpiled to (notice parentheses):
```lua
function foo(str)
    return (string.gsub(str, ".", "*"))
end
```
But, it clearly doesn't and therefore leads to multiple returns, because `string.gsub` also returns the amount of replacements as a secondary return value, which invalidates the specified return type `: string` of the `foo`... It works fine in case of `bar` and `baz` since they make use of a variable.

lolleko:
We try to avoid using `select()` when we use `[0]` to access the first value of multiple returns (e.g. `multiCall()[0]`). In that case, we  omit the `select()`. Unfortunately, this only works if the value is assigned directly (`local x = multiCall()`). In cases where the value is not assigned, omitting the select and keeping just the call expression (`multiCall()`) will lead to all return values being passed on. (See #1411)

To fix this, we now wrap the call in parentheses using a new Lua parentheses expression. Using parentheses is the correct way to keep only the first value of multiple returns (see https://www.lua.org/manual/5.1/manual.html#2.5)

TL;DR: `multiCall()[0]` used to transpile to `multiCall()` now it transpiles to `(multiCall())` which ensures correct semantics across all potential usages.

This fixes #1411

### DIFF
--- a/src/LuaAST.ts
+++ b/src/LuaAST.ts
@@ -44,6 +44,7 @@ export enum SyntaxKind {
     MethodCallExpression,
     Identifier,
     TableIndexExpression,
+    ParenthesizedExpression,
 
     // Operators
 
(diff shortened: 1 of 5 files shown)


Perryvw (approved):
Using parentheses is a nice tidy way to do this!

PR #1440, opened 2023-05-04
support extends "tstl" object for tsconfig.json in node_modules

pilaoda:
#1436 doesn't handle such case:
```
{
    "extends": "my-npm-module/tsconfig.dev.json",
}
```
This PR fix this with ts internal API.

### DIFF
--- a/src/cli/tsconfig.ts
+++ b/src/cli/tsconfig.ts
@@ -58,10 +58,6 @@ export function parseConfigFileWithSystem(
     return updateParsedConfigFile(parsedConfigFile);
 }
 
-function pathIsRelative(path: string): boolean {
-    return /^\.\.?($|[\\/])/.test(path);
-}
-
 function resolveModuleConfig(
     moduleName: string,
     configRootDir: string,
@@ -79,9 +75,9 @@ function getExtendedTstlOptions(
     cycleCache: Set<string>,
     system: ts.System
 ): TypeScriptToLuaOptions {
-    const absolutePath = path.isAbsolute(configFilePath)
+    const absolutePath = ts.pathIsAbsolute(configFilePath)
         ? configFilePath
-        : pathIsRelative(configFilePath)
+        : ts.pathIsRelative(configFilePath)
         ? path.resolve(configRootDir, configFilePath)
         : resolveModuleConfig(configFilePath, configRootDir, system);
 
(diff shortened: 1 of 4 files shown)


src/cli/tsconfig.ts
```diff
@@ -58,13 +58,37 @@ export function parseConfigFileWithSystem(
     return updateParsedConfigFile(parsedConfigFile);
 }
 
+function pathIsRelative(path: string): boolean {
```
Perryvw, line 61: Could you add a comment on this function explaining why we need it and how it's different from path.absolute? Took a lot of mental effort to understand the need for this, and I'm sure I will have forgotten again next week!

Perryvw (changes requested):

src/cli/tsconfig.ts
pilaoda: First condition `path.isAbsolute` is for absolute path like `D:/my-project/tsconfig.json`,
Second condition `pathIsRelative` is for relative path like `./folder/tsconfig.json`,
Final is the npm module path like `@pilaoda/ts-config/tsconfig.dev.json`

`pathIsRelative` is simply copy from typescript sourcecode https://github.com/microsoft/TypeScript/blob/c333e14578984d0071da63fc28147f1e6fc43ae0/src/compiler/path.ts#LL95C1-L102C2
to keep consistent with its extends behavior

pilaoda: I found ts just export it but mark internal, so I add declaration for them and use both `ts.pathIsAbsolute` and `ts.pathIsRelative`.

src/cli/tsconfig.ts
```diff
@@ -58,13 +58,33 @@ export function parseConfigFileWithSystem(
...
     configRootDir: string,
     cycleCache: Set<string>,
     system: ts.System
 ): TypeScriptToLuaOptions {
-    const absolutePath = path.isAbsolute(configFilePath) ? configFilePath : path.resolve(configRootDir, configFilePath);
+    const absolutePath = ts.pathIsAbsolute(configFilePath)
+        ? configFilePath
+        : ts.pathIsRelative(configFilePath)
+        ? path.resolve(configRootDir, configFilePath)
+        : resolveModuleConfig(configFilePath, configRootDir, system);
```
Perryvw, line 82: I would still like you to add a comment in code explaining that if a path is not absolute and not relative that it is then a module path that needs to be resolved

Perryvw (approved):

PR #1441, opened 2023-05-11
Fix different function (address) return on use bind, on object method

websharik:
That importanf for friendly subscribe/unsubscribe callbacks.
```ts
class A {
  start() {
    subscribe(this.doSome.bind(this))
  }
  stop() {
    unsubscribe(this.doSome.bind(this))
  }
  doSome() {}
}
```

How its works - on binding method also wrap to arrow function, but then once save to `__TS__wrappedMethods` in binding context. On next bind same method with same content, result getting from __TS__wrappedMethods.

Any solution to make `.bind(this)` is auto to make code more comfortable and shortly ?


websharik:
Now it working different instance of V8 (original JS/TS runtime) and brings inconvenience.

PR #1442, opened 2023-05-12
Add "luaLibImport: require-minimal" option value under TSConfig schema

jacob1237:
According to the [docs](https://typescripttolua.github.io/docs/configuration#custom-options), it's allowed to set the following enum values for this option:

- `none`
- `inline`
- `require`
- `require-minimal`

However, the current TSConfig schema doesn't support `require-minimal`, making the VSCode complain about this value:
require-minimal-error

The fix attempts to resolve this problem.

### DIFF
--- a/tsconfig-schema.json
+++ b/tsconfig-schema.json
@@ -44,7 +44,7 @@
                     "description": "Specifies how js standard features missing in lua are imported.",
                     "type": "string",
                     "default": "require",
-                    "enum": ["none", "inline", "require"]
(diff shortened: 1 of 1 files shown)


Zamiell:
require-minimal was recently added, so we just forgot to update the schema, thanks!

Perryvw:
Actually I intentionally omitted this from the schema because require-minimal is still an experimental feature. I have not seen or heard any bugs related to it, so I think it would be alright to add it now

PR #1449, opened 2023-06-03
Ts 5.0 decorators

Perryvw:
Added support for new TS5.0 decorators.

Field decorators are only partially supported for now, the value initializers are currently ignored. If you try to use a value initializer, a warning will be shown.

### DIFF
--- a/src/LuaLib.ts
+++ b/src/LuaLib.ts
@@ -37,6 +37,7 @@ export enum LuaLibFeature {
     ClassExtends = "ClassExtends",
     CloneDescriptor = "CloneDescriptor",
     CountVarargs = "CountVarargs",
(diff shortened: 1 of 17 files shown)


Zamiell:
Can we have an overview of any new features?
Or is this PR more of an internal-only reorganization to comply with the new spec?

Perryvw:
> Can we have an overview of any new features? Or is this PR more of an internal-only reorganization to comply with the new spec?

It's support for the new TypeScript 5.0 decorators, you should probably just look at their overview: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-0.html#decorators

And I also found this article they linked to useful:
https://2ality.com/2022/10/javascript-decorators.html

PR #1455, opened 2023-06-06
Add import resolution plugin hook

SamuelRobertBond:
I was having issues with files with a . in the file name (e.g physics-api.component.ts) and needed some method of providing custom filepath resolution login when importing files from from a library into another project bundling the library.


SamuelRobertBond:
I found some issues with paths in this, currently working through it. Let me know if you have any thoughts on what I can add or change.

SamuelRobertBond:
Fixed the issue and moved the plugin hook call to `resolveLuaDependencyPathFromNodeModules`. I uploaded an example usage [here](https://github.com/SamuelRobertBond/tstl-import-resolution-plugin-example)

SamuelRobertBond:
Another idea I had for changes to the plugins would be to allow plugins definitions to be passed through the compiler options when using tstl through the api. It would be a seperate issue, but I wanted to get your thoughts on this before I start messing around with that.

This is an example of what im talking about, although I may just not be understanding the `TransformerImport`
image

src/transpilation/plugins.ts
```diff
@@ -45,6 +45,8 @@ export interface Plugin {
         emitHost: EmitHost,
         result: EmitFile[]
     ) => ts.Diagnostic[] | void;
+
+    onImportResolutionFailure?: (packageRoot : string, dependency : string) => string | undefined
```
Perryvw, line 49: A module resolution plugin is something we have long been considering, and I think a better interface would be something like:

```ts
interface ModuleResolutionRequest {
    moduleIdentifier: string; // "foo.bar.baz" from require("foo.bar.baz")
    requiringFile: string; // path of the (lua) file containing the require
    // maybe some other stuff I'm forgetting now?
    // maybe a 'defaultResolve: (ModuleResolutionRequest) => string' to call the default behavior?
    // Not sure why anyone would want that
}

moduleResolution?: (moduleToBeResolved: ModuleResolutionRequest) => string | undefined
```

Then crucially this plugin would be executed _before_ our regular module resolution process, allowing users to intercept before it even starts. If it returns undefined it will go into the default behavior. I'm just wondering if we want to somehow maybe also be able to indicate that a specific import should _not_ be emitted, but maybe that's for a future version.

Perryvw, line 49: I don't like you passing packageRoot to the outside world, that's really something internal to the tstl implementation. What is the reason you need this in your case? I'm trying to understand the usage scenario

src/transpilation/resolve.ts
```diff
@@ -184,17 +186,35 @@ class ResolutionContext {
...
             if (fileFromPath) {
                 return fileFromPath;
             } else {
                 // Did not find file at current root, try again one directory deeper
-                packageRoot = path.join(packageRoot, splitPath[packageRootIndex++]);
+                currentPackage = path.join(currentPackage, splitPath[packageRootIndex++]);
+            }
+        }
+
+        if (this.plugins != null) {
```
Perryvw, line 204: I don't think this should be exclusive to nodeModules paths, I think a nicer place for this would be in `resolveImport`, before the call to `resolveDependencyPath` (going with the idea I mentioned on the plugin interface above, where users can intercept these calls before we do all of this quite complicated work on the file system while we already know we won't find anything)

src/transpilation/resolve.ts
```diff
@@ -235,6 +255,11 @@ class ResolutionContext {
             path.join(resolvedPath, "index.lua"), // lua index file in sources
             path.join(resolvedPath, "init.lua"), // lua looks for <require>/init.lua if it cannot find <require>.lua
         ];
+
+        if (resolvedPath.endsWith(".lua")) {
```
Perryvw, line 259: I don't think this if statement should be here at all. If the resolved path ends with .lua we do not have to search for it, because we apparently we already have a lua file path. I don't think this is possible in regular tstl module resolution, but I guess your plugin triggers this? I don't think this code path should really be possible ever.

SamuelRobertBond: I put this here as a convince thing for me, since the initial problem I intend to solve was being able to specify. When I was constructing the path it I included .lua in the resolved path. This was not being found since the possible files above appends .lua to the resolve path, the path was being set to /my/path/file.lua.lua. I didn't want to just add resolvedPath itself to the possible files, because someone could intentionally provide a path to a non lua file. I think the changes you mentioned above would make this a non issue though.

src/transpilation/plugins.ts
SamuelRobertBond: In my particular case, I was getting the path the package and the attempted path to the dependency and transforming that to what I needed it to be, so when the plugin. Before I made the changes to packageRoot, the path being passed to the plugin was inconsistent. This was just much more consistent for me to work with. I think the changes to the interface and location for the plugins call will resolve this issue.

SamuelRobertBond: I think this interface makes more sense, it gives more control to someone writing a plugin to identify the package they wish to use via the module identifier. Ill start working on it.

SamuelRobertBond:
I made some of the changes requested. I didnt include `defaultResolve` yet. I also still need to write the test. Figured I'd get them out there in case you wanted to take a look. Ill probably write the test tomorrow.

Perryvw (commented):
Starting to look good, need a test still, and of course you should make sure this works for your use-case too still. Also to avoid our formatter complaining, you can run `npm run lint` and `npm run fix:prettier` locally to make sure there are no formatting or linting complaints.

SamuelRobertBond:
Fixed an issue I found with the path resolution and added the test specified. I ran and fixed the formatting with the linter and ran the entire test suite with everything passing. I wasn't sure if I should add a test in the plugins spec as well. I included the moduleResolution plugin for the test in the plugins folder, but I wasn't sure if that was the appropriate place for it. Let me know if there's anything else I should include in the test or anything that needs to be changed.

src/transpilation/resolve.ts
```diff
@@ -93,6 +98,60 @@ class ResolutionContext {
...
 
+    private resolveDependencyPathsWithPlugins(required: ProcessedFile, dependency: string) {
+        const requiredFromLuaFile = required.fileName.endsWith(".lua");
+        const dependencyPath = requiredFromLuaFile ? luaRequireToPath(dependency) : dependency;
+
+        for (const plugin of this.plugins) {
+            if (plugin.moduleResolution != null) {
+                const pluginResolvedPath = plugin.moduleResolution(
+                    dependency,
+                    dependencyPath,
```
Perryvw, line 109: this parameter is the requiringFile, shouldn't this be `required.fileName`?

src/transpilation/resolve.ts
```diff
@@ -93,6 +98,60 @@ class ResolutionContext {
...
+                    const isRelative = ["/", "./", "../"].some(p => pluginResolvedPath.startsWith(p));
+
+                    // // If the import is relative, always resolve it relative to the requiring file
+                    // // If the import is not relative, resolve it relative to options.baseUrl if it is set
+                    const fileDirectory = path.dirname(required.fileName);
+                    const relativeTo = isRelative ? fileDirectory : this.options.baseUrl ?? fileDirectory;
+
+                    // // Check if file is a file in the project
+                    const resolvedPath = path.join(relativeTo, pluginResolvedPath);
+                    const fileFromPath = this.getFileFromPath(resolvedPath);
```
Perryvw, line 140: This looks just straight up duplicate from logic in `resolveDependencyPath`, please move to a shared function they both can use

test/transpile/plugins/moduleResolution.ts
```diff
@@ -0,0 +1,12 @@
+import type * as tstl from "../../../src";
+
+const plugin: tstl.Plugin = {
+    moduleResolution(moduleIdentifier, requiringFile) {
+        if (moduleIdentifier.includes("foo")) {
+            return requiringFile.replace("foo", "bar");
```
Perryvw, line 6: the requiring file is called main.ts (or .lua?), not foo, so I'm confused by this plugin. See also my other comment regarding the requirintFile parameter

src/transpilation/resolve.ts
SamuelRobertBond: I missed that, which make sense why I ran into the issues I did, I'll fix that ty

Perryvw (approved):
Looks good to me. Thanks for your contribution!

PR #1472, opened 2023-08-05
fix: add check to see in ownerSymbol parent is the global symbol

starsprung:
when determining when to transform global builtin calls.


src/transformation/builtins/index.ts
```diff
@@ -81,7 +81,8 @@ function tryTransformBuiltinGlobalMethodCall(
 ) {
     const ownerType = context.checker.getTypeAtLocation(calledMethod.expression);
     const ownerSymbol = tryGetStandardLibrarySymbolOfType(context, ownerType);
-    if (!ownerSymbol || ownerSymbol.parent) return;
```
Perryvw, line 84: Wondering if instead we shouldn't just be removing this `ownerSymbol.parent`, not really sure if it adds anything

PR #1473, opened 2023-08-05
Fix excessive sourcemap traceback import in bundle mode

Cassolette:
### Environment
- `luaBundle{Entry}` enabled
- `sourceMapTraceback` enabled

### Before:

`__TS__SourceMapTraceBack` being required in every module even when not being used.

```lua
____modules = {
["lualib_bundle"] = function(...) 
local function __TS__SourceMapTraceBack(fileName, sourceMap)
    ...
end
 end,
["mod1"] = function(...) 
local ____lualib = require("lualib_bundle")
local __TS__SourceMapTraceBack = ____lualib.__TS__SourceMapTraceBack
 end,
["mod2"] = function(...) 
local ____lualib = require("lualib_bundle")
local __TS__SourceMapTraceBack = ____lualib.__TS__SourceMapTraceBack
 end,
}

local __TS__SourceMapTraceBack = require("lualib_bundle").__TS__SourceMapTraceBack
__TS__SourceMapTraceBack(debug.getinfo(1).short_src, {...})
```

### After:
```lua
____modules = {
["lualib_bundle"] = function(...) 
local function __TS__SourceMapTraceBack(fileName, sourceMap)
    ...
end
 end,
["mod1"] = function(...) 
 end,
["mod2"] = function(...) 
 end,
}

local __TS__SourceMapTraceBack = require("lualib_bundle").__TS__SourceMapTraceBack
__TS__SourceMapTraceBack(debug.getinfo(1).short_src, {...})
```

### DIFF
--- a/src/LuaPrinter.ts
+++ b/src/LuaPrinter.ts
@@ -173,7 +173,8 @@ export class LuaPrinter {
 
     public print(file: lua.File): PrintResult {
         // Add traceback lualib if sourcemap traceback option is enabled
-        if (this.options.sourceMapTraceback) {
+        // Not required locally when in bundle mode
+        if (this.options.sourceMapTraceback && !isBundleEnabled(this.options)) {
             file.luaLibFeatures.add(LuaLibFeature.SourceMapTraceBack);
         }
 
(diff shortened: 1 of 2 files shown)


Perryvw:
What is the benefit of this change?

Cassolette:
There is no real functional change, it just reduces the number of useless traceback imports when transpiling in bundle mode.

This change just makes it so we include `__TS__SourceMapTraceBack` feature at a global level and not at the module level which causes those imports (although not used) to be present in all modules

Perryvw:
I don't like bundling logic leaking into individual file transformation and the bundled files no longer being valid by themself. Without any real benefit to doing this, I don't really feel like merging this.

PR #1475, opened 2023-08-24
Add support for Glob-Patterns in "noResolvePaths"

### LINKED ISSUE #1161: Allow noResolvePaths to be a pattern
In some environments (e.g. factorio mods) you want entire libraries to have no resolution (and just use typescript's resolution instead). It would be nice if we could specify something such as `noResolvePaths: ["someLib/*"]` to no-resolve all imports that starts with `someLib`.

Curve:
Add support for glob patterns in `noResolvePaths` using [picomatch](https://github.com/micromatch/picomatch ).

Also supplies `0` as the duration for the CachedInputFileSystem as updated typing requires so.  
This change should not change any caching behavior as it explicitly causes the current behavior to be used (see [here](https://github.com/webpack/enhanced-resolve/blob/3a28f47788de794d9da4d1702a3a583d8422cd48/lib/CachedInputFileSystem.js#L458))

Fixes #1161.

### DIFF
--- a/src/transpilation/resolve.ts
+++ b/src/transpilation/resolve.ts
@@ -10,11 +10,12 @@ import { couldNotReadDependency, couldNotResolveRequire } from "./diagnostics";
 import { BuildMode, CompilerOptions } from "../CompilerOptions";
 import { findLuaRequires, LuaRequire } from "./find-lua-requires";
 import { Plugin } from "./plugins";
+import * as picomatch from "picomatch";
 
 const resolver = resolve.ResolverFactory.createResolver({
     extensions: [".lua"],
     enforceExtension: true, // Resolved file must be a lua file
-    fileSystem: { ...new resolve.CachedInputFileSystem(fs) },
+    fileSystem: { ...new resolve.CachedInputFileSystem(fs, 0) },
     useSyncFileSystemCalls: true,
     conditionNames: ["require", "node", "tstl", "default"],
     symlinks: false, // Do not resolve symlinks to their original paths (that breaks node_modules detection)
@@ -26,7 +27,7 @@ interface ResolutionResult {
 }
 
 class ResolutionContext {
-    private noResolvePaths: Set<string>;
+    private noResolvePaths: picomatch.Matcher[];
 
     public diagnostics: ts.Diagnostic[] = [];
     public resolvedFiles = new Map<string, ProcessedFile>();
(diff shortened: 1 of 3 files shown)


Curve:
Could we maybe have a new release after this is merged? - We'd desperately need this for a project ^^

src/transpilation/resolve.ts
```diff
@@ -10,11 +10,12 @@ import { couldNotReadDependency, couldNotResolveRequire } from "./diagnostics";
 import { BuildMode, CompilerOptions } from "../CompilerOptions";
 import { findLuaRequires, LuaRequire } from "./find-lua-requires";
 import { Plugin } from "./plugins";
+import * as picomatch from "picomatch";
 
 const resolver = resolve.ResolverFactory.createResolver({
     extensions: [".lua"],
     enforceExtension: true, // Resolved file must be a lua file
-    fileSystem: { ...new resolve.CachedInputFileSystem(fs) },
+    fileSystem: { ...new resolve.CachedInputFileSystem(fs, 0) },
```
Perryvw, line 18: What does this do?

Perryvw (changes requested):
Looks good, but before merging this I would like to have some tests added here: https://github.com/TypeScriptToLua/TypeScriptToLua/blob/master/test/transpile/module-resolution.spec.ts#L444

The test can simply check if there are no diagnostics for non-existing noResolvePaths (with a glob pattern), and that existing files are still resolved.

Something along these lines:
```ts
    test("can ignore everything with glob pattern", () => {
        util.testModule`
            import * as a from "a";
            import * as bc from "b/c";
            import * as d from "./d";
            import * as ef from "./e/f";

            import 
        `
        .addExtraFile("a.d.ts", "export function foo1(): void;")
        .addExtraFile("b/c.d.ts", "export function foo2(): void;")
        .addExtraFile("d.d.ts", "export function foo3(): void;")
        .addExtraFile("e/f.d.ts", "export function foo4(): void;")
        .setOptions({ noResolvePaths: ["*"] })
        .expectToHaveNoDiagnostics();
    });

    test("can ignore specific files with glob pattern", () => {
        util.testModule`
            import * as a from "ignoreme1";
            import * as b from "ignoreme2";

            import * as c from "actualfile";

            export const result = c.foo();
        `
        .addExtraFile("a.d.ts", "export function foo1(): void;")
        .addExtraFile("b.d.ts", "export function foo2(): void;")
        .addExtraFile("actualfile.ts", "export function foo() { return 'foo'; }")
        .setOptions({ noResolvePaths: ["ignoreme*"] })
        .expectToHaveNoDiagnostics()
        .expectToEqual({ result: "foo" });
    });
```

(I haven't tested these to you might have to tweak them a little)

src/transpilation/resolve.ts
Curve: The new types for enhanced-resolver require a number to be specified, when using `0` we ensure that it behaves as it does currently (see the linked code in my PR message)

Curve:
I've now added 3 tests

* Test if `noResolvePaths` works with simple-glob pattern 
  * This is a modified version of your second suggested test case, which also includes a lua file to preload the missing file as to not error out at runtime
* Test if `noResolvePaths` works with `require()`
* Test if `noResolvePaths` works within a lua file

Perryvw (approved):

PR #1483, opened 2023-09-04
Auto publish language-extensions

lolleko:
When. a release of the main package is created via `npm version`, the subpackage `language-extensions` will get bumped to the same version. Both will then be published to npm via GitHub Actions

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -21,3 +21,5 @@ jobs:
       - run: npm publish
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_AUTH_TOKEN }}
+      - run: cd language-extensions
+      - run: npm publish
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
         "fix:prettier": "prettier --write .",
         "check:language-extensions": "tsc --strict language-extensions/index.d.ts",
         "preversion": "npm run build && npm test",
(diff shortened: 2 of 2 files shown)


Perryvw:
I see that in the tstl package.json we have a dependency on `"@typescript-to-lua/language-extensions": "1.0.0"`. I guess we should also be bumping that version to match the new version.

I am worried that in that case the build order becomes very important. We could not build (npm install) the version bump commit because the packages are not yet available...

lolleko:
> I see that in the tstl package.json we have a dependency on `"@typescript-to-lua/language-extensions": "1.0.0"`. I guess we should also be bumping that version to match the new version.
> 
> I am worried that in that case the build order becomes very important. We could not build (npm install) the version bump commit because the packages are not yet available...

Good catch. I will think a bit about on how to solve this. AFAIK this dependency is only there to "force" users to download language-extensions alongside TSTL. My initial thoughts and possible solutions:

1. Remove the dependency and let users install language-extensions manually if needed (not sure if that's good or not)
2. Change the build order and do something like `extension bump` -> `publish extensions` -> `tstl bump` -> `tstl publish`. However, this will be difficult with our current setup as we do the bumping locally and releasing via action.
3. Don't automate just do it manually as you did just now and use the script from the DotA monorepo.
To ensure we don't forgot about releasing an language-extension update in the future we could write a GitHub action that warns us if language-extensions files have been modified without being released.

Issue #1487, opened 2023-09-22
`nil,` is being prefixed to functions


Z3rio:
`nil` seems to be automatically being added as a parameter to functions where it shouldn't be added.
![image](https://github.com/TypeScriptToLua/TypeScriptToLua/assets/54480523/4de89461-9422-4e3b-8118-ccbeab59c588)
![image](https://github.com/TypeScriptToLua/TypeScriptToLua/assets/54480523/99e50dca-f7b2-4385-82f5-b72f3a691385)

I've read about this perhaps being the `self`/`this` reference, if so, how do I get rid of it? Since these functions break due to this fact.

Zamiell:
did you read [this page](https://typescripttolua.github.io/docs/the-self-parameter) yet?

Z3rio:
> did you read [this page](https://typescripttolua.github.io/docs/the-self-parameter) yet?

That does not seem to apply to functions outside of your own project :)

Z3rio:
I have been able to get around this by modifying TSTL's code and adding my own checks, although I do not really believe that is a great solution, so I would still highly appreciate suggestions.

Zamiell:
> Yeah, but that does not seem to apply to functions outside of your own project

are you talking about upstream lua? are you typing your own definitions?

Z3rio:
> > Yeah, but that does not seem to apply to functions outside of your own project
> 
> are you talking about upstream lua? are you typing your own definitions?

I'm making add-ons for a game which supports Lua and JS.
Whereas I need the files to be in Lua, but would really love to still be able to use TS.

So the functions I'm using are related to that game and defined in these npm packages:
https://www.npmjs.com/package/@citizenfx/client
https://www.npmjs.com/package/@citizenfx/server

Hence why the functions aren't defined in my local project files and most likely why it doesn't work.

Zamiell:
if the game supports JS then why are you using TSTL at all? can't you just use typescript and use the normal TS compiler to transpile to JS?

Z3rio:
> Whereas I need the files to be in Lua, but would really love to still be able to use TS.
From my previous message ^^


The game requires Lua for quite a lot of functionalities, especially for publishing your projects, hence why I would need to use Lua :) I already use Ts -> JS where it is possible.

Zamiell:
can you post a playground that illustrates the problem?

Z3rio:
[...]

PR #1489, opened 2023-09-23
Add `forceNoSelf` option to `tstl`

Z3rio:
This PR aims to provide a way of forcing the `noSelf` flag to files which you cant add this flag to.

An example usage, you are using a type library which is set up wrong, and therefore doesn't include `this: void`, even if it should be included.
And therefore, when you try to build the files, they end up having `nil` added as the first parameter of all functions from that library.

Personally, I had this exact issue (see issue https://github.com/TypeScriptToLua/TypeScriptToLua/issues/1487), and I cant edit the types file since their github repo is private and therefore maintaining the forked version would be a disaster.

Then, you could go into your `tsconfig.json` and add such file to the `noSelfExcludes` list, which would fix that.

I'm open for suggestions on how this could be done better :)

Example usage for FiveM's types (aka citizenfx)
image

### DIFF
--- a/src/transformation/utils/function-context.ts
+++ b/src/transformation/utils/function-context.ts
@@ -20,7 +20,11 @@ function hasNoSelfAncestor(options: CompilerOptions, declaration: ts.Declaration
     if (!scopeDeclaration) {
         return false;
     } else if (ts.isSourceFile(scopeDeclaration)) {
-        if (options.forceNoSelf !== undefined && typeof options.forceNoSelf === "object") {
+        if (
+            options.forceNoSelf !== undefined &&
+            typeof options.forceNoSelf === "object" &&
+            Array.isArray(options.forceNoSelf)
+        ) {
             for (const name of options.forceNoSelf) {
                 if (scopeDeclaration.fileName.includes(name)) {
                     return true;
--- a/src/transformation/utils/function-context.ts
+++ b/src/transformation/utils/function-context.ts
@@ -20,11 +20,7 @@ function hasNoSelfAncestor(options: CompilerOptions, declaration: ts.Declaration
     if (!scopeDeclaration) {
         return false;
     } else if (ts.isSourceFile(scopeDeclaration)) {
-        if (
-            options.forceNoSelf !== undefined &&
-            typeof options.forceNoSelf === "object" &&
-            Array.isArray(options.forceNoSelf)
-        ) {
+        if (Array.isArray(options.forceNoSelf)) {
             for (const name of options.forceNoSelf) {
                 if (scopeDeclaration.fileName.includes(name)) {
                     return true;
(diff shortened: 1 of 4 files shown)


src/transformation/utils/function-context.ts
```diff
@@ -20,11 +20,23 @@ function hasNoSelfAncestor(declaration: ts.Declaration): boolean {
     if (!scopeDeclaration) {
         return false;
     } else if (ts.isSourceFile(scopeDeclaration)) {
+        if (
+            options.forceNoSelf !== undefined &&
+            typeof options.forceNoSelf === "object" &&
```
Curve, line 25: I don't think this is necessary when you use `Array.isArray` afterwards anyway

Z3rio: Sounds fair, yeah. I did some quick testing in a JS compiler online and `Array.isArray` indeed seems to have checks for undefined and the object type built in.

Good catch.

Perryvw:
I'm hesitant of changing the transpiler for a one-off problem. This seems like something you could also do with a [beforeTransform plugin](https://typescripttolua.github.io/docs/api/plugins), that just appends a noSelfInFile comment to the start of your chosen declaration files, just like the compiler does internally: https://github.com/TypeScriptToLua/TypeScriptToLua/blob/master/src/transpilation/transformers.ts#L45

Z3rio:
> I'm hesitant changing the transpiler for a one-off problem. This seems like something you could also do with a [beforeTransform plugin](https://typescripttolua.github.io/docs/api/plugins), that just appends a noSelfInFile comment to the start of your chosen declaration files, just like the compiler does internally: https://github.com/TypeScriptToLua/TypeScriptToLua/blob/master/src/transpilation/transformers.ts#L45

Hm, I would personally not say that this is a one-off problem.
I would assume there are quite a lot of type libraries that have this exact same problem.

It might be possible to do it via a custom plugin, yeah, but perhaps mainstreaming the process / making it easier would be nice as well.
I went through the discord and GitHub issues, and there seem to be a few other people with the same issue.

Perryvw:
I was able to work around with the following plugin:

```ts
import * as tstl from "typescript-to-lua";

declare module "typescript" { interface Statement { jsDoc?: ts.JSDoc[]; } }

class Plugin implements tstl.Plugin {
  public beforeTransform(program: ts.Program, options: tstl.CompilerOptions, emitHost: tstl.EmitHost) {
    for (const file of program.getSourceFiles())
    {
        if (file.isDeclarationFile && file.statements.length)
        {
            if (!file.statements[0].jsDoc) {
                file.statements[0].jsDoc = [];
            }

            const noSelfInFile = ts.factory.createJSDocComment(undefined, [ts.factory.createJSDocUnknownTag(ts.factory.createIdentifier("noSelfInFile"))]);
            file.statements[0].jsDoc.unshift(noSelfInFile);
        }
    }
  }
}

const plugin = new Plugin();
export default plugin;
```

I'll take some more to think if this warrants a core transpiler change or not.

Issue #1471, opened 2023-08-04
Number.parseInt


Zamiell:
Right now, `Number.parseInt` throws an error, saying it is unsupported.

However, the normal `parseInt` is supported. Can `Number.parseInt` be mapped to `parseInt`?

Perryvw:
Added by #1491

PR #1491, opened 2023-09-24
Add Number.parseInt + Number.parseFloat

Z3rio:
Resolves issue: https://github.com/TypeScriptToLua/TypeScriptToLua/issues/1471

This maps `Number.parseInt` to `parseInt` and `Number.parseFloat` to `parseFloat`

### DIFF
--- a/src/LuaLib.ts
+++ b/src/LuaLib.ts
@@ -63,6 +63,8 @@ export enum LuaLibFeature {
     Number = "Number",
     NumberIsFinite = "NumberIsFinite",
     NumberIsNaN = "NumberIsNaN",
+    NumberParseInt = "ParseInt",
+    NumberParseFloat = "ParseFloat",
     NumberToString = "NumberToString",
     NumberToFixed = "NumberToFixed",
     ObjectAssign = "ObjectAssign",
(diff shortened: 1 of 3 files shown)


Zamiell:
Thank you!

Perryvw:
Please add one or two tests for this in https://github.com/TypeScriptToLua/TypeScriptToLua/blob/master/test/unit/builtins/numbers.spec.ts

Z3rio:
Sure ^^

Z3rio:
Would it perhaps look something like that?
Considering `parseInt` & `parseFloat` already exist, and they do the same thing as `Number.parseInt` & `Number.parseFloat`, I based it off of those test cases.

I haven't really worked a lot with Jest though, so would be great if someone could look over it :)

Perryvw:
Maybe a little extreme rerunning all the tests since like you said the functions already existed and were tested throroughly, but from tests-as-specification point of view I guess this is alright

Z3rio:
> Maybe a little extreme rerunning all the tests since like you said the functions already existed and were tested throroughly, but from tests-as-specification point of view I guess this is alright

Yeah, it should probably be possible to modify the old tests to test it both without and with the `Number.` prefix.
Would that be preferrable? If so, I'll go ahead and take a look at it.

Z3rio:
<3

PR #1500, opened 2023-10-16
Support loop continue for 5.0, 5.1, and universal targets

YoRyan:
Reinstates support for the `continue` statement using the original method proposed in #112, which is compatible with Lua versions that don't support `goto`.

Retains the `goto` method for 5.2+ since it is cleaner.

Transpiles from
```typescript
for (let i = 1; i <= 100; i++) {
    if (i < 25) {
        continue;
    }
    print(i);
}
```
to
```lua
do
    i = 1
    while i <= 100 do
        do
            local __continue2
            repeat
                if i < 25 then
                    __continue2 = true
                    break
                end
                print(i)
                __continue2 = true
            until true
            if not __continue2 then
                break
            end
        end
        i = i + 1
    end
end
```


test/unit/loops.spec.ts
```diff
@@ -515,13 +515,16 @@ for (const testCase of [
...
     const expectContinueGotoLabel: util.TapCallback = builder =>
         expect(builder.getMainLuaCodeChunk()).toMatch("::__continue2::");
 
     util.testEachVersion(`loop continue (${testCase})`, () => util.testModule(testCase), {
-        [tstl.LuaTarget.Universal]: builder => builder.expectDiagnosticsToMatchSnapshot([unsupportedForTarget.code]),
-        [tstl.LuaTarget.Lua50]: builder => builder.expectDiagnosticsToMatchSnapshot([unsupportedForTarget.code]),
-        [tstl.LuaTarget.Lua51]: builder => builder.expectDiagnosticsToMatchSnapshot([unsupportedForTarget.code]),
+        [tstl.LuaTarget.Universal]: expectContinueVariable,
+        [tstl.LuaTarget.Lua50]: expectContinueVariable,
+        [tstl.LuaTarget.Lua51]: expectContinueVariable,
```
Perryvw, line 527: I would expect at least one test that asserts that _this actually works_. For example it might be a good idea to do a `testEachVersion` for the "for with continue" and "while with continue" tests

Perryvw (changes requested):

test/unit/loops.spec.ts
YoRyan: Good idea. I've modified all of the `continue` tests to test against all versions.

Perryvw (approved):

Issue #1120, opened 2021-09-04
Proposal for `$return` extension


Cheatoid:
### Problem description
TypeScript does not allow `return` outside of function body (in my opinion, this is a severe limitation).
```typescript
if (os.time() % 2 == 0) {
    return $multi("foo", math.pi); // 'return' statement can only be used within a function body
}
```

### Suggested solution
Therefore I propose to introduce a new extension (`$return`) to remedy this limitation easily:
```typescript
if (os.time() % 2 == 0) {
    $return("foo", math.pi);
}
```

#### Expected Lua translation
This is the goal (in this case), when the above snippet is transpiled, it should produce exactly this output:
```lua
if os.time() % 2 == 0 then
    return "foo", math.pi
end
```
No more, no less.

#### Details
`$return` should work just like [`$multi`](https://typescripttolua.github.io/docs/advanced/language-extensions#multi), accepting multiple arguments (since Lua can return many values), with one important difference being that it would emit the `return` keyword/statement, of course.

In other words, `$return` would allow TSTL to return plain values directly from Lua script on *main* scope, which is otherwise not possible!

hazzard993:
There is an `export =` syntax in TypeScript to allow modules to return plain values

```ts
export = [1, 2, 3];
// this modules now returns [1, 2, 3], nothing else
```

You may need to enable the `esModuleInterop` flag when importing a module that exports using this syntax since the import syntax could work in both ways

```ts
import values from "./array";
import * as values from "./array";
```

Lua also only stores the first returned value from a module so `return "foo", math.pi` means that only `"foo"` is returned from the module

If you want if statements to effect the return value you can do something like this too

```ts
let out: any;

if (cond) {
  out = { ... };
}

export = out;
```

Cheatoid:
> There is an `export =` syntax in TypeScript to allow modules to return plain values

Already tried that and this doesn't work, as described in my post (once again due to TypeScript limitation):
```typescript
if (os.time() % 2 == 0) {
    export = ["foo", math.pi];
}
```
Results in: "An export assignment must be at the top level of a file or module declaration."

> You may need to enable the `esModuleInterop` flag when importing a module that exports using this syntax since the import syntax could work in both ways

This is actually irrelevant in my practical usecase, since it is a (game) engine loading and running a Lua script, and all I care about is that the transpiled Lua code must return plain values directly on *main* scope under certain conditions (when I say plain, I mean it should not have values wrapped in a table).

[...]

PR #1501, opened 2023-10-17
Support for returning values from the global scope

YoRyan:
Lua, unlike TypeScript, allows for the global scope to return values to the `require`'ing script. This is a niche use case, but if you need to interop with existing Lua scripts that are structured around this ability, it's essential to support it. See #1120 for background information.

First, I tried implementing @Cheatoid's `$return(...)` proposal, but the problem is that function calls are expressions rather than statements, so it's not trivial to drop a `return` statement in the intended location. I chose instead to implement a file annotation that instructs TSTL to emit the default export rather than the entire exports object.

Here's how it looks:
```typescript
/** @returnDefaultExport */
export default "Hello, World!";
```
```lua
local ____exports = {}
____exports.default = {"Hello, World!"}
return unpack(____exports.default)
```

And yes, it works with multiple values. 😁 
```typescript
/** @returnDefaultExport */
export default $multi(1, 2, 3);
```
```lua
local ____exports = {}
____exports.default = {1, 2, 3}
return unpack(____exports.default)
```

Tests not currently included. I'm wide open to suggestions on that front.

### DIFF
--- a/src/transformation/utils/annotations.ts
+++ b/src/transformation/utils/annotations.ts
@@ -6,6 +6,7 @@ export enum AnnotationKind {
     NoResolution = "noResolution",
     NoSelf = "noSelf",
     NoSelfInFile = "noSelfInFile",
+    ReturnDefaultExport = "returnDefaultExport",
 }
 
 const annotationValues = new Map(Object.values(AnnotationKind).map(k => [k.toLowerCase(), k]));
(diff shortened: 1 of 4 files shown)


YoRyan:
Withdrawing. I completely missed that we already have the `export =` syntax, so I need to narrow the scope down to the `$multi` case only.

Issue #1498, opened 2023-10-04
Implement `@customName` extension


Cheatoid:
Occasionally there are times where you want to be able to declare a *global* function which also happens to be TS reserved keyword (e.g. `import`, `delete`). But, this means you are unable to call such Lua function, that is without utilizing tstl plugin or a bit of cleverness to workaround such restriction (which is just ugly), the following is an example workaround for invoking global `import` function:
[Playground](https://typescripttolua.github.io/play/#code/PTAEEEBNNAXALAlgZ1Mg9gWwKYHd7YBO2oiAdqAJ7oCuhoAKgMoMAyoADoegFbYDGsAFwAoAGY0ygxOgoASRJg7pCsADwMAfAAoEKIaABu6RJAA0oMgEMcB5LELkA5gEoDDUAG8RoX6GKwdBTa2gD6AOKgVqhWZJRRqKw0VgxWAEYANtguAHRO2LDaAESKyqpFLglwlBzY6GKgCkoqsC7a1jguANwiAL4iIiCgAKIAHjYcWaA0yFb5ouSwRGJW-CQM2PYAsuiQNFPefqDY40pZukjIBsamFlYGZDSYaURuaA7OPf38svZwm7BQABeRqlFoaAE7PZZHRFJb2Co9eGwHInCbnACMAAYMd0gA).

And at some other times, you just feel like renaming Lua call expression/identifier while preserving TS naming (perhaps for better readability, or to remain consistent with the rest of your TS codebase), take this illustrative example:
Input TS:
```typescript
declare namespace net {
    /** @customName get_version */
    export function GetVersion(this: void): number;
}
// Usage:
print(net.GetVersion());
```
Expected transpiled Lua:
```lua
print(net.get_version())
```

The `@customName` extension would allow us to get around the reserved keyword restriction mentioned earlier.
I propose this extension to be added as part of TSTL language extensions out-of-box, because I think it would be future-proof and more beneficial for the users (as opposed to me implementing this through [npm package](https://github.com/Cheatoid/TSTL-extensions)).

Z3rio:
This sir, is fucking genius.

I created a custom plugin to solve this exact issue, but your solution for making it "built-in", would be even better.
I'll have a look at implementing this :)

Z3rio:
@Perryvw you can go ahead and assign me to this if you wish, I've started having a look at it.

Z3rio:
@Perryvw this can be closed as completed

[...]

PR #1505, opened 2023-10-18
Setup custom name support

Z3rio:
This PR aims to provide a way of renaming the built name of functions, or such.

Resolves issue: #1498 

## Example usage
```ts
/** @customName get_version **/
function GetVersion(this: void) {}

GetVersion();
```

### Lua output from the example
```lua
function get_version()
end

get_version()
```


Z3rio:
I'll have to add tests to this before it gets merged :)

Z3rio:
Noticed the reason for the lock files being in the repo, lol
So I added them back again

src/transformation/visitors/identifier.ts
```diff
@@ -53,9 +54,37 @@ function transformNonValueIdentifier(
...
+
+    const customNames = context.options.customNames as Array<{ name: string; value: string }>;
+
+    if (customNames) {
+        const customName = customNames.find(cN => cN.name === text);
+
+        if (customName) {
+            text = customName.value;
+        }
+    }
```
Perryvw, line 67: I don't like this at all, besides the fact this is terrible for tstl performance, it is also a paint o maintain and probably causing a bunch of weird unexpected bugs due to things unexpectedly being renamed.

Unless you have a very convincing argument for why we should provide both CLI and annotation options, I would prefer this be left out.

Z3rio: Just so I understand you correctly, you are referring to the tsconfig integration? Yeah, I would probably agree with you.

Perryvw: Yes, I would remove the CLI and tsConfig part of this MR, and just leave the annotation.

Z3rio: > Yes, I would remove the CLI and tsConfig part of this MR, and just leave the annotation.

yeah, fair enough, I'll do that.

Z3rio:
Perhaps further tests should be added, but please take a look at this one first and tell me what you think, aka if it is good enough or not. If it is, then I'll have a look at adding some more perhaps.

test/unit/identifiers.spec.ts
```diff
@@ -845,3 +845,22 @@ test("lua built-in as in constructor assignment", () => {
         export const result = new A("42").error;
     `.expectToMatchJsResult();
 });
+
+test("customName", () => {
```
Perryvw, line 849: Please give the tests a descriptive name, ie "customName annotation renames function declarations"

test/unit/identifiers.spec.ts
```diff
@@ -845,3 +845,22 @@ test("lua built-in as in constructor assignment", () => {
...
+    const result = util.testModule`
+        /** @customName test2 **/
+        function test(this: void) {}
+        test()
+    `.getLuaResult();
+    if (result.transpiledFiles.length === 0) {
+        throw new Error("No files transpiled");
```
Perryvw, line 856: Don't throw in tests, use expects instead:
```ts
expect(result.transpiledFiles).toHaveLength(1);
const lua = result.transpiledFiles[0].lua;
expect(lua).toBeDefined();
```

test/unit/identifiers.spec.ts
```diff
@@ -845,3 +845,22 @@ test("lua built-in as in constructor assignment", () => {
         export const result = new A("42").error;
     `.expectToMatchJsResult();
 });
+
+test("customName", () => {
+    const result = util.testModule`
+        /** @customName test2 **/
+        function test(this: void) {}
+        test()
+    `.getLuaResult();
```
Perryvw, line 854: The tests should execute the generated code and expect some kind of result, to show that it works.

Perryvw (changes requested):
Would also expect tests for renaming classes, variable declarations, namespaces, and also some declared functions that don't exist (ie `declare function bla`). The tests should execute the generated lua to show that all occurrences have indeed correctly been renamed.

test/unit/identifiers.spec.ts
Perryvw:
```ts
const result = util.testModule`
        /** @customName test2 **/
        function test(this: void) { return 3; }
        export const result = test();
    `
    .expectToEqual({ result: 3 })
    .getLuaResult();
 ```

Comment (Z3rio) [reply test/unit/identifiers.spec.ts:854]: Ah, but that test would still be valid if the function name didn't even change though?

Comment (Perryvw) [reply test/unit/identifiers.spec.ts:854]: Yeah also keep the rest, just add the expect for the execution result so we test it executes correctly

Comment (Z3rio) [reply test/unit/identifiers.spec.ts:854]: > Yeah also keep the rest, just add the expect for the execution result so we test it executes correctly

Ahhh, okay, great, thank you.
I'll do that and add some more tests then. (based on what you noted above)

**Comment by Z3rio** [comment]
There seems to be an issue with renaming the inner names of namespaces.
Aka, the actual namespace name will be changed, but if you have a function inside of that namespace with a custom name, the calling of that function will not receive that customName.
image

**Comment by Perryvw** [comment]
This code is probably hidden somewhere in the callexpression transformer. I have considered unifying that more with the transformIdentifier code but haven't looked at it into much detail.

**Comment by Z3rio** [comment]
I'll see if I can find where it is being transformed sometime later today :)
Are the tests looking fair enough though?

**Comment by Perryvw** [comment]
Tests look fine, I'm still missing one where you put the customName on a `declare function func()` (can't assert execution result on that one ofc).

**Comment by Z3rio** [comment]
> Tests look fine, I'm still missing one where you put the customName on a `declare function func()` (can't assert execution result on that one ofc).

I added a test for that as well now, this does not seem to be being renamed by the current code either though.
I'll see if this is caused by the function not being passed through the identifier transformation, or if the code in there is causing it.

Could you please have a look at `tstl-development` in discord? Thank you

**Comment by Z3rio** [comment]
🎉

File: src/transformation/visitors/call.ts [hunk 5]
```diff
@@ -196,12 +199,16 @@ function transformPropertyCall(
     }
 }
 
-function transformElementCall(context: TransformationContext, node: ts.CallExpression): lua.Expression {
+function transformElementCall(
+    context: TransformationContext,
+    node: ts.CallExpression,
+    customName?: string
```
Perryvw, line 205: This doesn't make sense to me, how can `a[]` have a customName?, if `a` has a custom name it will be handled by transformIdentifier. Please revert this

src/transformation/visitors/call.ts
```diff
@@ -211,6 +218,8 @@ function transformElementCall(context: TransformationContext, node: ts.CallExpre
 
 export const transformCallExpression: FunctionVisitor<ts.CallExpression> = (node, context) => {
     const calledExpression = getCalledExpression(node);
+    const symbol = context.checker.getSymbolAtLocation(calledExpression);
+    const customName = getCustomNameFromSymbol(symbol);
```
Perryvw, line 222: I would put this only in the transformPropertyCall function, as you only pass it there in this function (the transformElementCall seems useless to me, see the other comment). Requesting it this early has a huge performance impact so try to delay it until we are sure we need it.

Perryvw (changes requested):

src/transformation/visitors/call.ts
```diff
@@ -135,12 +136,16 @@ export function transformContextualCallExpression(
     ) {
         // table:name()
         const table = context.transformExpression(left.expression);
-        return lua.createMethodCallExpression(
-            table,
-            lua.createIdentifier(left.name.text, left.name),
-            transformedArguments,
-            node
-        );
+        let name: string = left.name.text;
```
Perryvw, line 139:
```suggestion
        let name = left.name.text;
```
slightly surprised our linter doesnt complain here

Perryvw (approved):

PR #1509, opened 2023-10-20
add support of [[ & ]] string require path in lua source code

### LINKED ISSUE #1508: Literal string includes require('path') cause `could not resolve lua source file` error
Example

A lua file `./src/foo.lua` which includes
```Lua
local usage_hint = 'include this module `const a = require('lib_name')`
```
and included in tstl will result in an error `error TSTL: Could not resolve lua source files for require pat`

XavierCHN:
and fixes #1508


Z3rio:
Please have a look at the failed checks :)

Issue #1387, opened 2023-01-21
@noSelfInFile configuration option is ignored for Function


Grafu:
`@noSelfInFile` configuration option is ignored for Function

After upgrading typescript-to-lua to never version I noticed that my code broke. I investigated and found out that @noSelfInFile directive is ignored for Function type. I am now unable to use Function type and have to write `(...args: any[]) => any` explicitly everywhere where I used to write `Function`. Please see this reproduction example in Playground:

https://typescripttolua.github.io/play/#code/PQKhAIAEDsHsGUCmAbAZgSWgMQJbMeGMAFDEDGs0AzgC7ioCu0ZAQgJ4BqAhsg4gFzgASogoAnACYAeWmJzQA5gBpwWJmRo5KAPnABecAG9i4U+ABEACxTJY5wQAoAlPt0BGJcQC+AblIVqOkZmfXp1dm5eRABtKxs7AF0-YLIHNyd-Slow5giePgAmQRFxaVl5ZXAHADparjEFKkEuaDZohJc9XRa2XQNjMwtrZFt7Ks73Ty9MwJyyAtCUvKiC2OHRpOIUgrSMoA

Grafu:
![image](https://user-images.githubusercontent.com/11540828/213882528-79440968-194b-4ad1-bf28-4da5b39afde0.png)

Z3rio:
The issue here seems to be in this function: https://github.com/Z3rio/TypeScriptToLua/blob/e697c6e3bc852049e80c5d6bf388c6a103634ca6/src/transformation/visitors/call.ts#L289

The one using the `Function` type returns undefined, whilst the other one returns the actual function 🤨

PR #1510, opened 2023-10-21
Fix no self in file check in call visitor

Z3rio:
This PR aims to resolve the issue addressed in issue #1387 

Also went ahead and made the `isContextualCallExpression` function more readable, as the previous one was a bit "hacky" in my opinion

### Code to reproduce the issue:
```ts
/** @noSelfInFile **/

const funcByValue: Record<string, Function> = {
  hello: () => 1
};

const func = funcByValue["hello"];
func(1);

const funcByValue2: Record<string, (...args: any[]) => any> = {
  hello: () => 1
};

const func2 = funcByValue2["hello"];
func2(1);
```

You will notice that the first function in the outputted code will include the self parameter, and the second one won't include it. Whereas neither one should have the self parameter, as the file has `noSelfInFile`

### Example output before changes
```lua
---
-- @noSelfInFile *
funcByValue = {hello = function() return 1 end}
func = funcByValue.hello
func(_G, 1)
funcByValue2 = {hello = function() return 1 end}
func2 = funcByValue2.hello
func2(1)
```

https://typescripttolua.github.io/play/#code/PQKhAIAEDsHsGUCmAbAZgSWgMQJbMeGMAFDEDGs0AzgC7ioCu0ZAQgJ4BqAhsg4gFzgASogoAnACYAeWmJzQA5gBpwWJmRo5KAPnABecAG9i4U+ABEACxTJY5wQAoAlPt0BGJcQC+AblIVqOkZmfXp1dm5eRABtKxs7AF0-YLIHNyd-Slow5giePgAmQRFxaVl5ZXAHADparjEFKkEuaDZohJc9XRa2XQNjMwtrZFt7Ks73Ty9MwJyyAtCUvKiC2OHRpOIUgrSMoA

### DIFF
--- a/src/transformation/visitors/call.ts
+++ b/src/transformation/visitors/call.ts
@@ -15,8 +15,8 @@ import { transformInPrecedingStatementScope } from "../utils/preceding-statement
 import { getOptionalContinuationData, transformOptionalChain } from "./optional-chaining";
 import { transformImportExpression } from "./modules/import";
 import { transformLanguageExtensionCallExpression } from "./language-extensions/call-extension";
-import { AnnotationKind, getFileAnnotations } from "../utils/annotations";
 import { getCustomNameFromSymbol } from "./identifier";
+import { AnnotationKind, getFileAnnotations } from "../utils/annotations";
 
 export function validateArguments(
     context: TransformationContext,
@@ -251,7 +251,6 @@ export const transformCallExpression: FunctionVisitor<ts.CallExpression> = (node
     }
 
     const signature = context.checker.getResolvedSignature(node);
-    const srcFile = node.getSourceFile();
 
     // Handle super calls properly
     if (calledExpression.kind === ts.SyntaxKind.SuperKeyword) {
@@ -270,9 +269,9 @@ export const transformCallExpression: FunctionVisitor<ts.CallExpression> = (node
     let callPath: lua.Expression;
     let parameters: lua.Expression[];
 
-    const isContextualCall = isContextualCallExpression(context, signature, srcFile);
+    const isContextualCall = isContextualCallExpression(context, signature, node.getSourceFile());
 
-    if (isContextualCall) {
+    if (!isContextualCall) {
         [callPath, parameters] = transformCallAndArguments(context, calledExpression, node.arguments, signature);
     } else {
         // if is optionalContinuation, context will be handled by transformOptionalChain.
@@ -299,19 +298,28 @@ function isContextualCallExpression(
     signature?: ts.Signature,
     srcFile?: ts.SourceFile
 ): boolean {
+    const hasNoSelfInFile = srcFile && getFileAnnotations(srcFile).has(AnnotationKind.NoSelfInFile);
+
     if (signature) {
         const declaration = signature.getDeclaration();
 
         if (declaration) {
-            return getDeclarationContextType(context, declaration) === ContextType.Void;
+            const contextTypeCheck = getDeclarationContextType(context, declaration) !== ContextType.Void;
+
+            // respect implicit self over noSelfInFile
+            if (hasNoSelfInFile && contextTypeCheck !== true) {
+                return false;
+            }
+
+            return contextTypeCheck;
         }
     }
 
-    if (srcFile && getFileAnnotations(srcFile).has(AnnotationKind.NoSelfInFile)) {
-        return true;
+    if (hasNoSelfInFile) {
+        return false;
     }
 
-    return context.options.noImplicitSelf ?? false;
+    return !context.options.noImplicitSelf;
 }
 
 export function getCalledExpression(node: ts.CallExpression): ts.Expression {
--- a/src/transformation/visitors/call.ts
+++ b/src/transformation/visitors/call.ts
@@ -288,10 +288,12 @@ export const transformCallExpression: FunctionVisitor<ts.CallExpression> = (node
 };
 
 function isContextualCallExpression(context: TransformationContext, signature: ts.Signature | undefined): boolean {
-    const declaration = signature?.getDeclaration();
+    if (signature) {
+        const declaration = signature.getDeclaration();
 
-    if (declaration) {
-        return getDeclarationContextType(context, declaration) === ContextType.Void;
+        if (declaration) {
+            return getDeclarationContextType(context, declaration) === ContextType.Void;
+        }
     }
 
     return context.options.noImplicitSelf !== undefined ? context.options.noImplicitSelf : false;
--- a/src/transformation/visitors/call.ts
+++ b/src/transformation/visitors/call.ts
@@ -271,7 +271,7 @@ export const transformCallExpression: FunctionVisitor<ts.CallExpression> = (node
 
     const isContextualCall = isContextualCallExpression(context, signature, node.getSourceFile());
 
-    if (isContextualCall) {
+    if (!isContextualCall) {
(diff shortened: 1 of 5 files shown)


Z3rio:
Yeah, this perhaps does not seem to be working as it should (considering the failed tests)

src/transformation/visitors/call.ts
```diff
@@ -284,18 +287,39 @@ export const transformCallExpression: FunctionVisitor<ts.CallExpression> = (node
...
         optionalContinuation.contextualCall = callExpression;
     }
     return wrapResultInTable ? wrapInTable(callExpression) : callExpression;
 };
 
-function isContextualCallExpression(context: TransformationContext, signature: ts.Signature | undefined): boolean {
-    const declaration = signature?.getDeclaration();
-    if (!declaration) {
-        return !context.options.noImplicitSelf;
+function isContextualCallExpression(
```
Perryvw, line 296: You seem to have inverted the meaning of this function -> is contextual call == has non-void context, currently you changed it to return the opposite of that

Perryvw (changes requested):
Missing a test verifying this change actually does what is advertised

src/transformation/visitors/call.ts
Z3rio: Yeah, I tried to make the function a bit more readable, as nearly all returns were reversed in the previous function.
I guess that does make the name of it incorrect though, so I'll go ahead and change it back.

src/transformation/visitors/call.ts
```diff
@@ -267,7 +268,9 @@ export const transformCallExpression: FunctionVisitor<ts.CallExpression> = (node
 
     let callPath: lua.Expression;
     let parameters: lua.Expression[];
-    const isContextualCall = isContextualCallExpression(context, signature);
+
+    const isContextualCall = isContextualCallExpression(context, signature, node.getSourceFile());
```
Perryvw, line 272: I actually don't think this is correct. the sourcefile you need to check is the one where the signature is declared, not the file where the call happens. I'd also like to see this in a test (ie the signature is in a different file to the call). If you need help writing the test ping me in discord.

test/unit/functions/noSelfAnnotation.spec.ts
```diff
@@ -62,3 +62,30 @@ test("explicit this parameter respected over @noSelf", () => {
         export const result = foo(1);
     `.expectToMatchJsResult();
 });
+
+test("respect noSelfInFile over noImplicitSelf", () => {
+    const result = util.testModule`
+        /** @noSelfInFile **/
+
+        const funcByValue: Record<string, Function> = {
```
Perryvw, line 70: Why complicate this with a record? Can't you just write `const func: Function = () => 1;`

Perryvw (changes requested):

test/unit/functions/noSelfAnnotation.spec.ts
Z3rio: Im unsure if the issue really occurs then, I could check, that was the example provided by the Github Issue atleast.

Z3rio: Nevermind, seems like that is possible, odd why the issue would have typed it out in that way then.

src/transformation/visitors/call.ts
Z3rio: Hello, I'm unsure of how this would really be achieved.
For example in the example code above, the declaration variable here is undefined.
Which makes it a bit hard to fetch the actual source file from the declaration variable.

Perryvw: I'm not sure either, I would expect to indeed go via the declaration or symbols somehow. I'm pretty sure it should be possible somehow but not sure how without playing around with the API first

Z3rio: I'll see if I can play around a bit with it.
Fairly weird why the declaration would be undefined in the first place though.

src/transformation/visitors/call.ts
```diff
@@ -290,12 +293,44 @@ export const transformCallExpression: FunctionVisitor<ts.CallExpression> = (node
...
+    const symbol = context.checker.getSymbolAtLocation(calledExpression);
+    let hasNoSelfInFile = false;
+
+    if (symbol?.declarations) {
+        for (const declaration of symbol.declarations) {
+            const srcFile = declaration.getSourceFile();
+
+            hasNoSelfInFile = getFileAnnotations(srcFile).has(AnnotationKind.NoSelfInFile);
+            break;
+        }
```
Perryvw, line 310: This is just not right, you only check the first declaration and break, wether it had noselfinfile or not. You only want to break when the annotation was found. I would just write this entire thing as
```ts
const hasNoSelfInFile = symbol?.declarations.some(d => getFileAnnotations(d.getSourceFile()).has(AnnotationKind.NoSelfInFile)) ?? false;
```

Z3rio: Ah, I was assuming that there would anyways only be one file that it is declared in, my fault then.

Z3rio: My latest commit should resolve what was noted here.

test/unit/functions/noSelfAnnotation.spec.ts
```diff
@@ -62,3 +62,25 @@ test("explicit this parameter respected over @noSelf", () => {
         export const result = foo(1);
     `.expectToMatchJsResult();
 });
+
+test("respect noSelfInFile over noImplicitSelf", () => {
```
Perryvw, line 66: Still missing a test where the declaration of the function is in a different file than the call to that function,

src/transformation/visitors/call.ts
```diff
@@ -290,12 +293,41 @@ export const transformCallExpression: FunctionVisitor<ts.CallExpression> = (node
...
+    let hasNoSelfInFile = false;
+
+    if (symbol?.declarations) {
+        hasNoSelfInFile =
+            symbol.declarations.some(d => getFileAnnotations(d.getSourceFile()).has(AnnotationKind.NoSelfInFile)) ??
+            false;
     }
-    return getDeclarationContextType(context, declaration) !== ContextType.Void;
+
+    if (signature) {
```
Perryvw, line 310: why did you unwrap the ?. that was there before to this if statement actually? Seems unnecessary

Z3rio:
I've added the requested test, although there currently seems to be some issues with the current solution, when importing functions from other files with `noSelfInFile`, as it uses the `noSelfInFile` from the calling file, and not the declaration file.

Perryvw:
> I've added the requested test, although there currently seems to be some issues with the current solution, when importing functions from other files with `noSelfInFile`, as it uses the `noSelfInFile` from the calling file, and not the declaration file.

Oh I was just expecting a more complicated version of the test you already had with `testFunction` and `addExtraFile`, we try to avoid actual file system operations if possible. If you need help getting that setup contact me on discord.

Z3rio:
> > I've added the requested test, although there currently seems to be some issues with the current solution, when importing functions from other files with `noSelfInFile`, as it uses the `noSelfInFile` from the calling file, and not the declaration file.
> 
> Oh I was just expecting a more complicated version of the test you already had with `testFunction` and `addExtraFile`, we try to avoid actual file system operations if possible. If you need help getting that setup contact me on discord.

Ah, yeah, I could do that.
I've written to you on discord concerning a bigger issue though :)

Perryvw:
Closing this as superceded by #1519

Issue #1503, opened 2023-10-17
"match is unsupported" error message should be improved


Zamiell:
The message should explicitly tell you to use `string.match(foo, regex)`.

Z3rio:
Could you please supply a sample ts snippet to show what the issue is a bit more specifically, thank you :)

Zamiell:
Sorry, I was not clear enough in the original post.

Consider the following code snippet, which is extremely common:

```ts
const string1 = "The quick brown fox jumps over the lazy dog.";
const match = string1.match("brown (\w+) jumps");
```

The playground for this is here:
 https://typescripttolua.github.io/play#code/MYewdgzgLgBNBOBLMBzAjDAvDARAFQAsBTGARwFdFgBrGAI3hAHcwYAzEADxgCtyBbAA4QYIAG5F4MKMRgAbAIYAvAJ4wAJiBQA6HAG4AUKEix+CqMAJY4UJKjTazFggAocDZqxcAdJgGoASl4BYRwAwyA

We can see that it throws the error:
"string.match is unsupported".

But this error message sucks and is misleading, because what it really means is that the `String.match` method is unsupported. The Lua global `string.match` is absolutely supported.

What we really need to do here is to reconfigure the code to look like this:

```ts
const string1 = "The quick brown fox jumps over the lazy dog.";
const match = string.match(string1, "brown (%w+) jumps");
```

So, in conclusion, I think the error message should say:

> The `String.match` method is unsupported. Consider including the Lua types for your environment and then changing the code to `string.match(foo, regex)` instead. Note that the Lua match works differently than the JavaScript/TypeScript match, so for example you would have to change `\w+` to `%w+`.

PR #1512, opened 2023-10-22
Add `String.prototype.match()` support

Z3rio:
This PR aims to resolve issue #1503, aka adds support for `String.prototype.match()`

### Example usage
```ts
const string1 = "The quick brown fox jumps over the lazy dog.";
const match = string1.match("brown");
```

Only concern I have is whether the functions really return the same things, it seems to do so though.


Perryvw:
These methods do not return the same thing at all, nor do they accept the same input arguments. I wish this mapping was so simple but it's not, and pretending this works to our users is just going to cause a lot of bugs and frustration.

Z3rio:
Could this be reopened but marked as a draft? As this is something I would like to have a look at.

PR #1513, opened 2023-10-30
String, Number & Boolean classes & methods

Z3rio:
This PR aims to fix the following:
- The `Boolean` prototype/method
- The `Boolean` class
- The `String` prototype/method
- The `String` class
- The `Number` class

This TSTL Playground should showcase all of the issues noted above:
https://typescripttolua.github.io/play#code/MYewdgziA2CmB00QHMAUZYHcAEBlALgE4CWYaARABbHkCUtA3AFCiQwJJoElmpU31mrKHEQpUAIRDsAhmD4AGOoxbgRHcRhxTZ88gEZlQtezFot2AHIBXALYAjWIT76ATAGYjTIA


Z3rio:
looks like this perhaps needs some fine tuning aswell, considering the failed tests lol

src/lualib/New.ts
```diff
@@ -1,4 +1,29 @@
...
+            return __TS__Boolean(args[0]);
+        case "Number":
+            return __TS__Number(args[0]);
+        default:
+            return undefined;
+    }
+}
+
 export function __TS__New(this: void, target: LuaClass, ...args: any[]): any {
+    if ("name" in target && typeof target.name === "string") {
```
Perryvw, line 19: The niche usecases of these functions do not warrant this performance hit on every new object being created. The choice between the different constructors can be made int he compiler at compile time, no need to burden the runtime with this.

src/lualib/BooleanClass.ts
```diff
@@ -0,0 +1,17 @@
+import { __TS__Boolean } from "../lualib/Boolean";
+
+export class Boolean {
```
Perryvw, line 3: I don't really see the added value of providing these as actual classes, I think it will just cause confusing bugs, especially since the equality comparisons will be broken. I think I would prefer to just translate `new Boolean(x)` the same as `Boolean(x)`.

src/lualib/StringValueOf.ts
```diff
@@ -0,0 +1,16 @@
+import { __TS__String } from "./String";
+import { String } from "./StringClass";
+
+export function __TS__StringValueOf(this: unknown): string | undefined {
```
Perryvw, line 4: I also don't see the added value of adding these valueOf functions.

Perryvw (changes requested):

src/lualib/StringValueOf.ts
Z3rio: I added those to assist with the test cases, mainly for the Number class as it didn't have a way of getting the raw number.

src/lualib/New.ts
Z3rio: Concerning the classes, what way do you think we should go about handling those instead? As I did not personally find any way of directly translating `new Boolean` to `Boolean`, etc
(This comment is related to both this conversation, and the one below)

PR #1519, opened 2023-11-19
Reworked function context checking based on #1510

### LINKED ISSUE #1387: @noSelfInFile configuration option is ignored for Function

Perryvw:
Update on top of #1510 

- Got rid of isContextualCallExpression.
- Changed the signature of `getDeclarationContextType(context, ts.SignatureDeclaration)` to `getCallContextType(context, ts.CallLikeExpression)`
- If context checking fails based on signature, instead try to find if the node of the called function is in a file with noSelfInFile

Fixes #1387


Z3rio:
Havent had a deep dive in all of the changes yet, but it does look fair enough on a first glance, good work :)

PR #1530, opened 2024-02-03
Optimize Promise and Await (fixes #1499), fix Promise.resolve bug

### LINKED ISSUE #1499: C stack overflow when using await in a loop
Hello,

I've encountered a "C stack overflow" error while running the following TypeScript code which gets transpiled to Lua using TypeScriptToLua:
`for (let i = 0; i < 40; i++) {
    await delay(0.1);
    print('times:', i);
}`
The function delay is an asynchronous function that returns a Promise that resolves after the specified delay. The print function logs the current iteration to the console.

The issue occurs when the loop reaches the 36th iteration (i=35), at which point the program crashes with a "C stack overflow" error. The expected behavior would be for the loop to successfully complete all 40 iterations without any issues.

I've tried debugging the issue on my end but haven't been able to find a workaround. I've also ensured that the issue is not due to any memory leaks, high CPU usage, or problems with the delay function or the Lua runtime.

I'm using TypeScriptToLua version 1.20.1 and running the resulting Lua code in environment Warcraft III

Any help on this issue would be greatly appreciated.

Thank you!

rhazarian:
Fixes https://github.com/TypeScriptToLua/TypeScriptToLua/issues/1499.

I've tried to find a shaky balance between readable and optimized code. I've refactored almost all of the recursive calls into tail calls, and also prevented calling coroutine.resume from inside pcall's. Additionally, I got rid of some of the unnecessary allocations, and discovered & fixed a bug with Promise.resolve (tests added).

The new Promise + Async version can handle virtually any amount of chained waits. Due to the coroutine.resume/pcall optimization, C stack doesn't grow, and due to tail calls Lua stack doesn't grow as well. It stays constant and very small: http://tpcg.io/_ZQYND5

### DIFF
--- a/src/lualib/Await.ts
+++ b/src/lualib/Await.ts
@@ -21,11 +21,6 @@ const coresume = coroutine.resume;
 const costatus = coroutine.status;
 const coyield = coroutine.yield;
 
-// The precise implementation type is important here because we use special functions due to necessary optimizations
-function adopt(this: void, value: unknown): __TS__Promise<unknown> {
-    return value instanceof __TS__Promise ? value : (__TS__Promise.resolve(value) as __TS__Promise<unknown>);
-}
-
 // Be extremely careful editing this function. A single non-tail function call may ruin chained awaits performance
 // eslint-disable-next-line @typescript-eslint/promise-function-async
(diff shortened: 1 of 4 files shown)


Perryvw (approved):
Looks great!

PR #1534, opened 2024-02-06
Add basic check for customName in property access transform

### LINKED ISSUE #1528: @customName is ignored if @noSelf is applied
I'm currently looking for a solution and if I find anything I will create a pull request

Works without `@noSelf` attribute
Same issue with `@noSelfInFile`

https://typescripttolua.github.io/play#code/PQKhAIAEDsHsGUCmAbAZuEwBQATRBjZAQwCdFxoiBbRAZwAcj9yBzaugFyI-IG8twg8KAiR8AV1odYVAHLtwjWrQzYh4VOOj4OAS1jQNRXcgAUASgBc4AG6xdOLAF8sWNjSndEAOlTGz5gDcQA

```ts
/** @noSelf */
declare namespace gamestate {
    /** @customName pass */
    function fail(): void
}

gamestate.fail();
```

Expected:
```lua
gamestate.pass()
```

Got:
```lua
gamestate.fail()
```

aabajyan:
While simple name changes work as expected names with nested methods will still behave like string literals.
I will try to find a solution for this as well, but since this is my first commit I want to make this as simple as possible.

partially fixes #1528

### DIFF
--- a/src/transformation/visitors/access.ts
+++ b/src/transformation/visitors/access.ts
@@ -23,6 +23,7 @@ import {
     captureThisValue,
 } from "./optional-chaining";
 import { SyntaxKind } from "typescript";
+import { getCustomNameFromSymbol } from "./identifier";
 
 function addOneToArrayAccessArgument(
     context: TransformationContext,
@@ -108,9 +109,15 @@ export function transformPropertyAccessExpressionWithCapture(
     node: ts.PropertyAccessExpression,
     thisValueCapture: lua.Identifier | undefined
 ): ExpressionWithThisValue {
-    const property = node.name.text;
     const type = context.checker.getTypeAtLocation(node.expression);
     const isOptionalLeft = isOptionalContinuation(node.expression);
+
+    let property = node.name.text;
+    const symbol = context.checker.getSymbolAtLocation(node.name);
+    const customName = getCustomNameFromSymbol(symbol);
+    if (customName) {
+        property = customName;
+    }
 
     const constEnumValue = tryGetConstEnumValue(context, node);
     if (constEnumValue) {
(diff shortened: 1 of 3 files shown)


src/transformation/visitors/access.ts
```diff
@@ -108,10 +109,19 @@ export function transformPropertyAccessExpressionWithCapture(
     node: ts.PropertyAccessExpression,
     thisValueCapture: lua.Identifier | undefined
 ): ExpressionWithThisValue {
-    const property = node.name.text;
     const type = context.checker.getTypeAtLocation(node.expression);
     const isOptionalLeft = isOptionalContinuation(node.expression);
 
+    // FIXME: Stuff like very.nested.call won't work, because
```
Perryvw, line 115: I tested with 
```ts
/** @noSelf */
interface Nested {
    /** @customName pass */
    fail(): void;
}

declare namespace TestNamespace {
    const nested: Nested;
}

TestNamespace.nested.fail();
```
And
```ts
/** @noSelf */
interface Nested {
    /** @customName pass */
    fail(): void;
}

declare const TestObject: {
    nested: Nested;
};

TestObject.nested.fail();
```

And they both seem to work fine with custom name. Can you give an example  test case where it doesn't work?

aabajyan: Sure
```ts
/** @noSelf */
declare namespace TestNamespace {
    /** @customName very.nested.call */
    function fail(): void;
}

TestNamespace.fail();
```

Will output:
```lua
TestNamespace["very.nested.call"]()
```

Instead of:
```lua
TestNamespace.very.nested.call()
```

Perryvw: `/** @customName very.nested.call */` Is not really intended behavior for customName at the moment.

I would prefer if you remove this FIXME comment and just deliver this code as it does seem to resolve the original bug. I think allowing nested customName values might be a nice follow-up feature for a separate PR.

aabajyan: Sure thing

PR #1536, opened 2024-02-06
Fix tsconfig.spec.ts not working on MacOS

### LINKED ISSUE #1535: tsconfig.spec.ts fails in MacOS, because /var is actually located in /private/var
I think having something like `fs.realpathSync` will solve the issue. I will test this and create a pull request later today.

```
  expect(received).toBe(expected) // Object.is equality

    Expected: "/var/folders/l1/q852l81d1lv95zhblgdvf0gr0000gn/T/tstl-test-FXslwE/tsconfig.json"
    Received: "/private/var/folders/l1/q852l81d1lv95zhblgdvf0gr0000gn/T/tstl-test-FXslwE/tsconfig.js
```

aabajyan:
Closes #1535

### DIFF
--- a/test/cli/tsconfig.spec.ts
+++ b/test/cli/tsconfig.spec.ts
@@ -20,7 +20,7 @@ afterEach(async () => {
 const locate = (project: string | undefined, fileNames: string[] = []) =>
     locateConfigFile({ errors: [], fileNames, options: { project } });
 
-const normalize = (name: string) => normalizeSlashes(path.resolve(temp, name));
+const normalize = (name: string) => normalizeSlashes(fs.realpathSync(path.resolve(temp, name)));
 
 describe("specified", () => {
     for (const separator of process.platform === "win32" ? ["/", "\\"] : ["/"]) {


Perryvw:
I don't quite understand what the issue is here. What is the expected/actual on the failing tests? What does realpath change?

aabajyan:
In MacOS tsconfig.spec.ts fails because the paths do not match. It expected /private directory which is missing.
Realpath returns the actual path so if you get realpath of /var and /var is a symlink of /private/var it will return /private/var.
I also noticed that it doesn't work on Windows now, so I will run this function on darwin only.

test/cli/tsconfig.spec.ts
```diff
@@ -20,7 +20,15 @@ afterEach(async () => {
 const locate = (project: string | undefined, fileNames: string[] = []) =>
     locateConfigFile({ errors: [], fileNames, options: { project } });
 
-const normalize = (name: string) => normalizeSlashes(path.resolve(temp, name));
+const normalize = (name: string) => {
+    const result = normalizeSlashes(path.resolve(temp, name));
+
+    if (process.platform === "darwin") {
```
Perryvw, line 26: Please revert to what it was before, that was much cleaner, but put the realPathSync call inside normalizeSlashes instead of ourside to prevent windows breaking

aabajyan: Sure thing. sorry about that.

PR #1583, opened 2024-09-15
Fix function assignment context validation bug

### LINKED ISSUE #1568: Generic callback functions and 'this'
```
/** @noSelf */
declare namespace Test {
  export function testCallback<T extends (...args: any[]) => void>(
    callback: T,
  ): FunctionContainer<T>;
}

testCallback(() => {});
```

This fails with `error TSTL: Unable to convert function with a 'this' parameter to function 'callback' with no 'this'. To fix, wrap in an arrow function, or declare with 'this: void'.`.

validateFunctionAssignment fails here. I could make it work by returning immediately if `toType.isTypeParameter()` is true, but I suspect that's not the correct solution to this particular problem.

Perryvw:
Fixes #1568

### DIFF
--- a/test/unit/functions/validation/validFunctionAssignments.spec.ts
+++ b/test/unit/functions/validation/validFunctionAssignments.spec.ts
@@ -248,3 +248,21 @@ test("Does not fail on union type signatures (#896)", () => {
         )
         .expectToHaveNoDiagnostics();
 });
+
+// https://github.com/TypeScriptToLua/TypeScriptToLua/issues/1568
+test("No false positives when using generic functions (#1568)", () => {
+    util.testModule`
+        /** @noSelf */
+        declare namespace Test {
+            export function testCallback<T extends (...args: any[]) => void>(
+                callback: T,
+            ): void;
+            export function testCallback2(
+                callback: (...args: any[]) => void,
+            ): void;
+        }
+
+        Test.testCallback(() => {});
+        Test.testCallback2(() => {});
+    `.expectToHaveNoDiagnostics();
+});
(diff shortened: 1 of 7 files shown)


test/unit/functions/validation/validFunctionAssignments.spec.ts
```diff
@@ -248,3 +263,21 @@ test("Does not fail on union type signatures (#896)", () => {
...
+            export function testCallback<T extends (...args: any[]) => void>(
+                callback: T,
+            ): void;
+            export function testCallback2(
+                callback: (...args: any[]) => void,
+            ): void;
+        }
+
+        Test.testCallback(() => {});
+        Test.testCallback2(() => {});
```
lolleko, line 281: Could we add a test for:

```
const x = () => {};

Test.testCallback(x);
```

as this also fails on the current version:

https://typescripttolua.github.io/play#code/PQKhAIAEDsHsGUCmAbAZuEwBQATRBjZAQwCdFxoiBbRAZwAcj9yAVOgF3AG8txxEAHvVglOqAK7R87AJaxo4dhwDCRZMgBGTANYAeFvwFLoOWuAAUAOmukA5rQBc4ItACeAbQC6ASnABeAD5wADdYGRwA814+cHw1TR0nFgAaaO8nF1cAbiwAXywsfHlaTgF-C19A7lycthLLJRLVdS18bXMBbyygA

Perryvw:
This solution was too hard to get to work, and due to a large risk of regressions in user code, we decided to go with #1589 as solution instead

PR #1588, opened 2024-09-24
chore: update eslint to latest

### LINKED ISSUE #1582: Upgrade TypeScript ESlint to the most recent version
We are on an old outdated version of eslint, we should upgrade the repo to the newest version, and possibly adapt our configuration to the new configuration format.

Zamiell:
Closes #1582 

RE: no-default-export
I spent a few hours trying to get `eslint-plugin-import` working, but was not able to. Afterwards, I realized that we are only using one rule from that plugin called "no-default-export", which just bans "export default foo". I don't think we are getting much value from that rule, so it makes sense just to remove it, especially since there are a bunch of "eslint-disable" for it in the codebase.

RE: ?? changes
Newer versions of the nullish-coalescing rules trigger on the existing codebase, so I went ahead and fixed the code to satisfy them.

RE: enhanced-resolve
I had to lock the version to prevent a build error. You might want to open an issue here: https://github.com/webpack/enhanced-resolve/issues

RE: strict
I had to add strict to the tsconfigs for tests, since several rules require strict to be on. (The new version of TSESLint uses the closest tsconfig.json to lint the file.)

RE: ci
had to update stuff in CI because linting was failing due to an old node version

In a future PR I think we should use the recommended rules from ESLint + TSESLint.


src/transformation/context/context.ts
```diff
@@ -39,7 +39,8 @@ export class TransformationContext {
     public readonly luaTarget = this.options.luaTarget ?? LuaTarget.Universal;
     public readonly isModule = ts.isExternalModule(this.sourceFile);
     public readonly isStrict =
-        (this.options.alwaysStrict ?? this.options.strict) ||
+        this.options.alwaysStrict ??
+        this.options.strict ??
         (this.isModule && this.options.target !== undefined && this.options.target >= ts.ScriptTarget.ES2015);
```
Perryvw, line 44: This looks like it has a different meaning, before if alwaysStrict and strict were both false, but the latter clause is true, the result would be true, now if alwaysStrict is not undefined but false, it will not even look at the final clause

test/transpile/module-resolution/project-with-tstl-library-has-exports-field/tsconfig.json
```diff
@@ -1,5 +1,6 @@
 {
     "compilerOptions": {
+        "strict": true,
```
Perryvw, line 3: I don't like we are being forced to add strict everywhere. What would happen if we don't do this, and which rules would we need to disable?

Zamiell: One clean option is to just remove the strict everywhere and exempt the testing files from linting. Would you be okay with that? Otherwise you just have to exempt all the rules that require strict, I'm not sure which, it gets more complicated and harder to maintain.

src/transformation/context/context.ts
Zamiell: I'm not really sure, this is what the ESLint rule auto-fixed to, so I assume that it is safe. If you are worried about it, I can revert the change and then ignore the ESLint error, just let me know.

Perryvw: It's not the same, this should be reverted. Maybe if you change `(this.options.alwaysStrict ?? this.options.strict)` to `(this.options.alwaysStrict ?? this.options.strict ?? false)` it will swallow it without needing an eslint ignore, or else maybe pulling that out to a separate variable might do the trick

test/transpile/module-resolution/project-with-tstl-library-has-exports-field/tsconfig.json
Perryvw: I don't want to remove the strict everywhere, I just don't want to see any strict changes in the diff for this pull request. Can you answer my question? What would happen if we don't add the stricts, and what rules would we need to disable to let eslint agree

src/transformation/context/context.ts
Zamiell: ok its reverted now

test/transpile/module-resolution/project-with-tstl-library-has-exports-field/tsconfig.json
Zamiell: Yeah when I said "remote the strict everywhere" I meant only in my PR, not elsewhere in the project. I don't know how many rules conflict with strict, it would be a bunch of trial and error to figure out of commenting them out one by one. As I said before, maintaining a separate blacklist for conflicting rules is a pretty bad long term solution, so for now I am just using the legacy project behavior, which allows you to keep strict off.

Perryvw (approved):
Thanks for shis change

PR #1594, opened 2024-11-08
chore: update deps to latest

Zamiell:
The deps were very old so this PR updates them to the latest.
Additionally, this fixes the big warning when running `npm run lint`.
Updating Prettier to latest required also running `npm run fix:prettier`, which changed a lot of files.
Please do a release after this PR so that end-users can solve the peer dependencies problem with TypeScript.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,34 +42,34 @@
         "node": ">=16.10.0"
     },
     "peerDependencies": {
-        "typescript": "5.6.2"
+        "typescript": "5.6.3"
     },
     "dependencies": {
         "@typescript-to-lua/language-extensions": "1.19.0",
-        "enhanced-resolve": "5.8.2",
-        "picomatch": "^2.3.1",
-        "resolve": "^1.15.1",
-        "source-map": "^0.7.3"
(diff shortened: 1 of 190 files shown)


package.json
```diff
@@ -42,34 +42,34 @@
...
         "lua-wasm-bindings": "^0.3.1",
-        "prettier": "^2.8.8",
-        "ts-jest": "^29.1.0",
-        "ts-node": "^10.9.1",
-        "typescript": "^5.6.2",
-        "typescript-eslint": "^8.7.0"
+        "prettier": "^3.3.3",
+        "ts-jest": "^29.2.5",
+        "ts-node": "^10.9.2",
+        "typescript": "5.6.3",
```
Zamiell, line 72: I removed the caret here, since having it here was a bug. (This should be kept exactly the same as what the peer dependency specifies.)

Zamiell:
can you update the deps then? it should be easy to replicate this pull request, just type the following:

```
npx npm-check-updates -u
npm install
npm run fix:prettier
```

and then fix the typescript bug in package.json, and update the changelog.

PR #1608, opened 2024-12-23
Basic `Luau` target

Vurv78:
[`Luau`](https://luau.org) is a variant of Lua semi-recently becoming popular due to its open-sourcing and use by Roblox. It has a mix of features from 5.1-5.4 that make it not really fit into any traditional Lua version.

## Why

There's [roblox-ts](https://github.com/roblox-ts/roblox-ts) but that's another project and not everyone is using Luau for Roblox. We're using it for our own game engine - and I'm aware of other projects making use of Luau (the popular [mlua](https://github.com/mlua-rs/mlua) rust library supports it, despite being a "lua" library)

`Luau` does include a type system, but it's not yet capable enough for my usecases. Plus, that's ignoring all the other benefits that Typescript brings.

## Changes

This PR:
* Adds a `Luau` `luaTarget`
* Adds a `continue` statement used solely by Luau.
* Adds the `if <COND> then <X> else <Y>` ternary operator used solely by Luau.

There are other Luau-only features like type hints and compound assignment operators - but they could be done later on and aren't as pertinent

Please let me know how I should incorporate test cases - couldn't figure out how to do so myself, although I tested these features in a local project.


Perryvw (commented):
Looks good, though the test run seems to indicate you forgot some stuff still. 

It is not really possible to execute luau in our tests so that's fine, but looking at the change it might be nice if you add a luau-specific translation snapshot test containing a continue and ternary expression here: https://github.com/TypeScriptToLua/TypeScriptToLua/blob/master/test/translation/transformation.spec.ts

Vurv78:
Added a transformation test although I'd like to make sure you're okay with the current implementations for two potential issues:

1. I implemented the `Luau` specific tests simply by checking the filename in the test spec. Not the most elegant solution but the simplest. Maybe better off splitting into folders per version. Up to you.

2. I don't understand the rationale behind `expectEachVersionExceptJit` to confide in my solution (I have it currently passing the expectation for Luau as well). Depending on the intention, it could be renamed and exclude Luau as well.

test/translation/transformation.spec.ts
```diff
@@ -12,9 +12,12 @@ const fixtures = fs
     .sort()
     .map(f => [path.parse(f).name, fs.readFileSync(path.join(fixturesPath, f), "utf8")]);
 
-test.each(fixtures)("Transformation (%s)", (_name, content) => {
+test.each(fixtures)("Transformation (%s)", (name, content) => {
     util.testModule(content)
-        .setOptions({ luaLibImport: tstl.LuaLibImportKind.Require })
+        .setOptions({
+            luaLibImport: tstl.LuaLibImportKind.Require,
+            luaTarget: name.includes("luau") ? tstl.LuaTarget.Luau : undefined,
```
Perryvw, line 19: I was expecting just a second test in this file that hardcodes your file name and luau target, but this works too I guess

test/util.ts
```diff
@@ -87,6 +87,7 @@ export function expectEachVersionExceptJit<T>(
         [tstl.LuaTarget.Lua53]: expectation,
         [tstl.LuaTarget.Lua54]: expectation,
         [tstl.LuaTarget.LuaJIT]: false, // Exclude JIT
+        [tstl.LuaTarget.Luau]: expectation,
```
Perryvw, line 90: The reasoning for this function is that we are running WASM-versions of the lua runtimes (5.0-5.4) during our test runs, but we do not have JIT running. So this function is mostly for asserting things about the execution results in the Lua runtime, hence I think just putting Luau to false here is probably more appropriate

Perryvw:
You can just run the tests locally with `npm test` (after npm install ofc) to check if everything is working. or even `npx jest <test file name>` to run specific files.

Vurv78:
Tests should be in a good state now

test/unit/loops.spec.ts
```diff
@@ -553,6 +553,7 @@ for (const testCase of [
         [tstl.LuaTarget.Lua53]: expectContinueGotoLabel,
         [tstl.LuaTarget.Lua54]: expectContinueGotoLabel,
         [tstl.LuaTarget.LuaJIT]: expectContinueGotoLabel,
+        [tstl.LuaTarget.Luau]: () => undefined, // TODO: This is N/A.
```
Perryvw, line 556: Why is this this N/A? Shouldn't this be expecting the Luau continue statement?

Vurv78: That test specifically looks for the label or sentinel variable generated by the code for the other targets which isn't applicable to Luau since it just uses the builtin continue without a need for either.

PR #1609, opened 2024-12-25
Allow new Array() syntax for constructing lua tables

Perryvw:


Z3rio:
This does not take cases such as
```js
new Array(length)
new Array(item1, item2)
```
into account

Perryvw:
> This does not take cases such as
> 
> ```js
> new Array(length)
> new Array(item1, item2)
> ```
> 
> into account, it would just overwrite both of them into `[]`, no?

Good call, I supported the `new Array(item1, item2)` syntax too now (simply becomes `{ item1, item2 }`, constructing array with predefined length is not supported due to Lua struggling with nils in arrays, I added a diagnostic in that case

PR #1610, opened 2025-01-13
Allow stacktraces in Lua 5.1 and LuaJIT and fix level in lualib

emmachase:
Lua 5.1 and LuaJIT have a bug where it's not possible to specify the level without a message. Previously the stacktrace was just ignored in TSTL for these versions; but we can workaround this bug by just specifying an empty message and trimming the newline with string.sub.

Also:
* Fix the constructor passed when getting the stacktrace, as `constructor.new` doesn't seem to exist. I suspect that .new used to be a thing but was never updated when this was changed.
  * Is there any circumstance where finding the `__TS__New` would result in the wrong level? Since `__TS__New` is not specifically for Error.
* I also noticed a typo and what looks like a bug in the testEachVersion util where if it found any target that was disabled, it would completely exit even if other versions were specified.

### DIFF
--- a/src/lualib/Error.ts
+++ b/src/lualib/Error.ts
@@ -1,3 +1,5 @@
+import { __TS__New } from "./New";
+
 interface ErrorType {
     name: string;
     new (...args: any[]): Error;
@@ -22,6 +24,10 @@ function getErrorStack(constructor: () => any): string | undefined {
 
     if (_VERSION.includes("Lua 5.0")) {
         return debug.traceback(`[Level ${level}]`);
+    } else if (_VERSION === "Lua 5.1") {
+        // Lua 5.1 and LuaJIT have a bug where it's not possible to specify the level without a message.
+        // @ts-ignore Fails when compiled with Lua 5.0 types
+        return string.sub(debug.traceback("", level), 2);
     } else {
         // @ts-ignore Fails when compiled with Lua 5.0 types
         return debug.traceback(undefined, level);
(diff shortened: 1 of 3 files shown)


src/lualib/Error.ts
```diff
@@ -22,6 +24,10 @@ function getErrorStack(constructor: () => any): string | undefined {
 
     if (_VERSION.includes("Lua 5.0")) {
         return debug.traceback(`[Level ${level}]`);
+    } else if (_VERSION === "Lua 5.1") {
```
Perryvw, line 27: Guess we should use `.includes` here too, to trick whatever type nonsense is going on with the version-specific types

test/unit/error.spec.ts
```diff
@@ -343,3 +344,25 @@ test("still works without debug module", () => {
...
+    "error stacktrace omits constructor and __TS_New",
+    () => util.testFunction`
+        const e = new Error();
+        return e.stack;
+    `,
+    {
+        ...util.expectEachVersionExceptJit(builder => {
+            builder.expectToHaveNoDiagnostics();
+            const luaResult = builder.getLuaExecutionResult();
+            expect(luaResult.split('\n').length).toBe(4);
```
Perryvw, line 358: ugh, linter seems confused about our helper function here. I don't like it but I guess you can add `// eslint-disable-next-line jest/no-standalone-expect` here

Perryvw (commented):
Thanks for the contribution, looks good aside from the issues the linter seems to be having! You can check  for linter errors locally with `npm run lint`.

src/lualib/Error.ts
emmachase: I was going by the declaration which only seems to have 5.1, I guess includes is used to catch all of the 5.0 subversions. I think I just need to include another `@ts-ignore` here. This one is my bad since I changed this last minute without running another build.

emmachase:
Thanks for the comments, I totally forgot to run lint, I was assuming it would run alongside build. Will address these.

emmachase:
Should be good now :)

PR #1613, opened 2025-01-22
fix: ensure parent pointers are correctly set for transformed nodes

ChouUn:
I've been working on writing plugins recently and encountered some issues, so I've been reading source codes of TSTL to learn more. Thus, I noticed some parts that seem a bit unreasonable.

Previously, the `setParent` function was applied to `updatedBlock` instead of `result`, which may lead to incorrect parent pointers in the final transformed AST. This fix ensures that parent pointers are correctly set for the final transformed node `result` and its subtree, maintaining the integrity of the AST during transformations.

As I am just beginning to delve into the source code, my understanding of it may not be entirely accurate. If this is an erroneous fix, please correct me and do not hesitate to explain the rationale behind your coding approach.


Perryvw (commented):
The before and after of this code looks like they should behave identically to me. Did this actually solve the problem you were having? Can you give an example of a transformer that breaks here?

PR #1614, opened 2025-01-30
Add in-memory plugins

iced-wav:
Allows passing plugins or plugin factories directly in `luaPlugins`. Some unit tests are included to illustrate usage.

### DIFF
--- a/src/CompilerOptions.ts
+++ b/src/CompilerOptions.ts
@@ -1,6 +1,7 @@
 import * as ts from "typescript";
 import { JsxEmit } from "typescript";
 import * as diagnosticFactories from "./transpilation/diagnostics";
+import { Plugin } from "./transpilation/plugins";
 
 type OmitIndexSignature<T> = {
     [K in keyof T as string extends K ? never : number extends K ? never : K]: T[K];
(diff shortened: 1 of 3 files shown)


src/CompilerOptions.ts
```diff
@@ -23,14 +24,19 @@ export interface LuaPluginImport {
     [option: string]: any;
 }
 
+export interface InMemoryLuaPlugin {
+    plugin: Plugin | ((options: Record<string, any>) => Plugin);
+    [option: string]: any;
```
Perryvw, line 29: Do these options and the `((options: Record<string, any>) => Plugin)` really add anything? Since the plugin is in-memory, it can just capture options from its scope, I don't see what first passing the options into tstl and then back via a factory function actually adds.

iced-wav: I think it's mostly useful for plugin compatibility, this way you can load third-party plugins directly in memory even if they were written for `tsconfig` the same way and without manually loading it separately first.

It also makes internal representation slightly simpler by separating the responsibility of loading a plugin from importing it, and don't end up with two different kinds of plugin factories.

If removing it, I think the best approach is to remove the callback version (from `InMemoryLuaPlugin`) entirely, without the options parameter it just needlessly adds another plugin factory type to handle.

Perryvw: Fair enough, that makes sense

Perryvw (approved):
Looks great, thanks!

PR #1640, opened 2025-06-06
Fix continue in loops sometimes generating dead code

### LINKED ISSUE #1638: TSTL generates unrechable code when converts 'continue' statement on Lua 5.1
Here's example of TS code:
```ts
while (true) continue;
```

Here's generated code:
```lua
while true do
    do
        local __continue2
        repeat
            __continue2 = true
            break
            __continue2 = true
        until true
        if not __continue2 then
            break
        end
    end
end
```

TSTL generated unneeded `__continue2 = true` after `break` statement.

Perryvw:
Fixes #1638


inklesspen1rus:
I love you! ♥️
As a user, surely.

PR #1648, opened 2025-07-12
fix: resolve #1647

### LINKED ISSUE #1647: Bug Report: Dynamic property method calls in @noSelf classes incorrectly use colon syntax (:) instead of dot syntax (.)
```ts
/** @noSelf **/
declare class testClass {
  static [key: string]: () => void;
  static testFunction(): void;
}


testClass.testFunction();
testClass.testKey();
```
```lua
testClass.testFunction()
testClass:testKey() -- Incorrect
```

XiYang6666:


Zamiell:
I think you need to add a failing test (that fails without your new code addition).

Perryvw (approved):
Thanks! looks good

PR #1650, opened 2025-07-15
fix: only parse arguments from first line of tag comment

aabajyan:
I had this issue, where I was trying to use customName annotation to rename methods and I noticed that it was also including a new line and the first word of the next line. For example:
```ts
/** @noSelf */
declare namespace TestNamespace {
    /**
     * @customName pass
     * The first word should not be included.
     **/
    function fail(): void;
}

TestNamespace.fail();
```
Would result in:
```ts
TestNamespace["pass\nThe"]()
```

The solution was pretty simple, I only had to get the first line before splitting it by space.
I ran the tests and everything passes. Let me know if there's other changes I need to make.


Zamiell:
nice work!

Issue #1502, opened 2023-10-17
Support recompile lualib with custom plugin?


pilaoda:
Here is the case, I want to implement a proxy for array in luajit environment with `__index` and `__newindex` that read write data in another table.
However, luajit which is lua5.1 doesn't support `__len` `__ipair` metamethod, and length operator `#`, global next function, global ipairs function would return wrong result on proxy.
Then I come up with an idea that write a printer plugin that print  `#` `next` `ipairs` into some custom implementation, but this plugin would ignore code in lualib.
I think the problem is, the plugin doesn't give user enough power to modify the compilation process, since lualib are already compiled before plugin taking over.

Z3rio:
Hello, I'm sadly having some trouble understanding what you are trying to achieve.

Would it perhaps be possible for you to send a code snippet explaining it? Or a playground link.
Or try to explain it in some other way, I guess

Thank you

Perryvw:
Sounds like this is a duplicate of #973, I still think this would be a good idea to do. However, the design is not yet clear to me, and we need to decide on what the interface for this should be.

pilaoda:
Yes this is a duplicate, and sorry I havn't check it.
Currently, lualib are already precompiled into lua code and package in the npm package. 
On user compile time, lualib just skip the compiling and use the precompiled code, and I know it is for performance reason.

What I am hoping is, there is an option to choose using precompiled lualib or not, so that the lualib code can be recompile with user's plugin if user choose not to use precompiled code.

For example, we have code `array.length` compiled into `#array`. Now I have a plugin modify it into `Len(array)`, and I implement this global `Len` function so I can call the `__len` metamethod if it exist. 
Thus, those `#` length operator in lualib_bundle.lua, should also be recompiled to use `Len` since this is user expected.

Issue #1478, opened 2023-08-25
Consider using xpcall in try block


pilaoda:
Error message from pcall only include the first line that error happened but not the full stack traceback, but this can be done by compiling to using `xpcall` and pass `debug.traceback` as second arg:
```
local result
local function ____catch(err)
    print("catch:", err)
end
local ____try, ____hasReturned = xpcall(
    function()
        result = dosomething(args)
    end,
    debug.traceback
)
if not ____try then
    ____catch(____hasReturned)
end
```

Although custom plugin can be used to implement this feature, but plugin can not rewrite lualib such as pcall in Promise.lua

lolleko:
Unfortunately, it is not as simple as replacing `pcall` with `xpcall` & `debug.traceback`.
Because using `xpcall` & `debug.traceback` will break our `throw` implementation.

e.g.

```
      try {
          throw "Hello, I am an Error!";
      } catch (error) {
          return error;
      }
```        

would return:

```
Hello, I am an Error!
    stack traceback:
        [C]: in function 'error'
        [string \"...\"]:17: in function <[string \"...\"]:16>
        [C]: in function 'xpcall'
        [string \"...\"]:15: in function <[string \"...\"]:6>
        (...tail calls...)
        [string \"...\"]:3: in main chunk
```

instead of the correct/JS behavior
```
Hello, I am an Error!
```

This can possibly be solved by changing the way we throw string literals. Another solution might be to create a custom/adapted implementation of debug.tracepack.

BribeFromTheHive:
Likely the best middle ground is that an empty catch block can compile to an xpcall, and a non-empty one can compile as-is. Same with un-caught rejected promise objects, if they don't already work that way.

pilaoda:
https://github.com/TypeScriptToLua/TypeScriptToLua/blob/16a8eb3a5aed582f31d5f2306d70d3801983deca/src/lualib/Await.ts#L55-L58
`coroutine.resume` also lack of stack trace, the `asyncCoroutine` should pass into `debug.traceback` as the first argument to get the error trace in coroutine context.

fadabaerobe6:
```lua
-- Reimplemented pcall function
_G.pcall = function(func, ...)
    local results = table.pack(
        xpcall(func, function(err)
            -- Do nothing here, just return the error object
            return err..'\n'..debug.traceback()
        end, ...)
    )
    local status = table.remove(results, 1)
    if not status then
[...]

PR #1656, opened 2025-08-23
New option recompileLuaLib: Recompile the Lua standard library with custom plugins.

pilaoda:
Implementation about #1478 #1502
all `luaLibImport` options are supported.

### DIFF
--- a/test/transpile/lualib.spec.ts
+++ b/test/transpile/lualib.spec.ts
@@ -31,7 +31,7 @@ test("Lualib bundle does not assign globals", () => {
         .expectNoExecutionError();
 });
 
-test("Lualib bundle recompile", () => {
+test("Lualib recompile with import kind require", () => {
     const { transpiledFiles } = util.testExpression`Array.isArray({})`
         .setOptions({
             recompileLuaLib: true,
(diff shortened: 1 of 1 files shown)


src/CompilerOptions.ts
```diff
@@ -20,12 +20,15 @@ export interface TransformerImport {
 export interface LuaPluginImport {
     name: string;
     import?: string;
+    skipRecompileLuaLib?: boolean;
```
Perryvw, line 23: Why is this needed? Couldn't the plugin itself decide to skip files if desired?

src/LuaLib.ts
```diff
@@ -175,7 +182,21 @@ export function getLuaLibExportToFeatureMap(
 
 const lualibFeatureCache = new Map<LuaTarget, Map<LuaLibFeature, string>>();
 
-export function readLuaLibFeature(feature: LuaLibFeature, luaTarget: LuaTarget, emitHost: EmitHost): string {
+export function readLuaLibFeature(
+    feature: LuaLibFeature,
+    luaTarget: LuaTarget,
+    emitHost: EmitHost,
+    useCache = true
```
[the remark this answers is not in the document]
Perryvw: I still don't understand this: Either we read the file from disk and we want to cache the data we read, or we recompile and we want to cache the compilation result. Why would we ever want to skip the cache in either case? I don't see why the lualib would change after first compilation within a single tstl run

test/transpile/lualib.spec.ts
```diff
@@ -29,3 +30,39 @@ test("Lualib bundle does not assign globals", () => {
...
+});
+
+test("Lualib require minimal recompile", () => {
+    util.testExpression`Array.isArray(123)`
+        .setOptions({
+            luaLibImport: LuaLibImportKind.RequireMinimal,
+            recompileLuaLib: true,
+            luaPlugins: [{ name: path.join(__dirname, "./plugins/visitor.ts") }],
+        })
+        .expectToHaveNoDiagnostics()
```
Perryvw, line 55: Could you add an assertion that the lualib file (s) are output at the expected location?

test/transpile/lualib.spec.ts
```diff
@@ -29,3 +30,39 @@ test("Lualib bundle does not assign globals", () => {
...
+test("Lualib bundle recompile", () => {
+    const { transpiledFiles } = util.testExpression`Array.isArray({})`
+        .setOptions({
+            recompileLuaLib: true,
+            luaPlugins: [{ name: path.join(__dirname, "./plugins/beforeEmit.ts") }],
+        })
+        .expectToHaveNoDiagnostics()
+        .getLuaResult();
+    const lualubBundle = transpiledFiles.find(f => path.basename(f.outPath) === "lualib_bundle.lua");
+    expect(lualubBundle?.lua).toContain(
```
Perryvw, line 43: Could you also add a test with bundling?

src/transpilation/transpile.ts
```diff
@@ -145,9 +145,16 @@ export function getProgramTranspileResult(
         transpiledFiles = [];
     }
 
+    const proxyEmitHost =
```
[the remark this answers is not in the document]
Perryvw: Couldn't you solve this in lualib.ts where you start the new transpile (with this explanation as comment because code is very hard to understand), instead of here in the global transpile process?

PR #1662, opened 2025-10-12
Add additional options for emitting lualib bundle

MCJack123:
This PR adds two new configuration options, which control how the lualib bundle is saved to disk:
- `luaLibName`: Sets the name to use for the lualib bundle. For example, I find the name `lualib_bundle` to be vague, so I usually change it to `typescript` to make it clear what the library is for.
- `luaLibEmit`: Setting this to `false` prevents the lualib bundle from being emitted. This is useful to avoid storing lualib in bundled code when it's available outside the bundle already, or simply to reduce generated file bloat.

These options have been useful for my environment, where I have a multitude of common libraries in a store made available system-wide. Being able to change the name allows me to use a more descriptive file name in that store, and being able to suppress its generation means I don't have the duplicated file dangling around in bundles or the build directory.

Neither of these options can be accomplished through plugins. Changing the name in the generated code could be done with `moduleResolution` (I haven't tried), but it would not change the generated file's name on disk or in a bundle, and would potentially break other parts of the code. Suppressing emission can't be done for a bundle, since the file is incorporated between `afterPrint` and `beforeEmit` with no chance to intervene.

These changes will not affect users who do not use them; all default settings match how they were before.

### DIFF
--- a/src/cli/parse.ts
+++ b/src/cli/parse.ts
@@ -58,7 +58,8 @@ export const optionDeclarations: CommandLineOption[] = [
     },
     {
         name: "luaLibEmit",
-        description: "Whether the Lua library file is emitted with the code. Only effective when luaLibImport is require.",
+        description:
+            "Whether the Lua library file is emitted with the code. Only effective when luaLibImport is require.",
         type: "boolean",
     },
     {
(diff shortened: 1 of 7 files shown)


MCJack123:
> Regarding `luaLibEmit` I would have thought that setting `lualibImport` to `None` would prevent the file from being emitted.

That would prevent the lualib from being imported entirely, i.e. there would be no `require` for it at all, which I want to keep in my applications. All I want is to maintain the `require` calls without writing the lualib to a bundled output.

Perryvw:
This seems like a very specific use-case I would rather fix with the use of our existing plugin infrastructure rather than pulling this logic and configuration complexity into tstl core. I'm currently lacking sufficient motivation to want to merge something like this.

Issue #1665, opened 2025-10-28
debug.traceback failed on anonymous function


ChouUn:
https://github.com/TypeScriptToLua/TypeScriptToLua/blob/1ed3b4d5f9968e821d10c9e6742c3b1025561495/src/lualib/SourceMapTraceBack.ts#L52

When anonymous functions are included in TS code, stack.traceback printing doesn't work exactly as expected, such as in the form `SOME_PATH\dist\main.ts:43: in function <SOME_PATH\dist\main.lua:65>`.
This is because the "<" escapes as a non-whitespace character matching the `%S`, causing it to be considered part of the path.

On Windows, you can quickly fix this by changing the pattern to `([^%s<]+)%.lua:(%d+)`.
This isn't the most elegant fix, considering that "<" can be part of a filename on Unix-like systems.

Additional, my guess is that if there are spaces in the entire path, it may also cause unexpected behavior because of `%S`.

However, collecting the paths to build a regular expression can be quite risky.
Do you have any suggestions?

Perryvw:
Which Lua runtime are you using? In my quick check I don't see the `<>` notation used in Lua 5.3 or 5.4.

And do you have some example code that actually produces the problem? I've been using this to test:
```lua
function bla()
  local f = function()
    error("hi")
  end
  f()
end

bla()
```

PR #1666, opened 2025-10-28
fix `debug.traceback` failing on anonymous function (#1665)

ChouUn:
more details in #1665

### DIFF
--- a/test/unit/error.spec.ts
+++ b/test/unit/error.spec.ts
@@ -362,6 +362,13 @@ test("sourceMapTraceback maps anonymous function locations in .lua files (#1665)
         "\t[C]: in ?",
     ].join("\n");
 
+    // Inject sourcemap for "main.lua" and mock debug.traceback to return file-based frames.
+    const header = `
+        __TS__sourcemap = { ["main.lua"] = ${mapping} };
+        local __real_tb = debug.traceback
+        debug.traceback = function() return ${JSON.stringify(fakeTraceback)} end
+    `;
+
     const builder = util.testFunction`
         return (() => {
             return (() => {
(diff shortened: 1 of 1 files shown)


test/unit/error.spec.ts
```diff
@@ -345,6 +345,47 @@ test("still works without debug module", () => {
...
+
+    const result = util.testFunction`
+        return (() => {
+            return (() => {
+                return debug.traceback();
+            })();
+        })();
+    `
+        .setLuaHeader(
+            `__TS__sourcemap = { ["main.lua"] = {["3"] = 7, ["4"] = 8, ["5"] = 9, ["6"] = 8} }\n` +
```
Perryvw: Why are you overriding the source map here? Shouldn't tstl already produce the sourcemap due to the sourcmaptraceback below?

Perryvw: Both line 4 and 6 map to line 8 here, which seems strange. If you are overriding the default traceback (which I would rather not do unless absolutely necessary), better pick some different numbers so we can assert that both <main.lua:4> and \tmain.lua:6: are mapped to the correct values.

test/unit/error.spec.ts
```diff
@@ -345,6 +345,47 @@ test("still works without debug module", () => {
...
+                `debug.traceback = function() return ${JSON.stringify(fakeTraceback)} end`
+        )
+        .setOptions({ sourceMapTraceback: true })
+        .getLuaExecutionResult();
+
+    // Regular line references should be mapped
+    expect(result).toContain("main.ts:9");
+    expect(result).toContain("main.ts:8");
+    // Anonymous function definitions <main.lua:4> and <main.lua:3> should also be mapped
+    expect(result).not.toContain("main.lua");
```
Perryvw: I would prefer if we explicitly assert `expect(result).toContain("<main.ts:7>")` and `expect(result).toContain("<main.ts:8>")`

PR #1678, opened 2026-01-22
fix(bundle): prevent tail-call in entry module for proper stack trace

dnrovs:


Perryvw:
I don't quite understand what this is trying to solve. Could you provide an example stack trace with and without your fix?

PR #1492, opened 2023-09-25
Number static data properties

Z3rio:
**This is not done, please do not merge yet. Tests will be added once all data properties have been added.**

This PR aims to provide all the data properties from JS's Number library.

Closes: https://github.com/TypeScriptToLua/TypeScriptToLua/issues/1108

Progress:
- ✅NEGATIVE_INFINITY
- ✅POSITIVE_INFINITY
- ✅NaN
- ✅MIN_VALUE
- ✅MIN_SAFE_INTEGER
- ✅MAX_VALUE
- ✅MAX_SAFE_INTEGER
- ✅EPSILON

PR #1698, opened 2026-03-26
fix Number constants

RealColdFry:
For reference, the correct ordering is: `NEGATIVE_INFINITY < MIN_SAFE_INTEGER < 0 < MIN_VALUE < EPSILON < MAX_SAFE_INTEGER < MAX_VALUE < POSITIVE_INFINITY`

All four `Number` constants emit wrong Lua: they all end up as `inf` or `-inf`

| Constant | Before (wrong) | After (correct) |
|---|---|---|
| `MAX_SAFE_INTEGER` | `2 ^ 1024` → `inf` | `2 ^ 53 - 1` |
| `MIN_SAFE_INTEGER` | `-2 ^ 1074` → `-inf` | `-(2 ^ 53 - 1)` |
| `MIN_VALUE` | `-2 ^ 1074` → `-inf` | `2 ^ -1074` |
| `MAX_VALUE` | `2 ^ 1024` → `inf` | `(2 - 2 ^ -52) * 2 ^ 1023` |

Also `MIN_SAFE_INTEGER` and `MIN_VALUE` were emitting identical code, same for `MAX_SAFE_INTEGER` and `MAX_VALUE`.

You can verify with:

```ts
console.log("MAX_SAFE_INTEGER:", Number.MAX_SAFE_INTEGER);  // 9007199254740991
console.log("MIN_SAFE_INTEGER:", Number.MIN_SAFE_INTEGER);  // -9007199254740991
console.log("MIN_VALUE:", Number.MIN_VALUE);                // 5e-324
console.log("MAX_VALUE:", Number.MAX_VALUE);                // 1.7976931348623157e+308
```

See [Number on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number) for the correct values.

Repro in the TSTL playground: https://typescripttolua.github.io/play/#code/5.4/MYewdgziA2CmB00QHMAUAiAsgQQBoH0BlbAMQFF8BJAOQBUyBxMgJQC50AaAAmoFcBbAEawATvBwFi5KnUYsAlAG4uXAPSquATgAM2gOwBGTZoBMAVgAsei9uMGAUKEgwESNFhpFSFGvSZtOHgFhMUxPKR9ZfyUVdS4AWh19I1NLa1tNBycoOEQUDDDqfAA1bAAZAFUydm4+IVFxT1LKshiVdo7OtQ0zWHiAZhMLR3Ac13ysPBLyqpqg+tCp5qq2rrXurgN4PU09ADZNfoN+iwAOPZMjsz1YAGp+7VOgA

The existing test didn't catch this because it used `expectToEqual(true)` with comparisons that the wrong values accidentally satisfied (e.g. `-inf <= -inf` is `true`). Switched to `expectToMatchJsResult` with stricter comparisons that actually verify the ordering and specific values.

### DIFF
--- a/src/transformation/builtins/number.ts
+++ b/src/transformation/builtins/number.ts
@@ -77,31 +77,62 @@ export function transformNumberProperty(
                 node
             );
         case "MIN_VALUE":
+            // 2 ^ -1074 = 5e-324 (smallest positive double)
             return lua.createBinaryExpression(
-                lua.createNumericLiteral(-2),
-                lua.createNumericLiteral(1074),
+                lua.createNumericLiteral(2),
+                lua.createNumericLiteral(-1074),
                 lua.SyntaxKind.PowerOperator,
                 node
             );
         case "MIN_SAFE_INTEGER":
-            return lua.createBinaryExpression(
-                lua.createNumericLiteral(-2),
-                lua.createNumericLiteral(1074),
-                lua.SyntaxKind.PowerOperator,
+            // -(2 ^ 53 - 1) = -9007199254740991
+            return lua.createUnaryExpression(
+                lua.createParenthesizedExpression(
+                    lua.createBinaryExpression(
+                        lua.createBinaryExpression(
+                            lua.createNumericLiteral(2),
+                            lua.createNumericLiteral(53),
+                            lua.SyntaxKind.PowerOperator
+                        ),
+                        lua.createNumericLiteral(1),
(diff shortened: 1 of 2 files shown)


lolleko (approved):
Nice good catches i think this has been broken since we added these constants in #1492
Switching to `expectToMatchJsResult` in tests gives me confidence that everything is correct.

In theory this breaks API / behvaiour, in this case that is warranted though.

PR #1701, opened 2026-03-30
rough ts6 fixes to make ts happy

RealColdFry:
Happy to rework any of this.

Upgrade TypeScript from 5.9.3 to 6.0.2, along with `typescript-eslint` 8.58.0 and `ts-jest` 29.4.9 so they have peer deps on TS 6.0.2

## Not included here

- `Map.getOrInsert` / `getOrInsertComputed` lualib support. Node runtime doesn't have these yet so `expectToMatchJsResult` in tests cannot cover them.
- Resolver support for `paths` inherited from extended tsconfigs without `baseUrl`. Supporting that correctly means resolving each `paths` entry relative to the tsconfig where it is defined, not just the leaf config.


RealColdFry:
things to check, all checked.

- [x] restore pre-compiled picomatch matchers, currently recompiling per call which is a perf regression
- [x] duplicate test name "in function parameter" in destructuring tests, rename one
- [x] verify paths-simple snapshot change, output path changed from `dist/main.lua` to `dist/myprogram/main.lua`
- [x] strictNullChecks: false additions in a few tests, might be masking real bugs
- [x] tsconfig swaps experimentalDecorators for moduleResolution bundler, does the source still build with module commonjs?
- [x] tsconfig.eslint.json only includes src but eslint lints tests too
- [x] comment on the overrides block explaining typescript-eslint doesn't support ts 6.0 yet
- [x] comment on ignoreDiagnostics([2783]) explaining the ts 6.0 change

PR #1702, opened 2026-03-30
fix dots in file/directory names breaking Lua require resolution (#1445)

### LINKED ISSUE #1445: Dots in module path breaks Lua import in a non-bundle compilation
### Abstract
`tstl` fails to produce valid `lua` code from a subset of valid (accordingly to `tstl` docs) code if `ts` file names contain dots in paths.

This problem may naturally arise, for example, when compiling from Purescript to JS and then from JS to Lua using tstl.

The example is not the natural use case for tstl, however, the bug is the bug.
### Example
Let's say we have two files:
`./src/Foo.Bar/index.ts`
```ts
export const answer: number = 42;
```
`./src/Qux/index.ts`
```ts
import {answer} from "./Foo.Bar";
export const printAnswer = () => print(answer)
```
So `import {answer} from "./Foo.Bar";` will be compiled (in a non-bundle mode) to
```lua
local ____Foo_2EBar = require("Foo.Bar.index")
local answer = ____Foo_2EBar.answer
```
and `lua` will be searching at `./Foo/Bar/index.lua"` for that module.

That will result an error:
```
lua: ./Qux/index.lua:3: module 'Foo.Bar.index' not found:
        no field package.preload['Foo.Bar.index']
        no file './Foo/Bar/index.lua'
        no file '/usr/local/share/lua/5.1/Foo/Bar/index.lua'
        no file '/usr/local/share/lua/5.1/Foo/Bar/index/init.lua'
        no file '/usr/local/lib/lua/5.1/Foo/Bar/index.lua'
        no file '/usr/local/lib/lua/5.1/Foo/Bar/index/init.lua'
        no file '/usr/share/lua/5.1/Foo/Bar/index.lua'
        no file '/usr/share/lua/5.1/Foo/Bar/index/init.lua'
        no file './Foo/Bar/index.so'
        no file '/usr/local/lib/lua/5.1/Foo/Bar/index.so'
        no file '/usr/lib/x86_64-linux-gnu/lua/5.1/Foo/Bar/index.so'
        no file '/usr/lib/lua/5.1/Foo/Bar/index.so'
        no file '/usr/local/lib/lua/5.1/loadall.so'
        no file './Foo.so'
        no file '/usr/local/lib/lua/5.1/Foo.so'
        no file '/usr/lib/x86_64-linux-gnu/lua/5.1/Foo.so'
        no file '/usr/lib/lua/5.1/Foo.so'
        no file '/usr/local/lib/lua/5.1/loadall.so'
stack traceback:
        [C]: in function 'require'
        ./Qux/index.lua:3: in main chunk
        [C]: in function 'require'
        ./main.lua:3: in main chunk
        [C]: ?
```
### Soulution
When compiled to a file `./dist/Foo.Bar.Baz/index.lua`, split it's name by dots (excluding extension) and put the file into directory `./Foo/Bar/Baz`.
This will fix module resolution for `lua`.
So basically the ts directory
```ls
.
├── Foo/
│   └── index.ts
├── Foo.Bar/
│   └── index.ts
├── Foo.Bar.Baz/
│   └── index.ts
├── Qux/
│   └── index.ts
└── index.ts
```
should be compiled into
```ls
.
├── Foo/
│   ├── index.lua
│   └── Bar/
│       └── index.lua
│       └── Baz/
│           └── index.lua
├── Qux/
│   └── index.lua
└── index.lua
```
instead of
```ls
.
├── Foo/
│   └── index.lua
├── Foo.Bar/
│   └── index.lua
├── Foo.Bar.Baz/
│   └── index.lua
├── Qux/
│   └── index.lua
└── index.lua
```

RealColdFry:
Fixes #1445 

Lua's require() uses dots as path separators, so a file at Foo.Bar/index.lua is unreachable via require("Foo.Bar.index") since Lua looks for Foo/Bar/index.lua. Expand dotted path segments into nested directories in the emit output, and emit a diagnostic when this expansion causes output path collisions.

### DIFF
--- a/test/unit/modules/resolution.spec.ts
+++ b/test/unit/modules/resolution.spec.ts
@@ -1,5 +1,5 @@
 import * as ts from "typescript";
-import { couldNotResolveRequire } from "../../../src/transpilation/diagnostics";
+import { couldNotResolveRequire, emitPathCollision } from "../../../src/transpilation/diagnostics";
 import * as util from "../../util";
 
 const requireRegex = /require\("(.*?)"\)/;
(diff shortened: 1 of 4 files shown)


Perryvw:
While this is a possible solution, wouldn't it be simpler (and the output more reasonable) if we were to just find/replace periods with e.g. underscores in the output file names?

RealColdFry:
> While this is a possible solution, wouldn't it be simpler (and the output more reasonable) if we were to just find/replace periods with e.g. underscores in the output file names?

Agreed, done

test/unit/modules/resolution.spec.ts
```diff
@@ -166,6 +167,37 @@ test.each([
         .tap(expectToRequire(expectedPath));
 });
 
+// https://github.com/TypeScriptToLua/TypeScriptToLua/issues/1445
+// Can't test this via execution because the test harness uses package.preload
+// instead of real filesystem resolution, so require() always finds the module
+// regardless of output path. We check the output path directly instead.
+test("dots in directory names are replaced with underscores in output", () => {
+    const { transpiledFiles } = util.testModule`
+        import { answer } from "./Foo.Bar";
```
Perryvw, line 176: Could you also add a test where you're including a file like "Foo.Bar.ts", or maybe "foo.test.ts" since that was mentioned in the issue, rather than only directories with dots

RealColdFry: Added: https://github.com/TypeScriptToLua/TypeScriptToLua/pull/1702/changes/c43850140f254180771e702e5c664cece68eba91#diff-90683c66d20ef437879beec3960e869dfb418e44caf0bcccffa005c8f64bafceR189

PR #1703, opened 2026-04-01
fix >>> (unsigned right shift) producing wrong results on Lua 5.3+

RealColdFry:
https://typescripttolua.github.io/play/#code/5.4/MYewdgzgLgBGMF4YFoCsBuGXs93-BhRxJpZ+AUKJCADYCmAdLSAOYAU8AfDzAAwBKTFgD0ImABYATAE4JMgGwB2WQEZyGzVu2Eq4CHSYsO6nl35DsYmOp137DjXpoNmbdslO9Bw69LmKKjKoQA

TS >>> is a logical right shift (zero-fills from the left), but TSTL was mapping it to Lua's >> which is arithmetic (sign-extends). This gave wrong results for negative numbers: e.g. -1 >>> 0 should be 4294967295 but produced -1.

Fix by masking to unsigned 32-bit first: (left & 0xFFFFFFFF) >> right. The Lua 5.2 (bit32.rshift) and LuaJIT (bit.rshift) paths were already correct.

```typescript
const n = -5;                                                                                                                
console.log(n >>> 0);   // 4294967291                                                                                              
console.log(1 >>> 0);   // 1                                                                                                       
console.log(-1 >>> 0);  // 4294967295
```

### DIFF
--- a/src/transformation/visitors/binary-expression/bit.ts
+++ b/src/transformation/visitors/binary-expression/bit.ts
@@ -75,6 +75,22 @@ export function transformBinaryBitOperation(
         case LuaTarget.Lua52:
             return transformBinaryBitLibOperation(node, left, right, operator, "bit32");
         default:
+            // Lua 5.3+ `>>` is arithmetic (sign-extending), but TS `>>>` is logical (zero-fill).
+            // Emit `(left & 0xFFFFFFFF) >> right` to convert to unsigned 32-bit first.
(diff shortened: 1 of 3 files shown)


rhazarian:
@RealColdFry @Perryvw this breaks >>> operator for Lua 5.3 32-bit target which doesn't have unsigned integers and cannot handle 4294967295 (which is now emitted) as an integer

PR #1704, opened 2026-04-03
Upgrade to TypeScript 6.0

Perryvw:
50 failing tests remaining, they still need some careful looking into.

Took many test fixes from #1701, all ones I'd be willing to merge. Might be more in there for remaining failing tests, but ran out of time for now.


test/transpile/module-resolution.spec.ts
```diff
@@ -472,6 +472,7 @@ describe("module resolution should not try to resolve modules in noResolvePaths"
                 }`
             )
             .setOptions({ noResolvePaths: ["ignore*"] })
+            .debug()
```
RealColdFry:
```suggestion
```

src/transpilation/resolve.ts
```diff
@@ -215,14 +213,16 @@ class ResolutionContext {
         const fileFromPath = this.getFileFromPath(resolvedPath);
         if (fileFromPath) return fileFromPath;
 
-        if (this.options.paths && this.options.baseUrl) {
+        if (this.options.paths) {
```
RealColdFry, line 216: Ordering diverges from TS 6: project file lookup (like a sibling file) can win over a `paths`-mapped target.

Ref: https://www.typescriptlang.org/docs/handbook/modules/reference.html#paths
> TypeScript offers a way to override the compiler’s module resolution for bare specifiers with the `paths` compiler option.

---

Minimal failing repro at RealColdFry/TypeScriptToLua@e43097e871aa4c6efb7a903e556bbb4cd13699b8 with [CI test result](https://github.com/TypeScriptToLua/TypeScriptToLua/actions/runs/24705430888/job/72257569793#step:6:254)

test/transpile/module-resolution.spec.ts
```diff
@@ -447,7 +446,7 @@ describe("module resolution should not try to resolve modules in noResolvePaths"
             .expectToHaveNoDiagnostics();
     });
 
-    test("can ignore specific files with glob pattern", () => {
+    test.only("can ignore specific files with glob pattern", () => {
```
RealColdFry:
```suggestion
    test("can ignore specific files with glob pattern", () => {
```

RealColdFry:
LGTM modulo module resolution comments.

Awaiting CI but unit tests on my local branch also passed, including after resolving merge conflicts with master. SKimmed other transformer changes but test suite covers so comfortable approving.

src/transpilation/resolve.ts
RealColdFry: Actually, it's not clear to me that this is a TS 6.0 change from the release notes. I suppose paths without baseUrl now works so this might get exercised more, but can probably ignore / can be added separately.

PR #1705, opened 2026-04-05
fix Number(), parseFloat, parseInt not handling some string formats

RealColdFry:
not sure number parsing code should be grown though


RealColdFry:
I've moved this back to draft because there isn't any filed issue, and it does grow the lualib too

RealColdFry:
closing for now -- can resurrect/reference later if this comes up later

PR #1707, opened 2026-04-06
fix return, break and continue inside try in async functions (#1706)

### LINKED ISSUE #1706: `return`, `break` and `continue` inside `try` in async functions don't work correctly
### Description

`return`, `break` and `continue` statements inside `try` blocks in `async` functions don't preserve TypeScript semantics:

1. **`return` inside `try`** — code after the `try/catch` block still executes (should be unreachable)
2. **`break` inside `try` in a loop** — Lua load-time error: `<break> not inside a loop`
3. **`continue` inside `try` in a loop** — Lua load-time error: `no visible label`

### Reproduction: `return` inside `try`

```typescript
let resolveLater!: (value: string) => void;

function deferredPromise(): Promise<string> {
    return new Promise(resolve => {
        resolveLater = (v) => resolve(v);
    });
}

async function fn(): Promise<string> {
    try {
        return await deferredPromise();
    } catch {
        return 'caught';
    }
    print('unreachable!');
}

(async () => {
    const promise = fn();
    resolveLater('ok');
    print(await promise);
})();
```

**Expected:** `print('unreachable!')` is dead code — both `try` and `catch` branches return.

**Actual:** `print('unreachable!')` executes.

**Note:** the `return` issue only manifests when the awaited promise resolves asynchronously. With synchronously resolved promises (`Promise.resolve()`) the bug does not reproduce.

`return` in a loop (same deferred promise issue — exits only the nested `__TS__AsyncAwaiter`, loop continues):

```typescript
async function fn(): Promise<string> {
    while (true) {
        try {
            return await deferredPromise();
        } catch {
            return 'caught';
        }
        print('unreachable!');
    }
}
```

### Reproduction: `break` and `continue` inside `try` in a loop

`break`:

```typescript
async function fn(): Promise<void> {
    while (true) {
        try {
            await Promise.resolve();
            break;
        } catch {}
    }
}
```

**Expected:** breaks out of the loop.  
**Actual:** Lua load-time error: `<break> at line ... not inside a loop`

`continue`:

```typescript
async function fn(): Promise<void> {
    for (let i = 0; i < 3; i++) {
        try {
            await Promise.resolve();
            if (i === 1) continue;
        } catch {}
        print(i);
    }
}
```

**Expected:** prints `0` and `2`, skips `1`.  
**Actual:** Lua load-time error: `no visible label '__continue' for <goto>`

### Root cause

TSTL compiles `try` into a nested `__TS__AsyncAwaiter` (simplified):

```lua
local ____try = __TS__AsyncAwaiter(function()
    return ____awaiter_resolve(nil, __TS__Await(deferredPromise()))
end)
__TS__Await(____try.catch(____try, function(____)
    return ____awaiter_resolve(nil, "caught")
end))
-- code below executes unconditionally after ____try completes
print("unreachable!")
```

- `return ____awaiter_resolve(...)` resolves the outer function's promise but only exits the inner `__TS__AsyncAwaiter` coroutine — it does not terminate the outer async function
- After `__TS__Await(____try.catch(...))` the outer function continues executing (resolve + continue)
- `break`/`continue` are emitted inside the inner coroutine function where no enclosing loop or label exists, producing invalid Lua at load time

### Workaround

Avoid `return`, `break` and `continue` inside `try`/`catch` blocks in async functions. Use variable assignment inside `try`/`catch`, then control flow after the block:

```typescript
// return workaround
async function fn(): Promise<string> {
    let result: string;
    try {
        result = await deferredPromise();
    } catch {
        result = 'caught';
    }
    return result;
}

// break workaround
async function fn2(): Promise<void> {
    let done = false;
    while (!done) {
        try {
            await doWork();
            done = true;
        } catch {}
    }
}
```

### Environment

- TSTL version: 1.34.0
- Lua target: 5.4

RealColdFry:
Fixes #1706 

* replaces return/break/continue inside async try/catch with flag variables checked after the awaiter resolves              
* nested try-in-try not yet handled

### DIFF
--- a/src/transformation/visitors/return.ts
+++ b/src/transformation/visitors/return.ts
@@ -3,7 +3,7 @@ import * as lua from "../../LuaAST";
 import { FunctionVisitor, TransformationContext } from "../context";
 import { validateAssignment } from "../utils/assignment-validation";
 import { createUnpackCall, wrapInTable } from "../utils/lua-ast";
-import { ScopeType, walkScopesUp } from "../utils/scope";
+import { findAsyncTryScopeInStack, ScopeType, walkScopesUp } from "../utils/scope";
 import { transformArguments } from "./call";
 import {
     returnsMultiType,
@@ -68,6 +68,8 @@ export function transformExpressionBodyToReturnStatement(
 }
 
 export const transformReturnStatement: FunctionVisitor<ts.ReturnStatement> = (statement, context) => {
+    const asyncTryScope = isInAsyncFunction(statement) ? findAsyncTryScopeInStack(context) : undefined;
+
     let results: lua.Expression[];
 
     if (statement.expression) {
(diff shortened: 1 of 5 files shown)


src/transformation/visitors/return.ts
```diff
@@ -77,12 +79,35 @@ export const transformReturnStatement: FunctionVisitor<ts.ReturnStatement> = (st
...
+        asyncTryScope.asyncTryHasReturn = true;
+        const stmts: lua.Statement[] = [
+            lua.createAssignmentStatement(
+                lua.createIdentifier("____hasReturned"),
+                lua.createBooleanLiteral(true),
+                statement
+            ),
+        ];
+        if (results.length > 0) {
+            stmts.push(lua.createAssignmentStatement(lua.createIdentifier("____returnValue"), results[0], statement));
```
Perryvw: Guess this breaks for multi-return, any easy wayt to serve those cases? (otherwise I'm okay with merging the easier change first)

RealColdFry: multi-return actuallyy works here [20b205dc](https://github.com/TypeScriptToLua/TypeScriptToLua/pull/1707/changes/20b205dc0598ee5f870c7ae506823d8011b07bc5) because the multi-return values are wrapped into a single table before storage so `results[0]` captures everything. But it was somewhat confusing and relied on implicit assumptions so I split out  [`transformReturnExpressionForTryCatch`](https://github.com/TypeScriptToLua/TypeScriptToLua/pull/1707/changes#diff-ebcbe1444b4c8adbaffda2e3e41fb52bf7779b7202ea94f017914e413cf59a72R53) which returns a single `lua.Expression` instead of indexing into an array.

PR #1709, opened 2026-04-15
fix await inside catch/finally in async try (#1659)

### LINKED ISSUE #1659: Await should check in coroutine before yield
https://typescripttolua.github.io/play/#code/5.4/FAGwpgLgBADgTgewLYEsDOYBKYBWYDGEAXFAIYB2AnlALxQCu5AJmAGYrlhMDcwpalcviitGhFAnJQAKmDQQATAAoAlFADewKNqj5J82IlQZaUTgHcoABSPowSpXDkIQANzAAaKE7yE1NAD4NLR1Q+GQ7bF9oOh8CCF5QgF8VRJ0nCHo4KXDjMF4k4D4BIRExCAkpWXkARlVg0Ig4ak1Q0NJzUhRoasVVNO0k3VIIfAALKCUwNVa27XgOCCUAIjBllRC5vXIDUlMOrugbCIwAOic0F3clJvppgbaF8iXl0mWvUlSoAHpvszB3HBDIsAIRkCAQMBIGDQCAIKCUFBgEBMERGKAIegQNAoFhkXQIRBYjhgTZQQqFYC9OobFCsSa5SK4eIzMmMjBReKqYCFIA

We can use `await` in async `catch` block, but when execute to this `catch` block, the program may not in `coroutine` but in `main` lua thread.

Maybe we should fix it in this way?

```typescript
export function __TS__Await(this: void, thing: unknown) {
    if (!coroutine.running()) {
        return thing;
    }
    return coyield(thing);
}
```

RealColdFry:
Fixes #1659 

- wrap catch and finally callbacks in `__TS__AsyncAwaiter` so `await` works inside them
- fix the chain ordering (`.catch()` before `.finally()`) and wire up the full chain before awaiting.

### DIFF
--- a/test/unit/builtins/async-await.spec.ts
+++ b/test/unit/builtins/async-await.spec.ts
@@ -815,4 +815,92 @@ describe("try/catch in async function", () => {
             },
         });
     });
+
+    // https://github.com/TypeScriptToLua/TypeScriptToLua/issues/1659
+    test("await inside catch handler resolves correctly (#1659)", () => {
+        util.testFunction`
+            let reject: (reason: string) => void = () => {};
+
+            async function failing() {
+                return new Promise((_, rej) => { reject = rej; });
+            }
+
+            async function run() {
+                try {
+                    await failing();
+                } catch (e) {
+                    log("catch");
+                    const a = await Promise.resolve(true);
+                    log("a", a);
+                }
+            }
+
+            run();
+            reject("error");
+
+            return allLogs;
+        `
+            .setTsHeader(promiseTestLib)
+            .expectToEqual(["catch", "a", true]);
+    });
+
+    // https://github.com/TypeScriptToLua/TypeScriptToLua/issues/1659
+    test("await inside finally handler resolves correctly (#1659)", () => {
+        util.testFunction`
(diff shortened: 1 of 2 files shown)


test/unit/builtins/async-await.spec.ts
```diff
@@ -815,4 +815,92 @@ describe("try/catch in async function", () => {
...
+            async function failing() {
+                return new Promise((_, rej) => { reject = rej; });
+            }
+
+            async function run() {
+                try {
+                    await failing();
+                } catch (e) {
+                    log("catch");
+                    const a = await Promise.resolve(true);
```
Perryvw, line 833: Could you add a test where this promise is not instantly resolved, and its return value is returned to `run`? Something like:
```ts
try {
    await failing();
    return "succeeded";
} catch (e) {
    return await somePromiseResolvedLater();
}
```
and then resolving the promise after rejecting the first one

RealColdFry: Done: https://github.com/TypeScriptToLua/TypeScriptToLua/pull/1709/changes#diff-392634086752122c2ace4b89fd9640371efcc3700127844e6f941e672826f720R908

PR #1721, opened 2026-05-02
fix >> (signed right shift) on Lua 5.3+ (#1720)

### LINKED ISSUE #1720: Right shift operator not supported for target 5.4
Although the tsconfig target is 5.4, I still face the following error:

Right shift operator is not supported for target Lua 5.3. Use `>>>` instead.

```json
  "tstl": {
    "luaLibImport": "require-minimal",
    "luaBundle": "./dist/server.lua",
    "luaBundleEntry": "./src-ts/index.ts",
    "luaTarget": "5.4",
    "noImplicitSelf": true,
    "sourceMapTraceback": false
  }
```

Playground: https://typescripttolua.github.io/play#code/5.4/A4JwlgdgLgFAbAAgHxIQJgJRA

Note: LuaJIT works, but I still need LuaFloorDivision, which don't work on JIT, so I can't just change to it.

Image

RealColdFry:
Fixes #1720 

alternatively, can improve diagnostic and let users provide their own functions, which is potentially less controversial and stays truer to the "pure transpile" story.

Separately, not directly relevant:

- any scope to allow users to provide overrides for lowering/transpiling at AST leaves that could be *inlined*? something more "granular" than visitor overrides) 
- scope to fix error codes (not auto-increment via factory), show error codes in diagnostics, combined with a link to an error code listing on TSTL website describing the error in more detail?

### DIFF
--- a/test/unit/__snapshots__/expressions.spec.ts.snap
+++ b/test/unit/__snapshots__/expressions.spec.ts.snap
@@ -372,13 +372,6 @@ ____exports.__result = a << b
 return ____exports"
 `;
 
-exports[`Bitop [5.3] ("a>>=b") 1`] = `
-"local ____exports = {}
-a = (((a & 4294967295) ~ 2147483648) - 2147483648) // (1 << (b & 31))
-____exports.__result = a
-return ____exports"
-`;
-
 exports[`Bitop [5.3] ("a>>>=b") 1`] = `
 "local ____exports = {}
 a = (a & 4294967295) >> b
@@ -389,12 +382,6 @@ return ____exports"
 exports[`Bitop [5.3] ("a>>>b") 1`] = `
 "local ____exports = {}
 ____exports.__result = (a & 4294967295) >> b
-return ____exports"
-`;
-
-exports[`Bitop [5.3] ("a>>b") 1`] = `
-"local ____exports = {}
-____exports.__result = (((a & 4294967295) ~ 2147483648) - 2147483648) // (1 << (b & 31))
 return ____exports"
 `;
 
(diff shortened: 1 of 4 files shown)


test/unit/__snapshots__/expressions.spec.ts.snap
```diff
@@ -527,6 +560,12 @@ ____exports.__result = (a & 4294967295) >> b
 return ____exports"
 `;
 
+exports[`Bitop [5.5] ("a>>b") 1`] = `
+"local ____exports = {}
+____exports.__result = (((a & 4294967295) ~ 2147483648) - 2147483648) // (1 << (b & 31))
```
Perryvw: Not sure this amount of magic in the generated code is really explainable. I  would be perfectly happy with fixing the wording on the diagnostic and telling people to use `>>>` (or their own function) instead - and mentioning it in the caveat docs.

RealColdFry: agreed, this was a demonstration of the "magic" approach. Reverting to option 2 with improved diagnostics

RealColdFry: Side note: a chunk of the apparent magic was the printer, not the values. In the source the constants were `0xFFFFFFFF` (low 32 bits) and `0x80000000` (sign bit), readable in hex but opaque after `lua.createNumericLiteral` rendered them as decimal (`4294967295`, `2147483648`).

Might be worth letting the emitter preserve the original radix for numeric literals. Not relevant for this PR, but in other cases where hex is the meaningful form it would keep generated code readable. Could file a separate issue for this?

Issue #1676, opened 2026-01-20
Lua 5.5 support


God-damnit-all:
I haven't seen any issues open that reported on Lua 5.5, so it's possible you're not aware.

[Lua 5.5.0 was released on 22 Dec 2025](https://www.lua.org/versions.html#5.5). Its [main new features](https://www.lua.org/manual/5.5/readme.html#changes) are declarations for global variables, named vararg tables, more compact arrays, and major garbage collections done incrementally.

### Main changes

+ declarations for global variables
+ named vararg tables
+ for-loop variables are read only
+ floats are printed in decimal with enough digits to be read back correctly.
+ more levels for constructors
+ table.create
+ utf8.offset returns also final position of character
+ external strings (that use memory not managed by Lua)
+ new functions luaL_openselectedlibs and luaL_makeseed
+ major garbage collections done incrementally
+ more compact arrays (large arrays use about 60% less memory)
+ lua.c loads 'readline' dynamically
+ static (fixed) binaries (when loading a binary chunk in memory, Lua can reuse its original memory in some of the internal structures)
+ dump and undump reuse all strings
+ auxiliary buffer reuses buffer when it creates final string

PR #1723, opened 2026-05-02
test: run lua 5.5 tests on the lua 5.5 wasm binding

RealColdFry:
Lua 5.5 changes:

- 8.1 Incompatibilities in the Language https://www.lua.org/manual/5.5/manual.html#8.1
- Error Handling https://www.lua.org/manual/5.5/manual.html#2.3

Related to #1676 but doesn't address any of the listed changes there. Note that this incompat is not listed in the issue or in the linked change list: https://www.lua.org/manual/5.5/readme.html#changes (although the change list does link to the incompat list)

### DIFF
--- a/src/transformation/visitors/errors.ts
+++ b/src/transformation/visitors/errors.ts
@@ -175,6 +175,7 @@ export const transformTryStatement: FunctionVisitor<ts.TryStatement> = (statemen
     // is replaced by a string when it propagates out of a protected call. To
     // preserve `throw undefined` semantics, route through xpcall with a message
     // handler that wraps nil into a sentinel, and unwrap before the user catch.
+    // See: https://www.lua.org/manual/5.5/manual.html#8.1
     const wrapErrorObjects =
         context.options.luaTarget === LuaTarget.Lua55 || context.options.luaTarget === LuaTarget.Universal;
     if (wrapErrorObjects) {
--- a/src/transformation/visitors/errors.ts
+++ b/src/transformation/visitors/errors.ts
@@ -4,7 +4,7 @@ import * as lua from "../../LuaAST";
 import { FunctionVisitor, TransformationContext } from "../context";
 import { unsupportedForTarget, unsupportedForTargetButOverrideAvailable } from "../utils/diagnostics";
 import { createUnpackCall } from "../utils/lua-ast";
-import { transformLuaLibFunction } from "../utils/lualib";
+import { importLuaLibFeature, transformLuaLibFunction } from "../utils/lualib";
 import { findScope, LoopContinued, Scope, ScopeType } from "../utils/scope";
 import { isInAsyncFunction, isInGeneratorFunction } from "../utils/typescript";
 import { wrapInAsyncAwaiter } from "./async-await";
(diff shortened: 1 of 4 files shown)


src/transformation/visitors/errors.ts
```diff
@@ -171,8 +171,23 @@ export const transformTryStatement: FunctionVisitor<ts.TryStatement> = (statemen
     const returnedIdentifier = lua.createIdentifier("____hasReturned");
     let returnCondition: lua.Expression | undefined;
 
-    const pCall = lua.createIdentifier("pcall");
-    const tryCall = lua.createCallExpression(pCall, [lua.createFunctionExpression(tryBlock)]);
+    // On Lua 5.5 (and Universal, which must work on 5.5), a `nil` error object
+    // is replaced by a string when it propagates out of a protected call. To
+    // preserve `throw undefined` semantics, route through xpcall with a message
+    // handler that wraps nil into a sentinel, and unwrap before the user catch.
+    // See: https://www.lua.org/manual/5.5/manual.html#8.1
```
Perryvw, line 178: What would be the result if we did not do this workaround? Do we get failing tests? Are we getting unexpected type failures at runtime?

RealColdFry: Running the below snippet

```lua
ok, err = pcall(function() error(nil) end)
print(ok, err, type(err))
```

In Lua 5.0-5.4, LuaJIT, Luau (Lune) gives

```text
false   nil     nil
```

while in 5.5:

```text
false	<no error object>	string
```

Without the wrapper, `error(nil)` in Lua 5.5 surfaces as the string `"<no error object>"` rather than `nil`. This breaks the `error.spec.ts` "throw and catch undefined" test (https://github.com/TypeScriptToLua/TypeScriptToLua/actions/runs/25261709047/job/74069980641?pr=1723#step:6:250).

```
  ● throw and catch undefined                                                                                                                
                                                                                                                                             
    expect(received).toEqual(expected) // deep equality                                                                                      
                                                                                                                                             
    Expected: undefined                                                                                                                      
    Received: "<no error object>"                                                                                                            
                                                                                                                                             
      405 |         const luaResult = this.getLuaExecutionResult();                                                                          
      406 |         const jsResult = this.getJsExecutionResult();                                                                            
    > 407 |         expect(luaResult).toEqual(jsResult);                                                                                     
          |                           ^                                                                                                      
      408 |                                                                                                                                  
      409 |         return this;                                                                                                             
      410 |     }                                                                                                                            
                                                                                                                                             
      at FunctionTestBuilder.expectToMatchJsResult (test/util.ts:407:27)                                                                     
      at test/unit/error.spec.ts:320:7 
```

Otherwise, in

```typescript
function doThing() {
  throw undefined;
}

function main() {
  try {
    doThing();
    return "didThing";
  } catch (
    e // e: unknown
  ) {
    if (e === undefined) {
      // handle the "thrown undefined" case
      return "undefined";
    }
    if (e instanceof Error) {
      return "Error[" + e.message + "]";
    }
    if (typeof e === "string") {
      // Lua 5.5 gives a string
      return "string[" + e + "]";
    }
    return "something else";
  }
}

console.log(main());

```

if `doThing()` threw undefined, on 5.5 `e` is actually the string `<no error object>` at runtime.

